PR #1, opened 2016-07-07
Add build model

d2lam:
Initial effort of refactoring code. This is simply moving build model outside of the builds plugin (https://github.com/screwdriver-cd/plugin-builds/tree/master/lib). 
Will slowly add on other models and figure out what should index be.


nkatzman:
As an initial move 👍

Filbird:
gr8 m8

PR #2, opened 2016-07-07
Add get,list,update to build model

d2lam:
Moving these functions to models. 
Also updated README to resolve https://github.com/screwdriver-cd/models/issues/3

### DIFF
--- a/lib/buildmodel.js
+++ b/lib/buildmodel.js
@@ -3,6 +3,7 @@ const async = require('async');
 const Executor = require('screwdriver-executor-k8s');
 const hashr = require('screwdriver-hashr');
 const hoek = require('hoek');
+const table = 'builds';
 
 /**
  * Given a Job ID, look up the associated Pipeline ID
@@ -66,7 +67,7 @@ const BuildModel = class {
     }
 
     /**
-     * Create & start a new job
+     * Create & start a new build
      * @method create
      * @param  {Object}   config           Config object
      * @param  {String}   config.jobId     The ID of the associated job to start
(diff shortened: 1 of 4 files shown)


lib/buildmodel.js
```diff
@@ -135,6 +135,66 @@ const BuildModel = class {
     stream(config, response) {
         return this.executor.stream(config, response);
     }
+
+    /**
+     * Get a build based on id
+     * @param  {Object}   config            Config object
+     * @param  {String}   config.table      The table name
```
nkatzman: Why does the buldmdoel require a table parameter? Would the buildModel ever get a record from outside the builds table?

d2lam: No, I was thinking bout removing it too and construct the config object here instead of inside the plugin.

lib/buildmodel.js
```diff
@@ -135,6 +135,81 @@ const BuildModel = class {
...
+                id
+            }
+        };
+
+        this.datastore.get(config, (err, data) => {
+            if (err) {
+                return callback(err);
+            }
+            if (!data) {
+                return callback(new Error('Record does not exist'));
```
nkatzman: Should the model's be the ones responsible for bubbling up Boom errors? If not, how do we plan on knowing that this is a 404 ?

lib/buildmodel.js
```diff
@@ -135,6 +135,81 @@ const BuildModel = class {
...
+            table: 'builds',
+            params: {},
+            paginate: {
+                count: paginate.count,
+                page: paginate.page
+            }
+        };
+
+        this.datastore.scan(config, (err, data) => {
+            if (err) {
```
nkatzman: Do we need this if statement?

lib/buildmodel.js
```diff
@@ -135,6 +135,81 @@ const BuildModel = class {
...
+        });
+    }
+
+    /**
+     * Update a build
+     * @param  {String}    id             The key of the record to retrieve
+     * @param  {Object}    data           The new data object to update with
+     * @return {Function} callback
+     */
+    update(id, data, callback) {
```
nkatzman: I think it's better to still pass `id` and `data` in a `config` object so that the arrity of all the functions are the same

lib/buildmodel.js
```diff
@@ -135,6 +135,61 @@ const BuildModel = class {
...
     }
+
+    /**
+     * Get a build based on id
+     * @param  {String}   id                The key of the record to retrieve
+     * @return {Function} callback
+     */
+    get(id, callback) {
+        const config = {
+            table: 'builds',
```
nkatzman: Maybe make this a const at the top of the module since we use it 3 times

lib/buildmodel.js
```diff
@@ -135,6 +136,61 @@ const BuildModel = class {
...
+    }
+
+    /**
+     * Update a build
+     * @param  {Object}    config         Config object
+     * @param  {String}    config.id      The key of the record to retrieve
+     * @param  {Object}    config.data    The new data object to update with
+     * @return {Function}  callback
+     */
+    update(config, callback) {
```
nkatzman, line 183: Who's responsibility should it be to make sure that the build exists before updating?

d2lam, line 183: Either model or datastore. Not sure where is better. What do you think?

nkatzman, line 183: Well I'm thinking between Model vs Handler. Right now lets say the handler needs to do it, but if we find that we are duplicating a lot of code lets move it into the model. Just things to think about with the behavior that we want to force. 

The question is: Should an update be allowed to happen if the data doesn't exist?

d2lam, line 183: Datastore doesn't do an update if the record doesn't exist.

nkatzman, line 183: Oh sweet you're right 👍 . This is fine then

test/lib/buildmodel.test.js
```diff
@@ -74,13 +76,74 @@ describe('Build Model', () => {
...
-
             assert.calledWith(executorMock.stream, {
                 buildId
             }, streamStub);
         });
     });
 
+    describe('get', () => {
+        it('calls datastore get and returns correct values', (done) => {
+            const build = new BuildModel(datastore);
```
nkatzman: should this go in the beforeEach?

nkatzman:
Besides my final comment in the test, looks fine 👍

nkatzman:
Readme ?? :D

tkyi:
👍  LGTM!
giphy

PR #4, opened 2016-07-12
Add pipeline model

d2lam:
Based on this morning discussion:
- In create: only create the pipeline
- In sync: syncing the pipeline. Right now it only creates the main job. I don't want to add code to do actual syncing yet since the purpose of this PR is to refactor, not to add new features. 
- It's the plugin's responsibility to call pipeline create then pipeline sync. 
- Model will be instantiated in every API call, inside the handler.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,0 +1,5 @@
+'use strict';
+const Build = require('./lib/buildmodel');
+const Pipeline = require('./lib/pipelinemodel');
+
+module.exports = { Build, Pipeline };
--- a/lib/buildmodel.js
+++ b/lib/buildmodel.js
@@ -3,7 +3,7 @@ const async = require('async');
 const Executor = require('screwdriver-executor-k8s');
 const hashr = require('screwdriver-hashr');
 const hoek = require('hoek');
-const table = 'builds';
+const PipelineModel = require('./pipelinemodel');
 
 /**
  * Given a Job ID, look up the associated Pipeline ID
@@ -26,7 +26,6 @@ function fetchPipelineId(datastore, jobId, callback) {
 /**
  * Given a Pipeline ID, look up the associated SCM Url
  *
- * TODO: Use PipelineModel instead
  * @method fetchScmUrl
  * @param  {Object}    datastore          The datastore object that can retrieve the data from the datastore
  * @param  {Object}    params             An object
@@ -35,14 +34,10 @@ function fetchPipelineId(datastore, jobId, callback) {
  *                                        keys: "pipelineID" and "scmUrl"
  */
 function fetchScmUrl(datastore, params, callback) {
+    const pipelineModel = new PipelineModel(datastore);
     const pipelineId = params.pipelineId;
 
-    datastore.get({
-        table: 'pipelines',
-        params: {
-            id: pipelineId
-        }
-    }, (err, data) => {
+    pipelineModel.get(pipelineId, (err, data) => {
         if (err) {
             return callback(err);
         }
@@ -64,6 +59,7 @@ const BuildModel = class {
     constructor(datastore) {
         this.datastore = datastore;
         this.executor = new Executor({});
+        this.table = 'builds';
     }
 
     /**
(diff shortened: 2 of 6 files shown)


lib/pipelinemodel.js
```diff
@@ -0,0 +1,134 @@
...
+                table,
+                params: {
+                    id: pipelineId,
+                    data: pipelineData
+                }
+            };
+
+            async.series([
+                async.apply(this.datastore.save, pipelineConfig),
+                (next) => {
```
nkatzman: Should the pipeline model handle the creation of a job, or should the `plugin-pipeline` handle the creation of the job?

d2lam: Let's discuss this with other people and see.

stjohnjohnson:
LGTM. 👍

index.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+const build = require('./lib/buildmodel');
```
tkyi: should we camelCase/snake_case these file names?

shruthi-venkateswaran:
👍

lib/buildmodel.js
```diff
@@ -35,14 +35,10 @@ function fetchPipelineId(datastore, jobId, callback) {
  *                                        keys: "pipelineID" and "scmUrl"
  */
 function fetchScmUrl(datastore, params, callback) {
```
nkatzman, line 36: Are there any other cases when we would need to `fetchScmUrl`? If so, should we consider making it a function provided by the model?

lib/pipelinemodel.js
```diff
@@ -0,0 +1,144 @@
...
+     */
+    constructor(datastore) {
+        this.datastore = datastore;
+    }
+
+    /**
+     * Create a pipeline
+     * @method create
+     * @param  {Object}   config           Config object to create the pipeline with
+     * @param  {String}   config.scmUrl    The scmUrl of the repository
```
nkatzman: config can also contain `configUrl`

lib/pipelinemodel.js
```diff
@@ -0,0 +1,144 @@
...
+            const id = hashr.sha1(`${pipelineId}${name}`);
+            const jobConfig = {
+                table: 'jobs',
+                params: {
+                    id,
+                    data: {
+                        name,
+                        pipelineId,
+                        state: 'ENABLED',
+                        triggers: [],
```
nkatzman, line 79: stjohn is proposing that we remove triggers in https://github.com/screwdriver-cd/data-schema/pull/4

lib/pipelinemodel.js
```diff
@@ -0,0 +1,144 @@
+'use strict';
+const hashr = require('screwdriver-hashr');
+const hoek = require('hoek');
+const table = 'pipelines';
```
nkatzman: should this be a value set in the constructor? 

i.e. this.table ?

d2lam: yeah let me add that.

lib/pipelinemodel.js
d2lam, line 79: Sure but that requires a change to data-schema first, otherwise it will break everything. Do we want to include those changes in this PR or later?

lib/buildmodel.js
d2lam, line 36: Not sure where else would use it, but I can add it to the pipeline Model since it would make more sense staying there.

lib/pipelinemodel.js
nkatzman, line 79: Can change it later

test/lib/pipelinemodel.test.js
```diff
@@ -0,0 +1,236 @@
...
+    it('constructs', () => {
+        const result = new PipelineModel(datastore);
+
+        assert.isOk(result);
+    });
+
+    it('has the correct API', () => {
+        const result = new PipelineModel(datastore);
+
+        assert.property(result, 'create');
```
nkatzman, line 65: Do you also want to test the other methods in the model?

d2lam, line 65: good ctach

nkatzman:
👍

PR #6, opened 2016-07-15
Add job model

d2lam:
- Add job model 
- Use jobModel.create inside pipeline model to create the job 'main'
- Remove trigger, triggeredBy

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,28 +13,18 @@ npm install screwdriver-models
 ```js
 'use strict';
 const Model = require('screwdriver-models');
+const Pipeline = new Model.Pipeline(datastore);
+const config = {
+    page: 2,
+    count: 3
+}
 
-module.exports = (datastore) => ({
-    method: 'GET',
-    path: '/pipelines',
-    config: {
-        description: 'Get pipelines with pagination',
-        notes: 'Returns all pipeline records',
-        tags: ['api', 'pipelines'],
-        handler: (request, reply) => {
-            const Pipeline = new Model.Pipeline(datastore);
-            Pipeline.list(request.query, reply);
-        },
-        response: {
-            schema: listSchema
-        },
-        validate: {
-            query: schema.pagination
-        }
+Pipeline.list(config, (err, result) => {
+    if (!err) {
+        console.log(result);
     }
 });
 ```
-
 
 #### Create
 Create a pipeline & create a default job called `main`
@@ -47,7 +37,7 @@ create(config, callback)
 | config        | Object | Yes | Configuration Object |
 | config.scmUrl | String | Yes | Source Code URL for the application |
 | config.configUrl | String | No | Source Code URL for Screwdriver configuration |
-| callback | Function | Yes | Callback function|
+| callback | Function | Yes | Callback function fn(err) |
 
 #### Get
 Get a pipeline based on id
@@ -58,7 +48,7 @@ get(id, callback)
 | Parameter        | Type  |  Description |
 | :-------------   | :---- | :-------------|
 | id | String | The unique ID for the pipeline |
-| callback | Function | Callback function|
+| callback | Function | Callback function fn(err, result) where result is the pipeline with the specific id|
 
 #### List
 List builds with pagination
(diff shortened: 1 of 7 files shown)


lib/pipelinemodel.js
```diff
@@ -65,24 +66,13 @@ const PipelineModel = class {
                 return callback(error);
             }
 
-            // TODO: use job model
-            const name = 'main';
-            const id = hashr.sha1(`${pipelineId}${name}`);
+            const jobModel = new JobModel(this.datastore);
```
nkatzman, line 69: would it be better to instantiate this in the configure?

nkatzman, line 69: we can do it later if we need to

nkatzman:
👍  once you add a README

FenrirUnbound:
The README is a bit confusing. The examples include a HapiJS-specific implementation. Can we just make it straight to the point and show the model being used without the context fluff?

README.md
```diff
@@ -98,6 +91,77 @@ sync(config, callback)
...
+#### Get
+Get a job based on id
+```
+get(id, callback)
+```
+
+| Parameter        | Type  |  Description |
+| :-------------   | :---- | :-------------|
+| id | String | The unique ID for the job |
+| callback | Function | Callback function|
```
FenrirUnbound: The callback isn't just a notification. The callback also returns back data. We should definitely document what gets passed to the callback function. Should we document what the return data looks like?

README.md
```diff
@@ -98,6 +91,77 @@ sync(config, callback)
...
+```
+list(paginate, callback)
+```
+
+| Parameter        | Type  |  Description |
+| :-------------   | :---- | :-------------|
+| paginate        | Object | Pagination Object |
+| paginate.page | Number | The page for pagination |
+| paginate.count | Number | The count for pagination |
+| callback | Function | Callback function |
```
FenrirUnbound: Same concern with the returned data as the `get()`

lib/jobmodel.js
```diff
@@ -0,0 +1,97 @@
...
+     */
+    constructor(datastore) {
+        this.datastore = datastore;
+        this.table = 'jobs';
+    }
+
+    /**
+     * Get a job based on id
+     * @param  {String}   id                The key of the record to retrieve
+     * @return {Function} callback
```
FenrirUnbound: What does the method signature look like for the callback?

lib/jobmodel.js
```diff
@@ -0,0 +1,97 @@
...
+
+        return this.datastore.get(config, callback);
+    }
+
+    /**
+     * List jobs with pagination
+     * @param  {Object}   paginate                  Config object
+     * @param  {Number}   paginate.count     Number of items per page
+     * @param  {Number}   paginate.page      Specific page of the set to return
+     * @return {Function} callback
```
FenrirUnbound: What does the method signature look like for the callback?

lib/jobmodel.js
```diff
@@ -0,0 +1,97 @@
...
+
+        return this.datastore.scan(config, callback);
+    }
+
+    /**
+     * Update a job
+     * @param  {Object}    config         Config object
+     * @param  {String}    config.id      The key of the record to retrieve
+     * @param  {Object}    config.data    The new data object to update with
+     * @return {Function}  callback
```
FenrirUnbound: What does the method signature look like for the callback?

lib/jobmodel.js
```diff
@@ -0,0 +1,97 @@
...
+            }
+        };
+
+        return this.datastore.scan(config, callback);
+    }
+
+    /**
+     * Update a job
+     * @param  {Object}    config         Config object
+     * @param  {String}    config.id      The key of the record to retrieve
```
FenrirUnbound: What's the "key"?

test/lib/jobmodel.test.js
```diff
@@ -0,0 +1,157 @@
...
+        datastore = null;
+        mockery.deregisterAll();
+        mockery.resetCache();
+    });
+
+    after(() => {
+        mockery.disable();
+    });
+
+    it('constructs', () => {
```
FenrirUnbound: The other tests infer that the model can be properly created. Do we need this specific test?

test/lib/jobmodel.test.js
```diff
@@ -0,0 +1,157 @@
...
+        const result = new JobModel(datastore);
+
+        assert.isOk(result);
+    });
+
+    it('has the correct API', () => {
```
FenrirUnbound: The other tests infer that the public methods exist in order to assert their functionality. Do we need this specific test?

PR #7, opened 2016-07-18
Create base model

minzcmu:
- Create base model 
- Base model has get, list and update methods for re-use in build, job and pipeline


index.js
```diff
@@ -1,5 +1,6 @@
 'use strict';
 const Build = require('./lib/buildmodel');
+const Job = require('./lib/jobmodel');
```
stjohnjohnson: Can we please use a better file naming schema here.

Something like:
- `./lib/models/job`
- `./lib/job`

lib/basemodel.js
```diff
@@ -0,0 +1,70 @@
+'use strict';
+
+class BaseModel {
+
```
stjohnjohnson: Extra line

stjohnjohnson:
LGTM (besides the file naming, which can be addressed in another PR if you want)

index.js
d2lam: ok we will change it to `./lib/job` since it's given that these are models.

tkyi:
👍 LGTM

PR #8, opened 2016-07-19
Add platform model

d2lam:
- Add platform model 
- Remove the check if scmUrl already exists inside pipeline model's `create`. This will be the plugin's responsibility as discussed with @nkatzman yesterday. Open to suggestion on this.
- Remove unnecessary fields

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,78 @@
 ```bash
 npm install screwdriver-models
 ```
+### Platform Model
+```js
+'use strict';
+const Model = require('screwdriver-models');
+const Platform = new Model.Platform(datastore);
+const config = {
+    page: 2,
+    count: 3
+}
+
+Platform.list(config, (err, result) => {
+    if (!err) {
+        console.log(result);
+    }
+});
+```
+
+#### Create
+Create a new platform
+```
+create(config, callback)
+```
+
+| Parameter        | Type  | Required | Description |
+| :-------------   | :---- | :---- | :-------------|
+| config        | Object | Yes | Configuration Object |
+| config.name | String | Yes | Platform name |
+| config.version | String | Yes | Platform version |
+| config.config | String | No | Config of the platform |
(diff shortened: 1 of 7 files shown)


lib/pipeline.js
```diff
@@ -25,30 +25,22 @@ class PipelineModel extends BaseModel {
...
+        const platform = 'generic@1';
+        const pipelineData = hoek.applyToDefaults({
+            createTime,
+            platform,
+            configUrl: config.scmUrl
+        }, config);
+        const pipelineConfig = {
+            table: this.table,
+            params: {
+                id: pipelineId,
```
nkatzman: should just call the variables `id` and `data`

lib/platform.js
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * Create a platform
+     * @method create
+     * @param  {Object}   config                Config object to create the platform with
+     * @param  {String}   config.name           The name of the platform
+     * @param  {String}   config.version        The version of the platform
+     * @param  {Function} callback              fn(err)
+     */
+    create(config, callback) {
+        const id = hashr.sha1({
```
nkatzman, line 25: We keep track of the of unique keys that should be used for hashing here: https://github.com/screwdriver-cd/data-schema/blob/master/model/platform.js#L97

Should we find a way to generically implement a hashing function for all models?

d2lam, line 25: Let's do this in a separate PR

nkatzman:
👍

README.md
```diff
@@ -8,6 +8,78 @@
...
+```
+
+#### Create
+Create a new platform
+```
+create(config, callback)
+```
+
+| Parameter        | Type  | Required | Description |
+| :-------------   | :---- | | :---- | :-------------|
```
tkyi: Extra `|` on this line.

tkyi:
Just need to fix the Create table in the README
👍

PR #12, opened 2016-07-21
Pass in Executor

### LINKED ISSUE #9: Builds Model should accept an executor
Right now we instantiate an Executor every time that a BuildModel is instantiated

Instead, because we want the Executor model to be a plugin model like the datastore, the model should support having an Executor passed in

nkatzman:
With the recent crash regarding `vogels`, should we allow the datastore interface to support promises as well?

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,7 +232,7 @@ update(config, callback)
 ```js
 'use strict';
 const Model = require('screwdriver-models');
-const Build = new Model.Build(datastore);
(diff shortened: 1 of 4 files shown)


FenrirUnbound:
👍 

Looks fine to me

t3_4txfxw

stjohnjohnson:
👍 woot 
this_cold_pizza_is_too_hot_to_handle_

PR #16, opened 2016-07-25
Ensuring that sync will return null when the pipeline does not exist

stjohnjohnson:
This seems like a strange problem to have.  I expected the datastore to throw an error, instead it just returns null.


test/lib/pipeline.test.js
```diff
@@ -146,5 +148,14 @@ describe('Pipeline Model', () => {
                 done();
             });
         });
+
+        it('returns error if pipeline does not exist', (done) => {
```
d2lam: returns null?

d2lam:
Sorry I think I wrote this code :(

test/lib/pipeline.test.js
stjohnjohnson: Fixed

tkyi:
👍 

giphy

d2lam:
👍

PR #17, opened 2016-07-26
Exposing the generateId function and consuming the keys from data-schema

stjohnjohnson:
Kinda significant refactoring.  Broke all the hashes (hence the major version bump).

We are now using the `tableName` and the `keys` from the `data-schema` module.

### DIFF
--- a/lib/base.js
+++ b/lib/base.js
@@ -1,14 +1,34 @@
 'use strict';
+const schema = require('screwdriver-data-schema');
+const hashr = require('screwdriver-hashr');
 
 class BaseModel {
-
     /**
      * Construct a BaseModel object
      * @method constructor
+     * @param  {String}    modelName         Name of the model to get from data-schema
      * @param  {Object}    datastore         Object that will perform operations on the datastore
      */
-    constructor(datastore) {
+    constructor(modelName, datastore) {
+        this.model = schema.models[modelName];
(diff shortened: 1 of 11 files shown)


lib/base.js
```diff
@@ -1,17 +1,36 @@
 'use strict';
+const schema = require('screwdriver-data-schema');
+const hashr = require('screwdriver-hashr');
 
 class BaseModel {
-
     /**
      * Construct a BaseModel object
      * @method constructor
      * @param  {Object}    datastore         Object that will perform operations on the datastore
```
tkyi: Add `modelName`

nkatzman:
👍

tkyi:
just need to add a quick comment
otherwise, LGTM👍

tkyi:
and update README examples please

tkyi:
👍

lib/pipeline.js
```diff
@@ -24,7 +22,7 @@ class PipelineModel extends BaseModel {
      * @param  {Function} callback              fn(err, data) where data is the newly created object
      */
     create(config, callback) {
-        const id = hashr.sha1(config.scmUrl);
+        const id = this.generateId(config);
```
d2lam: This might not work. Config sometimes includes configUrl and sometimes not. I think you need to create a new object here that only includes scmUrl.

PR #23, opened 2016-07-27
Get github permissions

d2lam:
This PR does 2 things:

1) Add a method to get the user's permissions of a specific repository. This takes in a config object:
`{ username: 'd2lam', scmUrl: 'git@github.com:screwdriver-cd/models.git'}` and returns the permissions:  `{admin: true, push: true, pull: true}`

2) Add admins field for pipeline create. This will be passed in from the api when calling pipeline.create

Waiting on https://github.com/screwdriver-cd/data-schema/pull/16

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,7 @@ create(config, callback)
 | Parameter        | Type  | Required  |  Description |
 | :-------------   | :---- | :---- | :-------------|
 | config        | Object | Yes | Configuration Object |
+| config.admins | Array | Yes | Array of admins for this pipeline |
 | config.scmUrl | String | Yes | Source Code URL for the application |
 | config.configUrl | String | No | Source Code URL for Screwdriver configuration |
 | callback | Function | Yes | Callback function fn(err, data) where data is the new pipeline that is created |
@@ -386,6 +387,20 @@ unsealToken(sealed, callback)
 | :-------------   | :---- | :-------------|
 | sealed | String | The token to unseal |
 | callback | Function | Callback function fn(err, unsealed) where unsealed is the unsealed token |
+
+
+#### Get User's Permissions For a Repo
+Get user's permissions for a specific repo
+```
+getPermissions(config, callback)
+```
+
+| Parameter        | Type  |  Description |
+| :-------------   | :---- | :-------------|
+| config | Object | Configuration object |
+| config.username | String | Username |
+| config.scmUrl | String | The scmUrl of the repo |
+| callback | Function | Callback function fn(err, permissions) where permissions is an object that includes the following keys: admin, pull, push with boolean values |
 
 ## Testing
 
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -17,7 +17,7 @@ class PipelineModel extends BaseModel {
      * Create a pipeline
      * @method create
      * @param  {Object}   config                Config object to create the pipeline with
-     * @param  {Object}   config.owners         The owners of this repository
+     * @param  {Object}   config.admins         The admins of this repository
      * @param  {String}   config.scmUrl         The scmUrl for the application
      * @param  {String}   [config.configUrl]    The configUrl for the application
      * @param  {Function} callback              fn(err, data) where data is the newly created object
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -17,6 +17,7 @@ class PipelineModel extends BaseModel {
      * Create a pipeline
      * @method create
      * @param  {Object}   config                Config object to create the pipeline with
+     * @param  {Object}   config.owners         The owners of this repository
      * @param  {String}   config.scmUrl         The scmUrl for the application
      * @param  {String}   [config.configUrl]    The configUrl for the application
      * @param  {Function} callback              fn(err, data) where data is the newly created object
--- a/lib/user.js
+++ b/lib/user.js
@@ -2,7 +2,8 @@
 const BaseModel = require('./base');
 const iron = require('iron');
 const async = require('async');
-const github = require('github');
+const Github = require('github');
+const schema = require('screwdriver-data-schema');
 
 class UserModel extends BaseModel {
     /**
@@ -65,6 +66,7 @@ class UserModel extends BaseModel {
      */
     getPermissions(config, callback) {
         const id = this.generateId({ username: config.username });
+        const github = new Github();
 
         async.waterfall([
             (next) => {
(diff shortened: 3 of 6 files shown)


README.md
```diff
@@ -107,6 +107,7 @@ create(config, callback)
 | Parameter        | Type  | Required  |  Description |
 | :-------------   | :---- | :---- | :-------------|
 | config        | Object | Yes | Configuration Object |
+| config.admins | Array | Yes | Array of admins for this pipeline |
```
nkatzman, line 110: did you update the data-schema?

lib/user.js
```diff
@@ -52,6 +54,45 @@ class UserModel extends BaseModel {
...
+
+        async.waterfall([
+            (next) => {
+                this.get(id, next);
+            },
+            (user, next) => {
+                this.unsealToken(user.token, next);
+            },
+            (unsealed, next) => {
+                const matched = config.scmUrl.match(/^git@([^:]+):([^\/]+)\/(.+?)\.git(\/.+)?$/);
```
nkatzman: is this regex in the data-schema?

README.md
d2lam, line 110: Good call. i didn't

lib/user.js
```diff
@@ -52,6 +54,45 @@ class UserModel extends BaseModel {
...
+
```
tkyi: Need to instantiate the github client with something like:

```
var github = new Client({});
```

lib/user.js
```diff
@@ -52,6 +54,45 @@ class UserModel extends BaseModel {
...
+                github.authenticate({
+                    type: 'oauth',
+                    token: unsealed
+                });
+                github.repos.get({
+                    user: matched[2],
```
tkyi: what if the repo does not belong to a user/is in an organization?

d2lam: then `user` would be the `org`

lib/user.js
d2lam: Added!

lib/user.js
tkyi: But do you need to make sure the `username` is in the `org`?

d2lam: Nope. this would check again the user with the githubtoken = `unsealed` against this particular repo.

tkyi: Ah I see. Nice!

tkyi:
👍  LGTM
giphy-1

d2lam:
Still need to add circuit breaker.

tkyi:
👍 LGTM

lib/pipeline.js
```diff
@@ -17,6 +17,7 @@ class PipelineModel extends BaseModel {
      * Create a pipeline
      * @method create
      * @param  {Object}   config                Config object to create the pipeline with
+     * @param  {Object}   config.owners         The owners of this repository
```
stjohnjohnson: Owners or Admins?  I see it both in this PR

lib/user.js
```diff
@@ -1,8 +1,32 @@
...
+ * @param  {Object}   options            An object that tells what command & params to run
+ * @param  {String}   options.token      Github token
+ * @param  {String}   options.action     Github method. For example: get
+ * @param  {Object}   options.params     Parameters to run with
+ * @param  {Function} callback           Callback function from github API
+ */
+function command(options, callback) {
+    const github = new Github();
+
+    console.log(options);
```
stjohnjohnson: Console.log?

lib/user.js
```diff
@@ -52,6 +76,46 @@ class UserModel extends BaseModel {
...
+        async.waterfall([
+            (next) => {
+                this.get(id, next);
+            },
+            (user, next) => {
+                this.unsealToken(user.token, next);
+            },
+            (unsealed, next) => {
+                const matched = (schema.config.regex.SCM_URL).exec(config.scmUrl);
+                const breaker = new Breaker(command);
```
stjohnjohnson: Because this is defined inside the scope of getPermissions, I don't think this will work globally.  @nkatzman can you comment here on the best way to use it

lib/user.js
```diff
@@ -52,6 +76,46 @@ class UserModel extends BaseModel {
...
+
+        async.waterfall([
+            (next) => {
+                this.get(id, next);
+            },
+            (user, next) => {
+                this.unsealToken(user.token, next);
+            },
+            (unsealed, next) => {
+                const matched = (schema.config.regex.SCM_URL).exec(config.scmUrl);
```
stjohnjohnson, line 99: It shouldn't, but what if this fails validation?  Should there be a check in here or is there already a check earlier?

lib/pipeline.js
tkyi: Should be admins

lib/user.js
d2lam, line 99: Does this count?
https://github.com/screwdriver-cd/screwdriver/blob/f4000828656687d784b21afbb51b6dd87c427a22/plugins/pipelines/create.js#L73

The scmUrl is passed in is from the payload, which should be validated in the api already. I can still add a check here if you prefer but I think it will never happen.

lib/pipeline.js
d2lam: @stjohnjohnson Good catch. I was debating if it should be named `owners` or `admins` and decided `admins` makes more sense but forgot to change the comment.

lib/user.js
nkatzman: right. We'd want to instantiate the breaker once and use it all around

lib/user.js
```diff
@@ -1,8 +1,31 @@
...
+/**
+ * Command to run
+ * @param  {Object}   options            An object that tells what command & params to run
+ * @param  {String}   options.token      Github token
+ * @param  {String}   options.action     Github method. For example: get
+ * @param  {Object}   options.params     Parameters to run with
+ * @param  {Function} callback           Callback function from github API
+ */
+function command(options, callback) {
+    const github = new Github();
```
nkatzman: do we need to instantiate this with every command too?

d2lam: I can put it outside

lib/user.js
```diff
@@ -12,6 +34,7 @@ class UserModel extends BaseModel {
     constructor(datastore, password) {
         super('user', datastore);
         this.password = password;
+        this.githubBreaker = new Breaker(githubCommand);
```
nkatzman: So even with this, how many times is the Users model going to be instantiated?

lib/user.js
stjohnjohnson, line 99: nah, that's perfect.  Just wanted to verify.

nkatzman:
Seems fine 👍

stjohnjohnson:
LGTM 👍 
haha

PR #25, opened 2016-07-28
Fix bug for userid

d2lam:
Bug bug


stjohnjohnson:
👎 GenerateId picks the right fields.

d2lam:
Right...

PR #27, opened 2016-08-01
Create github status when the build starts

d2lam:
This PR creates a github status when a build starts.
- Build model now takes in a 3rd parameter `password`. This is because we need to create a User model inside the constructor so that we can pass that into githubHelper. We would need to change the api to pass it in, similar to this: https://github.com/screwdriver-cd/screwdriver/blob/master/bin/server#L39
- Build create now takes in 2 extra params: config.username and config.sha: https://github.com/screwdriver-cd/screwdriver/pull/66
- Github helper has the run command, which get the user, unseal the user's token, and call github command.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -4,10 +4,11 @@ const hoek = require('hoek');
 const BaseModel = require('./base');
 const PipelineModel = require('./pipeline');
 const JobModel = require('./job');
+const UserModel = require('./user');
+const githubHelper = require('./github');
 
 /**
  * Given a Job ID, look up the associated Pipeline ID
- *
  * @method fetchJob
  * @param  {Object}        datastore The datastore object that can retrieve the data from the datastore
  * @param  {String}        jobId     The ID of the job to look at
@@ -21,7 +22,6 @@ function fetchJob(datastore, jobId, callback) {
 
 /**
  * Given a Pipeline ID, look up the associated SCM Url
- *
  * @method fetchScmUrl
  * @param  {Object}    datastore          The datastore object that can retrieve the data from the datastore
  * @param  {Object}    params             An object
@@ -54,23 +54,29 @@ class BuildModel extends BaseModel {
      * @method constructor
      * @param  {Object}    datastore         Object that will perform operations on the datastore
      * @param  {Object}    executor          Object that will perform executor operations
+     * @param  {String}    password          Login password
      */
-    constructor(datastore, executor) {
+    constructor(datastore, executor, password) {
         super('build', datastore);
         this.executor = executor;
+        this.password = password;
+        this.user = new UserModel(datastore, password);
     }
 
     /**
      * Create & start a new build
      * @method create
      * @param  {Object}   config                Config object
+     * @param  {String}   config.username       Username
      * @param  {String}   config.jobId          The ID of the associated job to start
+     * @param  {String}   config.sha            The sha of the build
      * @param  {String}   [config.container]    The kind of container to use
      * @param  {Function} callback              fn(err, data) where data is the newly created object
      */
     create(config, callback) {
         const container = config.container || 'node:4';
         const jobId = config.jobId;
+        const sha = config.sha;
         const now = Date.now();
         const id = this.generateId({
             jobId,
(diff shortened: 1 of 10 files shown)


lib/build.js
```diff
@@ -98,26 +108,38 @@ class BuildModel extends BaseModel {
                 });
             },
             async.apply(fetchJob, this.datastore, jobId),
-            async.apply(fetchScmUrl, this.datastore)
-        ], (err, data) => {
+            async.apply(fetchScmUrl, this.datastore),
+            (data, next) => {
```
tkyi, line 110: should this be `(err, data, next)`?

test/lib/github.test.js
```diff
@@ -0,0 +1,219 @@
...
+const mockery = require('mockery');
+const sinon = require('sinon');
+
+sinon.assert.expose(assert, { prefix: '' });
+
+describe('Github', () => {
+    const scmUrl = 'git@github.com:screwdriver-cd/data-schema.git';
+    let schemaMock;
+    let userMock;
+    let BreakerMock;
```
tkyi: make capitalization consistent

lib/build.js
stjohnjohnson, line 110: Nah, it's a waterfall.  It only gets data, next

lib/build.js
```diff
@@ -98,26 +108,38 @@ class BuildModel extends BaseModel {
...
+                // Create github status
+                githubHelper.run({
+                    User: this.User,
+                    username: config.username,
+                    action: 'createStatus',
+                    params: {
+                        user: repoInfo.user,
+                        repo: repoInfo.repo,
+                        sha,
+                        state: 'pending'
```
stjohnjohnson: Can we set the context to 'Screwdriver'.

lib/github.js
```diff
@@ -0,0 +1,84 @@
...
+    return {
+        user: matched[2],
+        repo: matched[3]
+    };
+}
+
+/**
+ * Run command. This first gets the user, then unseal the user's token and call the github command
+ * @param  {Object}   config            Configuration object
+ * @param  {Model}    config.User       User Model
```
stjohnjohnson: Why is User capitalized?  It's not a class, it's an instance.

lib/github.js
```diff
@@ -0,0 +1,84 @@
...
+            github.authenticate({
+                type: 'oauth',
+                token: unsealed
+            });
+            githubBreaker.runCommand({
+                action: config.action,
+                params: config.params || {}
+            }, next);
+        }
+    ], (err, res) => callback(err, res)
```
stjohnjohnson: Why not just put callback in here?

stjohnjohnson:
yesss
👍

lib/github.js
```diff
@@ -0,0 +1,83 @@
...
+const async = require('async');
+const Breaker = require('circuit-fuses');
+const schema = require('screwdriver-data-schema');
+const Github = require('github');
+const github = new Github();
+let cached = null;
+
+/**
+ * Github command to run
+ * @param  {Object}   options            An object that tells what command & params to run
```
petey: These are all missing the `@method` jsdoc key.

petey: Might be missing a docblockr setting.
screen shot 2016-08-01 at 5 52 48 pm

d2lam: Interesting. I do have that installed but it didn't enforce @method. 
I added them now.

d2lam:
I updated the PR to expose `github.js` since we will need to use it in the `api` to lookup the `sha` for `POST /v3/builds`

@stjohnjohnson  @petey  Can you guys review again? :D

stjohnjohnson:
Hmm, I figured you would leave the lookup to be in the models.  If the user doesn't provide a SHA, then look it up on build.create. 👍 otherwise.

d2lam:
@stjohnjohnson  Hmm true. I could do that as well. That might actually better since the logic for the handler will be simpler. In that case I don't really need to expose github.

d2lam:
@stjohnjohnson Let's add the lookup in the next PR since the intent of this story was just to get the Pull Request to work. I can work on that after this one is merged.

tkyi:
👍 
giphy-1

PR #30, opened 2016-08-02
Look up sha when it is not passed in

d2lam:
If sha is not passed in, call github API to get the sha

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -48,6 +48,31 @@ function fetchScmUrl(datastore, params, callback) {
     });
 }
 
+/**
+ * Look up repo info such as user, repo name, and sha
+ * @method getRepoInfo
+ * @param  {Model}    user          The user model
+ * @param  {Object}   config        Configuration object that might/might not include sha
+ * @param  {Object}   buildInfo     Other build info that will be returned. Includes
+ *                                  pipelineId, jobName, and scmUrl
+ * @param  {Function} callback      fn(err, sha)
+ */
+function getRepoInfo(user, config, buildInfo, callback) {
+    const repoInfo = githubHelper.getInfo(buildInfo.scmUrl);
+    const response = hoek.applyToDefaults(buildInfo, repoInfo);
+
+    if (config.sha) {
+        return callback(null, hoek.applyToDefaults(response, { sha: config.sha }));
+    }
+
+    return githubHelper.run({
+        user,
+        username: config.username,
+        action: 'getBranch',
+        params: repoInfo
+    }, (err, repo) => callback(err, hoek.applyToDefaults(response, { sha: repo.commit.sha })));
+}
+
 class BuildModel extends BaseModel {
     /**
      * Construct a BuildModel object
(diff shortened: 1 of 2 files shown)


lib/build.js
```diff
@@ -83,39 +83,62 @@ class BuildModel extends BaseModel {
             number: now
         });
         const initialBuildData = {
-            cause: 'Started by user',
+            cause: `Started by user ${config.username}`,
```
tkyi: 👍  cool!

tkyi:
👍 LGTM
giphy-2

lib/build.js
stjohnjohnson: It is cool, but it's not correct sadly.  Username is for looking up which GitHub token to use.  Maybe we should rename username and pass in it as another value.

lib/build.js
```diff
@@ -83,39 +83,62 @@ class BuildModel extends BaseModel {
...
+                buildInfo = data;
+                repoInfo = githubHelper.getInfo(buildInfo.scmUrl);
+                // If sha is passed in, save it
+                if (sha) {
+                    initialBuildData.sha = sha;
+
+                    return next(null, null);
+                }
+                // If sha is not passed in, look it up
+                githubHelper.run({
```
stjohnjohnson: Why not add return here?  Remove the eslint warning.

lib/build.js
d2lam: ah that's right.

lib/build.js
```diff
@@ -88,34 +88,56 @@ class BuildModel extends BaseModel {
...
         };
+        let buildInfo;
         let repoInfo;
 
         async.waterfall([
-            (next) => {
+            async.apply(fetchJob, this.datastore, jobId),
+            async.apply(fetchScmUrl, this.datastore),
+
+            /* eslint-disable consistent-return */
```
stjohnjohnson: Not needed

d2lam: oopsie

FenrirUnbound:
👍 

hxzmubg - imgur

lib/build.js
```diff
@@ -88,34 +88,55 @@ class BuildModel extends BaseModel {
...
         };
+        let buildInfo;
         let repoInfo;
 
         async.waterfall([
-            (next) => {
+            async.apply(fetchJob, this.datastore, jobId),
+            async.apply(fetchScmUrl, this.datastore),
+
+            (data, next) => {
```
petey: I think this is overly complex, and could be refactored.

You can make a method that would return a sha, either from the given data, or as a result of a github lookup. This can be refactored out of the create method to reduce complexity. The following method in the waterfall which performs `datastore.save()` could then just take in a sha, and save it to `initialBuildData`.

d2lam:
@petey  I refactored the code. Hope it looks cleaner now :)

stjohnjohnson:
LGTM 👍 
mrw_i_see_my_crazy_ex_gf_at_the_grocery_store

petey:
👍 

fishing

Issue #18, opened 2016-07-26
Configure Models once for Datastore and Executor


stjohnjohnson:
Right now we have to pass in datastore and executor each time we use a model.  Why not set them up once and pass the model collection into the plugins?

Would we ever have a model that has a different datastore than the others?  What about the executor?

If we go down that route, wouldn't a datastoreFactory or executorFactory be more appropriate?

FenrirUnbound:
The initial idea was to use a Factory for constructing the models. It was decided against it due to "future proofing".

At this point, we can slowly switch the models over to be obtained from a Factory method, and introduce it PR by PR.

FenrirUnbound:
To iterate our discussion from standup to here:

There were observed nuances with the current state of the models:

1. The models aren't treated like instances of an specific object. A `job` instance wouldn't represent a specific job and all its properties: it acts more as an interface to the datastore objects.
1. The model methods were treated as instance methods. It is now determined that some of those functions are better described as static methods, while some methods make sense to keep as instance methods.

petey:
Factories handle this: PR #35, #43

PR #31, opened 2016-08-04
[RFC] generate build model via factory

FenrirUnbound:
This is a **R**equest **F**or **C**omments

This is a proposal for addressing issue #18

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,15 @@
 'use strict';
 const Build = require('./lib/build');
+const factory = require('./lib/factory');
 const Job = require('./lib/job');
 const Pipeline = require('./lib/pipeline');
 const User = require('./lib/user');
 
-module.exports = { Build, Job, Pipeline, User };
+module.exports = {
+    Build,
+    configureBuildModel: factory.configureBuildModel,
+    getBuildModel: factory.getBuildModel,
+    Job,
+    Pipeline,
+    User
+};
--- a/lib/factory.js
+++ b/lib/factory.js
@@ -1,0 +1,28 @@
+'use strict';
+const Build = require('./build');
+let build;
+
+/**
+ * Configure the build model for future getBuildModel executions
+ * @method configureBuildModel
+ * @param  {Object}            datastore Object that will perform operations on the datastore
+ * @param  {Object}            executor  Object that will perform executor operations
+ * @param  {String}            password  Login password
+ */
+function configureBuildModel(datastore, executor, password) {
+    build = new Build(datastore, executor, password);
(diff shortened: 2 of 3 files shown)


index.js
```diff
@@ -1,7 +1,15 @@
...
 const Build = require('./lib/build');
+const factory = require('./lib/factory');
 const Job = require('./lib/job');
 const Pipeline = require('./lib/pipeline');
 const User = require('./lib/user');
 
-module.exports = { Build, Job, Pipeline, User };
+module.exports = {
+    Build,
+    configureBuildModel: factory.configureBuildModel,
```
stjohnjohnson, line 10: Should we do this here?  Why not just do this in the Build prototype?

nkatzman, line 10: Right, should every model just have a factoryMethod that will retrieve the instance of a model?

FenrirUnbound, line 10: > Why not just do this in the Build prototype?

If every model had a factoryMethod, what would the difference be between how we're doing it now?

Adding it to the prototype would be us using another method instead of the constructor. You're still exposing the user to invoke `new`, which doesn't yield us the benefits of centralizing the configuration in one call.

FenrirUnbound, line 10: > should every model just have a factoryMethod that will retrieve the instance of a model? 

Are you asking if there will be a method for each method, like `getJobModel`, `getPipelineModel`?

stjohnjohnson, line 10: Sorry, prototype is the not the right word I was looking for.  Static is the word I was looking for.

Example:

``` js
const Build = require('screwdriver-models');
Build.configure(datastore, executor);

const build = new Build();
```

stjohnjohnson, line 10: Ideally, it would be great to configure all of it in one place like:

``` js
const models = require('screwdriver-models');
const Build = models.Build;
const Job = models.Job;
models.configure({ datastore, executor, key });

const build = new Build();
const job = new Job()
```

stjohnjohnson, line 10: The more I think about this though, what's the point of using new?  These aren't treated like class instances.

An instance would be like:

``` js
const build = new Build(buildId);
build.status = 'FAILED';
build.save(next);
```

If we're going to be treating them statically (like we do now), then let's move it to entirely to a singleton pattern.

``` js
const models = require('screwdriver-models');
models.configure({ datastore, executor, key });

models.build.create(payload, next);
```

FenrirUnbound:
We discussed this further during standup. I've gone ahead and update Issue #18 with some of the high-level details. 

We'll close this PR and follow up with a different approach to resolve that issue.

PR #33, opened 2016-08-04
Use pipeline's admin to create status

d2lam:
Look up the admin of the pipeline to create github status. 
`Username` passed in is the person that triggered/created the build. Modify the cause to specify this: `Started by`

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -8,41 +8,26 @@ const UserModel = require('./user');
 const githubHelper = require('./github');
 
 /**
- * Given a Job ID, look up the associated Pipeline ID
- * @method fetchJob
- * @param  {Object}        datastore The datastore object that can retrieve the data from the datastore
- * @param  {String}        jobId     The ID of the job to look at
- * @param  {Function}      callback  fn(err, result), where result is a Job object
- */
-function fetchJob(datastore, jobId, callback) {
-    const jobModel = new JobModel(datastore);
-
-    jobModel.get(jobId, callback);
-}
-
-/**
  * Given a Pipeline ID, look up the associated SCM Url
  * @method fetchScmUrl
- * @param  {Object}    datastore          The datastore object that can retrieve the data from the datastore
- * @param  {Object}    params             An object
- * @param  {String}    params.pipelineId  The ID of the pipeline to look at
+ * @param  {Model }    pipeline           Instance of the Pipeline model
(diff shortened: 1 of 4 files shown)


lib/build.js
```diff
@@ -116,9 +116,9 @@ class BuildModel extends BaseModel {
         };
 
         async.waterfall([
-            async.apply(fetchJob, this.datastore, jobId),
-            async.apply(fetchScmUrl, this.datastore),
-            async.apply(getRepoInfo, this.user, config),
+            (next) => fetchJob(this.datastore, jobId, next),
+            (job, next) => fetchScmUrl(this.datastore, job, next),
```
stjohnjohnson: So much 👍 👍 💯

stjohnjohnson:
So much 👍  
woaaah

lib/build.js
```diff
@@ -145,10 +145,23 @@ class BuildModel extends BaseModel {
             },
 
             (response, next) => {
-                // Create github status
+                // Get the admin of the pipeline
+                const pipelineModel = new PipelineModel(this.datastore);
```
nkatzman: There is already a spot where the pipelineModel is instantiated, can we refactor that to happen only once?

lib/pipeline.js
```diff
@@ -88,6 +88,23 @@ class PipelineModel extends BaseModel {
...
     }
+
+    /**
+     * Given a pipeline id, fetch the build admin
+     * @method getAdmin
+     * @param  {String}       pipelineId    Identifier for the Pipeline
+     * @param  {Function}     callback      fn(error, admin) where admin is the admin of the pipeline
+    */
+   // TODO: Retry if admin access was revoked
+    getAdmin(pipelineId, callback) {
```
nkatzman: should this be getAdmins if we plan on retrying in the case where admin access was revoked?

Whoever consuming this should be doing the iteration, and not the function in the pipeline model

d2lam: This funciton should only get ONE admin that is valid.  If admin access was revoked, then it should retry until we get one that is valid. And at the same time, remove that person from the admin list. Whoever consumes this shouldn't have to do the retrying and removing. Seems more like the responsibility of the pipeline.

Also, when would ever need to get the list of admins? At least I don't see that use case right now.

nkatzman:
👍

PR #35, opened 2016-08-05
The models now promise

### LINKED ISSUE #5: Evaluate Promises
Should we be using promises internally in here?  Since we're going to be interacting with multiple models.

FenrirUnbound:
Decorated all the models to have a Promise interface. The endpoints are backwards compatible: if they are given a callback, they will invoke the callback as expected. If you omit a callback function, you'll receive a promise instead.

Merging this PR will resolve #5

### DIFF
--- a/lib/nodeify.js
+++ b/lib/nodeify.js
@@ -1,0 +1,66 @@
+'use strict';
+
+/**
+ * Determine whether to return a promise or use the callback
+ * @method nodeify
+ * @param  {Function}  func       The function to call
+ * @param  {Object}    config     The parameters to pass to the function
+ * @param  {Function}  [callback] Optional. Callback to invoke when func is completed.
+ * @return {Promise}              If no callback is provided, a Promise is returned.
+ */
+function nodeify() {
+    const args = Array.from(arguments);
+    const func = args.shift();
+    let callback = args[args.length - 1];
+
+    if (typeof callback === 'function') {
+        // To allow thrown errors to _NOT_ be caught in the promise chain
+        return process.nextTick(() => func.apply(null, args));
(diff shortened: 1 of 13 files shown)


lib/nodeify.js
```diff
@@ -0,0 +1,64 @@
...
+    const args = Array.from(arguments);
+    const func = args.shift();
+    let callback = args[args.length - 1];
+
+    if (typeof callback === 'function') {
+        // To allow thrown errors to _NOT_ be caught in the promise chain
+        return process.nextTick(() => func.apply(null, args));
+    } else if (typeof callback === 'undefined') {
+        args.pop();
+    }
```
stjohnjohnson: What would an else be here?

stjohnjohnson:
Okay, this is an awesome PR.  Super cool.  I would bump the minor rev on the package.json.

Quick question, is the intention to support both callback and promise for everything in models going forward?

lib/nodeify.js
FenrirUnbound: Nothing. It would fall through the rest of the method.

FenrirUnbound:
> is the intention to support both callback and promise for everything in models going forward?

For the near-term, yes. Going all-in with promises would require us to switch everything over from passing in callbacks to a Promise-chain.

petey:
:shipit:

nkatzman:
Squash commits and then add that comment in the nodeify code 👍

FenrirUnbound:
Commits squashed.

> add that comment in the nodeify code 

https://github.com/screwdriver-cd/models/pull/35/files#diff-cb6fb0408e4bce588c86b3c137060f54R20 😸

d2lam:
👍  Makes sense

PR #36, opened 2016-08-05
be explicit about the this arg for nodeify

FenrirUnbound:
We experienced an issue where the `datastore.save` performs: 
1. Validation
2. Calls `this._save`

Because of the extra level of indirection, a call to `nodeify(datastore.save)` would result in `this_save` to be unresolved. More specifically, the `this` is undefined, and so the datastore will throw a "cannot call 'save' of undefined". 

This PR aims to amend that bug.

### DIFF
--- a/lib/base.js
+++ b/lib/base.js
@@ -47,7 +47,7 @@ class BaseModel {
             }
         };
 
-        return nodeify(this.datastore.get, config, callback);
+        return nodeify.withContext(this.datastore, 'get', [config], callback);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


lib/base.js
```diff
@@ -47,7 +47,7 @@ class BaseModel {
             }
         };
 
-        return nodeify(this.datastore.get, config, callback);
+        return nodeify(this.datastore, this.datastore.get, config, callback);
```
petey: What if we did something like  `return nodeify(this.datastore, 'get', config, callback);` instead? Then we don't need to repeat the context twice.

petey: I guess we'd still need to determine if the second arg is a function or a string.

Alternatively, there could be a `nodeify.apply(context, fn, ...)` that would be more explicit.

stjohnjohnson:
mrw_i_hear_there_s_a_new_mcdonald_s_that_offers_all-you-can-eat_fries 👍

petey:
👍 

giphy

PR #38, opened 2016-08-06
Fix branch name case sensitivity

catto:
This PR fixes case sensitivity issue of the branch name. screwdriver-cd/screwdriver#82
`formatScmUrl` funtion will convert scmUrl to lowercase excepting a branch name.

This change breaks relation between pipelineId and scmUrl of existing pipelines. Please just close this PR if you don't want it.

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -82,13 +82,17 @@ class PipelineModel extends BaseModel {
     formatScmUrl(scmUrl) {
         let result = scmUrl;
         const matched = (schema.config.regex.SCM_URL).exec(result);
+        let branchName = matched[4];
 
         // Check if branch name exists
-        if (!matched[4]) {
-            result = result.concat('#master');
+        if (!branchName) {
+            branchName = '#master';
         }
 
-        return result.toLowerCase();
+        // Do not convert branch name to lowercase
+        result = result.split('#')[0].toLowerCase().concat(branchName);
+
+        return result;
     }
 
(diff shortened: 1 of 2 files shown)


lib/pipeline.js
```diff
@@ -82,13 +82,17 @@ class PipelineModel extends BaseModel {
     formatScmUrl(scmUrl) {
         let result = scmUrl;
         const matched = (schema.config.regex.SCM_URL).exec(result);
+        let branchName = matched[4];
```
FenrirUnbound: I'm not a terribly big fan of magic numbers. It isn't your fault (since this is how it was before), but could you define this `4` as constant?

FenrirUnbound:
👍 

Looks good

lf04obq

lib/pipeline.js
catto: Okay, I'll fix it.
Should I define that constant at `screwdriver-data-schema` module?

FenrirUnbound: At the top of this file should be ok

FenrirUnbound: Like a 

```
const MATCH_COMPONENT_BRANCH_NAME = 4;
```

catto: Done!

FenrirUnbound:
👍 Still good to me 😸

stjohnjohnson:
LGTM 👍 
excellent

PR #41, opened 2016-08-09
[F1] Pass new arguments to Executor (4th)

FenrirUnbound:
Resolving the merge conflicts with PR #32.

[Feature](https://github.com/screwdriver-cd/screwdriver/issues/77)

Passing:
- buildId - Build Identifier (to look up stuff)
- container - Container to start in
- apiUri - URI to hit our API
- token - Token to be able to read/write from the API

Closes #32

PR #42, opened 2016-08-10
Sanitize datastore saves by valid keys

FenrirUnbound:
Currently, anything that is passed to a `.create()` will be stored in the datastore table. This PR intends to filter the incoming data, so that we only save values assigned to valid columns. In addition, this change will make sure that the Model is constructed with the correct parameters.

This will help with PR #41 in order to filter out the overloaded parameters field.

### DIFF
--- a/lib/baseFactory.js
+++ b/lib/baseFactory.js
@@ -55,14 +55,21 @@ class BaseFactory {
             table: this.table,
             params: {
                 id,
-                data: config
+                data: {}
             }
         };
 
+        // Filter for valid keys
+        this.model.allKeys.forEach((key) => {
+            if (config[key]) {
+                modelConfig.params.data[key] = config[key];
+            }
+        });
+
         return nodeify.withContext(this.datastore, 'save', [modelConfig])
-            // add datastore to create config
-            .then(modelData => this.createClass(hoek.applyToDefaults(modelData, {
-                datastore: this.datastore
+            .then(modelData => this.createClass(hoek.applyToDefaults(config, {
+                datastore: this.datastore,
+                id: modelData.id
             })));
     }
 
--- a/test/lib/buildFactory.test.js
+++ b/test/lib/buildFactory.test.js
@@ -115,7 +115,6 @@ describe('Build Factory', () => {
                     createTime: dateNow,
                     number: dateNow,
                     status: 'QUEUED',
-                    username,
                     jobId,
                     sha
                 }
@@ -147,6 +146,17 @@ describe('Build Factory', () => {
                 assert.deepEqual(model.executor, executor);
                 assert.isFalse(jobFactoryMock.get.called);
                 assert.isFalse(userFactoryMock.get.called);
+            });
+        });
+
+        it('ignores extraneous parameters', () => {
+            const expected = {};
+            const garbage = 'garbageData';
+
+            datastore.save.yieldsAsync(null, expected);
+
+            return factory.create({ garbage, username, jobId, sha }).then(() => {
+                assert.calledWith(datastore.save, saveConfig);
             });
         });
 


stjohnjohnson:
LGTM 👍

lib/baseFactory.js
```diff
@@ -55,13 +55,21 @@ class BaseFactory {
...
+                modelConfig.params.data[key] = config[key];
+            }
+        });
+
         return nodeify.withContext(this.datastore, 'save', [modelConfig])
-            .then(modelData => this.createClass(hoek.applyToDefaults(modelData, {
-                datastore: this.datastore
+            .then(modelData => this.createClass(hoek.applyToDefaults(config, {
+                datastore: this.datastore,
+                id: modelData.id
```
d2lam, line 72: Huh? I don't get this. Why need id now?

lib/baseFactory.js
```diff
@@ -55,13 +55,21 @@ class BaseFactory {
...
+        this.model.allKeys.forEach((key) => {
+            if (config[key]) {
+                modelConfig.params.data[key] = config[key];
+            }
+        });
+
         return nodeify.withContext(this.datastore, 'save', [modelConfig])
-            .then(modelData => this.createClass(hoek.applyToDefaults(modelData, {
-                datastore: this.datastore
+            .then(modelData => this.createClass(hoek.applyToDefaults(config, {
```
nkatzman, line 70: why change this from `modelData` to `config` ? 

The base class will do essentially the same thing being done on lines 63 to 67, so you don't need all those extra "config" parameters anyway

https://github.com/screwdriver-cd/models/blob/master/lib/base.js#L19-L21

nkatzman, line 70: Oh except I see. The point of this PR is not just to sanitize datastore saves, but really to be able to call `this.createClass` with extra parameters

test/lib/buildFactory.test.js
```diff
@@ -143,6 +142,17 @@ describe('Build Factory', () => {
...
         });
 
+        it('ignores extraneous parameters', () => {
+            const expected = {};
+            const garbage = 'garbageData';
+
+            datastore.save.yieldsAsync(null, expected);
+
+            return factory.create({ garbage, username, jobId, sha }).then(() => {
+                assert.calledWith(datastore.save, saveConfig);
```
nkatzman, line 159: should also assert that build.create is called with garbage if thats an intended change you want to test

lib/baseFactory.js
FenrirUnbound, line 70: Right. The `config` argument is overloaded with elements to save in the datastore, as well as elements that will be utilized the construct the model. I've updated the PR description to reflect this.

lib/baseFactory.js
FenrirUnbound, line 72: The factories ask for it. Examples:
- [build](https://github.com/screwdriver-cd/models/blob/master/lib/buildFactory.js)
- [job](https://github.com/screwdriver-cd/models/blob/master/lib/jobFactory.js#L23)
- [pipeline](https://github.com/screwdriver-cd/models/blob/master/lib/pipelineFactory.js#L24)

test/lib/buildFactory.test.js
FenrirUnbound, line 159: That case was omitted. The datastore is sensitive towards what data it gets, so the intent is to only validate that the `datastore.save` is called with the correct parameters.

nkatzman:
👍

test/lib/buildFactory.test.js
petey, line 159: I kinda agree with Darren here. Class creation is treated in this test as an external run-time dependency (e.g. async), and should ultimately be tested there. One could argue that there should be a test for `base.constructor` that ensures that garbage keys are not set on the model when they are not part of the schema or otherwise expected by that class.

d2lam:
👍

PR #43, opened 2016-08-11
buildFactory handles executor in construction

petey:
Needed before: https://github.com/screwdriver-cd/screwdriver/pull/88
- The BuildFactory needs to handle adding executor to Build models.
- All factories should have a more robust getInstance method.
- `pipeline.formatScmUrl()` isn't actually used by anything, so it has been removed

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -1,8 +1,6 @@
 'use strict';
 const BaseModel = require('./base');
 const JobFactory = require('./jobFactory');
-const schema = require('screwdriver-data-schema');
-const MATCH_COMPONENT_BRANCH_NAME = 4;
 
 class PipelineModel extends BaseModel {
     /**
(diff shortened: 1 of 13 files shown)


lib/pipeline.js
```diff
@@ -37,28 +35,6 @@ class PipelineModel extends BaseModel {
     }
 
     /**
-     * Format the scm url to include a branch and make case insensitive
-     * @method formatScmUrl
-     * @param  {String}     scmUrl Github scm url
-     * @return {String}            Lowercase scm url with branch name
-     */
-    formatScmUrl(scmUrl) {
```
tkyi, line 45: this method is actually used in the API
https://github.com/screwdriver-cd/screwdriver/blob/7756636b22ebe84600f4f5e591bbe3c49d1e5208/plugins/pipelines/create.js#L22

tkyi, line 45: Nevermind, I see that you moved it in your other pull request
https://github.com/screwdriver-cd/screwdriver/pull/88

petey, line 45: Yeah, it was used statically in one place in the api in a place where it makes no sense to instantiate a new pipeline just to use this function. I moved the function to the appropriate place in the API. If it REALLY needs to be used elsewhere, we can make a helper module for it.

jer, line 45: Looks like it was moved there: https://github.com/screwdriver-cd/screwdriver/pull/88/files#diff-1eb6375f360812e82460d42f926123b1R14

jer:
👍  ogbvdqu - imgur

tkyi:
👍 
giphy-2

test/lib/buildFactory.test.js
```diff
@@ -245,10 +252,25 @@ describe('Build Factory', () => {
...
+            const f2 = BuildFactory.getInstance({ datastore, executor });
 
             assert.equal(f1, f2);
         });
+
+        it('should not require config on second call', () => {
+            const f1 = BuildFactory.getInstance({ datastore, executor });
+            const f2 = BuildFactory.getInstance();
+
+            assert.equal(f1, f2);
```
nkatzman, line 265: is this sufficient to check that two instances of build are the same?

petey, line 265: This would verify the references of the object are identical, as opposed to the content, so it should be enough.

PR #44, opened 2016-08-11
Fix call to get pipeline admin

tkyi:
The method `getAdmin` was changed to `get admin` and the call to get admins was never updated.
Fixing bug here.

https://github.com/screwdriver-cd/models/blob/f4fbf01f9d5f048d4f44feee602610023c23b01a/lib/pipeline.js#L43


FenrirUnbound:
👍 

Looks fine to me

fpdkh8g - imgur

petey:
👍 
cats_lol

PR #47, opened 2016-08-12
Fix circular dependency issues

petey:
- Lazy load factory dependencies to prevent circular dependency issues
  https://nodejs.org/api/modules.html#modules_cycles
- Utilize getInstance method
- Handle datastore misses when fetching data
- UserFactory should handle passing around the password for encryption


jer:
👍 I'm glad you figured that one out. That's tricky.

Filbird:
👍 Makes sense

nkatzman:
👍  looks fine

PR #50, opened 2016-08-12
Base model tracks changed fields

petey:
- Store the rowData passed to the constructor in a private field and dynamically generate getters and setters for class members.
- Save dirty bits for fields changed with setters.
- Update only those fields that have changed since the model was constructed, or the last update operation was run


nkatzman:
👍

jer:
👍

PR #60, opened 2016-08-12
Use repo's owner username in createstatus

d2lam:
The github API requires the Repository's user instead of the build user when updating the status.

```
github.com/<user>/<repo>
```

is the target for the update

### DIFF
--- a/test/lib/build.test.js
+++ b/test/lib/build.test.js
@@ -173,7 +173,6 @@ describe('Build Model', () => {
     describe('start', () => {
         let sandbox;
         let tokenGen;
-        const user = { username: 'me' };
         const adminUser = { username: 'batman' };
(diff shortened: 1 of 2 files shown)


FenrirUnbound:
Why do we want to use the repository owner's username instead? ( 😉  description 😉 )

test/lib/build.test.js
```diff
@@ -173,7 +173,6 @@ describe('Build Model', () => {
     describe('start', () => {
         let sandbox;
         let tokenGen;
-        const user = { username: 'me' };
```
FenrirUnbound, line 176: Minor, since it's just a test. Why not keep the `const user` and define it as a `screwdriver-cd` here?

FenrirUnbound:
👍 

Seems fine to me 

6hl7a6o - imgur

jer:
👍

PR #61, opened 2016-08-13
Make non-schema properties private

petey:
This makes all properties that are not part of the model schema more "private". This prevents accidental exposure of datastore, executor, and encryption password to objects that have a handle to various models. This causes things like Object.getOwnPropertyNames() to only expose properties that we want to be exposed; usually fields from the model schema.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -3,6 +3,9 @@ const BaseModel = require('./base');
 const nodeify = require('./nodeify');
 const githubHelper = require('./github');
 
+// Symbols for private members
+const executor = Symbol('executor');
+
 class BuildModel extends BaseModel {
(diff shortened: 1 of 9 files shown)


lib/build.js
```diff
@@ -3,6 +3,9 @@ const BaseModel = require('./base');
 const nodeify = require('./nodeify');
 const githubHelper = require('./github');
 
+// Symbols for private members
+const executor = Symbol('executor');
```
nkatzman, line 7: why `executor`?

petey, line 7: It isn't part of the schema. Username probably should be, but isn't.

jer:
👍

PR #63, opened 2016-08-15
Properly handle missing job or user model in build.create

petey:
When the datastore returns null when looking for a job, build.create throws a runtime error. This should catch the error and return something better.

Also added proper tests for build.job and build.user.


d2lam:
LGTM 👍

jer:
👍  Seems like a good thing to do.

When does this happen though? A build references a job that doesn't exist?

petey:
@jer Yes, I was able to replicate this with one of the synchronize payloads from github.

PR #67, opened 2016-08-16
Github status should be pending when build is running/queued

d2lam:
Right now the status is failure when the build is queued/running, which is not correct.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -229,12 +229,20 @@ class BuildModel extends BaseModel {
                     .then(pipeline =>
                         pipeline.admin
                         // update github
-                            .then(admin => updateSCM({
-                                user: admin,
-                                scmUrl: pipeline.scmUrl,
-                                sha: this.sha,
-                                state: (this.status === 'SUCCESS') ? 'success' : 'failure'
-                            }))
+                            .then(admin => {
+                                const stateMap = {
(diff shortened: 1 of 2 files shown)


lib/build.js
```diff
@@ -229,12 +229,20 @@ class BuildModel extends BaseModel {
...
                         pipeline.admin
                         // update github
-                            .then(admin => updateSCM({
-                                user: admin,
-                                scmUrl: pipeline.scmUrl,
-                                sha: this.sha,
-                                state: (this.status === 'SUCCESS') ? 'success' : 'failure'
-                            }))
+                            .then(admin => {
+                                const stateMap = {
```
FenrirUnbound, line 233: This is basically a map of constants. Define them at the top of the file

d2lam, line 233: This will probably be removed after the git scm is done anyway

FenrirUnbound, line 233: "Probably"? Or "Will"?

d2lam, line 233: will. that's what Noah is working on

FenrirUnbound, line 233: 👍 👌

FenrirUnbound:
👍 

Looks good to me

uxjywqc

nkatzman:
👍

PR #70, opened 2016-08-17
scmPlugin is a requirement for factories

petey:
- Forces factories to require an scmPlugin for `getInstance`
- Refactor `getInstance` methods so there is less duplicated code
- Base model exposes scmPlugin as `Model.scm`
- Refactor BaseFactory tests so as not to depend on a specific table's schema

### DIFF
--- a/lib/baseFactory.js
+++ b/lib/baseFactory.js
@@ -11,11 +11,13 @@ class BaseFactory {
      * @param  {String}    modelName            Name of the model to get from data-schema
      * @param  {Object}    config
      * @param  {Object}    config.datastore     Object that will perform operations on the datastore
+     * @param  {Object}    config.scmPlugin     Object that will perform operations on scm resource
      */
     constructor(modelName, config) {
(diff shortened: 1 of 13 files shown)


lib/baseFactory.js
```diff
@@ -145,6 +148,36 @@ class BaseFactory {
...
+     * @param  {Object}     instance            The current instance of a factory
+     * @param  {Object}     config
+     * @param  {Datastore}  config.datastore    A datastore instance
+     * @param  {Datastore}  config.scmPlugin    A scm plugin instance
+     * @return {Factory}
+     */
+    static getInstance(ClassDef, instance, config) {
+        let inst = instance;
+
+        if (!instance) {
```
nkatzman: Minor thing, but should this be `!inst` just for consistency of only using inst vs instance?

d2lam:
Looks good! 👍

nkatzman:
👍

FenrirUnbound:
👍

PR #77, opened 2016-08-19
Implement pipeline sync (update & disable jobs)

d2lam:
Update and disable jobs in sync when necessary

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -1,3 +1,4 @@
+/* eslint no-param-reassign: ["error", { "props": false }] */
 'use strict';
 const BaseModel = require('./base');
 const parser = require('screwdriver-config-parser');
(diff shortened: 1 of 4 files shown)


lib/pipeline.js
```diff
@@ -1,3 +1,4 @@
+/* eslint no-param-reassign: ["error", { "props": false }] */
```
stjohnjohnson, line 1: Why are we disabling this just in this file?

lib/pipeline.js
```diff
@@ -47,24 +47,44 @@ class PipelineModel extends BaseModel {
...
+                    const parsedConfigJobNames = Object.keys(parsedConfig.jobs);
+
+                    // Loop through all existing jobs
+                    existingJobs.forEach(job => {
+                        // if it's in the yaml, update it
+                        if (parsedConfigJobNames.includes(job.name)) {
+                            job.containers = parsedConfig.jobs[job.name].map(j => j.image);
+                            jobsToSync.push(job.update());
+                        // if it's not in the yaml, disable it
+                        } else if (!/^PR\-/.test(job.name)) {
```
stjohnjohnson: We should just expose a method in `Job` called `.isPR()`

d2lam: We could

lib/pipeline.js
d2lam, line 1: This is because of line 60 and 64, to update the job. We don't do this for other files.

d2lam:
@stjohnjohnson  Added .isPR() function to Job model

lib/pipeline.js
stjohnjohnson, line 1: I'd LOVE to be able to do this, but it's in our eslint rules.  If we want to do this, let's remove that rule globally, not per file.

stjohnjohnson:
👍 
mrw_i_dodged_yet_another_pre-ordered_video_game_disaster

PR #81, opened 2016-08-22
Remove github helper in favor of scmPlugin

petey:
- Now uses scmPlugin instead of built in github specific code
- Refactored tests calling `new Promise`
- Added support for adding build log link on create build


d2lam:
Don't need this dependency anymore https://github.com/screwdriver-cd/models/blob/master/package.json#L45

lib/build.js
```diff
@@ -192,22 +195,19 @@ class BuildModel extends BaseModel {
...
                 // update github
-                .then(admin =>
-                    updateSCM({
-                        user: admin,
-                        scmUrl: pipeline.scmUrl,
-                        sha: this.sha,
-                        state: 'failure'
-                    })
-                )
+                .then(() => this.updateCommitStatus(pipeline))
```
nkatzman: When the launcher notifies the builds webhook, it should also be updating the commitStatus.

Is there a need to do it here as well?

petey: I thought about that, and originally had it work that way, but I figured there wasn't a real good reason to do it twice, so took the update call out.

nkatzman:
👍

tkyi:
👍 
giphy-2

PR #84, opened 2016-08-23
Pass apiUri and tokenGen to build model

d2lam:
Currently, `apiUri` is only available on build `start`, so we can set the `target_url` for github status when we build starts. When the build finishes, we don't have `apiUri` passed in. 

This PR pass in `apiUri` & `tokenGen` to build model so that it's available everywhere.

### DIFF
--- a/lib/buildFactory.js
+++ b/lib/buildFactory.js
@@ -97,8 +97,6 @@ class BuildFactory extends BaseFactory {
      * @return {Promise}
      */
     create(config) {
-        const apiUri = config.apiUri;
-        const tokenGen = config.tokenGen;
         const number = Date.now();
         const createTime = (new Date(number)).toISOString();
         const modelConfig = hoek.applyToDefaults(config, {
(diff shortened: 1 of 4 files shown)


lib/buildFactory.js
```diff
@@ -108,10 +108,6 @@ class BuildFactory extends BaseFactory {
             status: 'QUEUED'
         });
 
-        // These aren't stored
-        delete modelConfig.apiUri;
```
nkatzman, line 112: should you change how you call start on line 137?

d2lam: Fixed

test/lib/build.test.js
```diff
@@ -259,7 +255,8 @@ describe('Build Model', () => {
                         token: 'foo',
                         scmUrl,
                         sha,
-                        buildStatus: 'ABORTED'
+                        buildStatus: 'ABORTED',
+                        url: `${apiUri}/v3/builds/${buildId}/logs`
```
petey: should probably make a constant for this.

petey:
👍 
jedi kittens

lib/buildFactory.js
```diff
@@ -133,10 +127,7 @@ class BuildFactory extends BaseFactory {
 
                         return super.create(modelConfig);
                     })
-                    .then(build =>
-                        build.start({ apiUri, tokenGen })
-                            .then(() => build)
-                    );
+                    .then(build => build.start());
```
nkatzman, line 130: does `build.start()` return build?

nkatzman, line 130: oh i see it does up above 👍

nkatzman:
👍

PR #87, opened 2016-08-24
createClass returns build that has apiUri and tokenGen

d2lam:
So that when factory.get is called (for example in https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/webhooks/build.js#L32), apiUri will be available in updateCommitStatus


petey:
👍

PR #91, opened 2016-09-01
pass uiUri into build models for target_url

petey:
In order to generate the correct urls for the build page, we need the ui uri to be passed through to the build model.


stjohnjohnson:
Could be pretty good, just update the screwdriver repo to send the right field name. 👍 

speling

PR #92, opened 2016-09-01
Add S3cr3t model

d2lam:
- Add a S3cr3t model. When `create`, it will seal the s3cr3t and store into the datastore. When `get`, it will unseal the s3cr3t and return the value. 
- S3cr3t does not use `scmPlugin`, so we add a check in `getInstance` so that it won't fail if `scmPlugin` is not passed in. Also, in `list`, we don't include `scmPlugin` in the config when calling `createClass`. 
- Add a `pipeline.secrets` to fetch all the s3cr3ts that belong to a pipeline
- Update README

Feature: https://github.com/screwdriver-cd/screwdriver/issues/148

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,10 @@ npm install screwdriver-models
 ```js
 'use strict';
 const Model = require('screwdriver-models');
-const factory = Model.PipelineFactory.getInstance({ datastore });
+const factory = Model.PipelineFactory.getInstance({
+    datastore,
+    scmPlugin
+});
 const config = {
     params: {
         configUrl: 'banana'
@@ -82,7 +85,10 @@ Example:
 ```js
 'use strict';
 const Model = require('screwdriver-models');
-const factory = Model.PipelineFactory.getInstance({ datastore });
+const factory = Model.PipelineFactory.getInstance({
+    datastore,
+    scmPlugin
+});
 const scmUrl = 'git@git.corp.yahoo.com:foo/BAR.git';
 factory.get({ scmUrl }).then(model => {
     model.configUrl = 'git@git.corp.yahoo.com:foo/bar.git#master';
@@ -110,7 +116,10 @@ Example:
 ```js
 'use strict';
 const Model = require('screwdriver-models');
-const factory = Model.PipelineFactory.getInstance({ datastore });
+const factory = Model.PipelineFactory.getInstance({
+    datastore,
+    scmPlugin
+});
 const scmUrl = 'git@git.corp.yahoo.com:foo/BAR.git';
 factory.get({ scmUrl }).then(model => {
     const formattedScmUrl = model.formatScmUrl(model.scmUrl);
@@ -123,7 +132,9 @@ factory.get({ scmUrl }).then(model => {
 ```js
 'use strict';
 const Model = require('screwdriver-models');
-const factory = Model.JobFactory.getInstance({ datastore });
+const factory = Model.JobFactory.getInstance({
+    datastore
+});
 const config = {
     params: {
         pipelineId: 'aabbccdd1234'
@@ -181,7 +192,9 @@ factory.get({ pipelineId, name }).then(model => {
 ```js
 'use strict';
 const Model = require('screwdriver-models');
-const factory = Model.JobFactory.getInstance({ datastore });
+const factory = Model.JobFactory.getInstance({
+    datastore
+});
 
 factory.get(id).then(model => {
     model.name = 'hello';
@@ -200,7 +213,12 @@ model.update()
 ```js
 'use strict';
 const Model = require('screwdriver-models');
-const factory = Model.BuildFactory.getInstance({ datastore });
+const factory = Model.BuildFactory.getInstance({  
+    datastore,
+    scmPlugin,
+    executor,
+    uiUri
+});
 const config = {
     params: {
         jobId: 'aaabbccdd1234'
@@ -261,7 +279,12 @@ factory.get({ jobId, number }).then(model => {
 ```js
 'use strict';
 const Model = require('screwdriver-models');
-const factory = Model.BuildFactory.getInstance(config);
+const factory = Model.BuildFactory.getInstance({  
+    datastore,
+    scmPlugin,
+    executor,
+    uiUri
+});
 
 factory.get(id).then(model => {
     model.state = 'FAILURE';
@@ -286,7 +309,11 @@ model.stream()
 ```js
 'use strict';
 const Model = require('screwdriver-models');
-const factory = Model.UserFactory.getInstance({ datastore });
+const factory = Model.UserFactory.getInstance({
+    datastore,
+    scmPlugin,
+    password            // Password to seal/unseal user's token
+});
 const config = {
     params: {
         username: 'batman'
@@ -344,7 +371,11 @@ factory.get({ username }).then(model => {
 ```js
 'use strict';
 const Model = require('screwdriver-models');
-const factory = Model.UserFactory.getInstance({ datastore });
+const factory = Model.UserFactory.getInstance({
+    datastore,
+    scmPlugin,
+    password                    // Password to seal/unseal user's token
+});
 const config = {
     username: 'myself',
     token: 'eyJksd3',            // User's github token
@@ -390,6 +421,92 @@ model.getPermissions(scmUrl)
 | Parameter        | Type  |  Description |
 | :-------------   | :---- | :-------------|
 | scmUrl | String | The scmUrl of the repo |
+
+
+### Secret Factory
+#### Search
+```js
+'use strict';
+const Model = require('screwdriver-models');
+const factory = Model.SecretFactory.getInstance({
+    datastore,
+    password            // Password for encryption operations
+});
+const config = {
+    params: {
+        pipelineId: '2d991790bab1ac8576097ca87f170df73410b55c'
+    },
+    paginate {
+        page: 2,
+        count: 3
+    }
+}
+
+factory.list(config).then(secrets => {
+    // Do stuff with list of secrets
+});
+```
+
+| Parameter        | Type  |  Description |
+| :-------------   | :---- | :-------------|
+| config        | Object | Config Object |
+| config.paginate.page | Number | The page for pagination |
+| config.paginate.count | Number | The count for pagination |
+| config.params | Object | fields to search on |
+
+#### Create
+```js
+factory.create(config).then(model => {
+    // do stuff with secret model
+});
+```
+
+| Parameter        | Type  |  Required | Description |
+| :-------------   | :---- | :-------------|  :-------------|
+| config        | Object | Yes | Configuration Object |
+| config.pipelineId | String | Yes | Pipeline that this secret belongs to |
+| config.name | String | Yes | Secret name |
+| config.value | String | Yes | Secret value |
+| config.allowInPR | String | Yes | Flag to denote if this secret can be shown in PR builds |
+
+#### Get
+Get a secret based on id. Can pass the generatedId for the secret, or the combination of pipelineId and secret name, and the id will be determined automatically.
+```js
+factory.get(id).then(model => {
+    // do stuff with secret model
+});
+
+factory.get({ pipelineId, name }).then(model => {
+    // do stuff with secret model
+});
+```
+
+| Parameter        | Type  |  Description |
+| :-------------   | :---- | :-------------|
+| id | String | The unique ID for the build |
+| config.pipelineId | String | Pipeline that the secret belongs to |
+| config.name | String | Secret name |
+
+
+### Secret Model
+```js
+'use strict';
+const Model = require('screwdriver-models');
+const factory = Model.SecretFactory.getInstance({
+    datastore,
+    password            // Password for encryption operations
+});
+const config = {
+    pipelineId: '2d991790bab1ac8576097ca87f170df73410b55c',
+    name: 'NPM_TOKEN',
+    value: banana,
+    allowInPR: false
+}
+
+factory.create(config)
+    .then(model => // do something
+    });
+```
 
 ## Testing
 
--- a/README.md
+++ b/README.md
@@ -508,6 +508,12 @@ factory.create(config)
     });
 ```
 
+#### Update
+Update a specific secret
+```
+model.update()
+```
+
 ## Testing
 
 ```bash
--- a/lib/baseFactory.js
+++ b/lib/baseFactory.js
@@ -158,7 +158,8 @@ class BaseFactory {
      * @param  {Object}     instance            The current instance of a factory
      * @param  {Object}     config
      * @param  {Datastore}  config.datastore    A datastore instance
-     * @param  {Datastore}  config.scmPlugin    A scm plugin instance
+     * @param  {ScmPlugin}  [config.scmPlugin]  A scm plugin instance
+     * @param  {Executor}   [config.executor]   An executor instance
      * @return {Factory}
      */
     static getInstance(ClassDef, instance, config) {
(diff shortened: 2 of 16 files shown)


lib/baseFactory.js
```diff
@@ -139,12 +139,17 @@ class BaseFactory {
                         `expected Array, got ${typeof data}`);
                 }
                 const result = [];
+                const extraConfig = {
+                    datastore: this.datastore
+                };
+
+                // Only add scmPlugin if it's not a secret model
+                if (!config.secret) {
+                    extraConfig.scmPlugin = this.scmPlugin;
```
petey: You probably don't actually need to do this. It shouldn't hurt anything to pass extra parameters to the model, even if their value is `undefined`. If you decide you DO have to do this, be consistent and do it in `get` and `create`, too.

lib/secretFactory.js
```diff
@@ -0,0 +1,144 @@
...
+            .then(secret => super.create(hoek.applyToDefaults(config, { value: secret })));
+    }
+
+    /**
+     * Get a record based on id
+     * @method get
+     * @param  {Mixed}   config    The configuration from which an id is generated or the actual id
+     * @return {Promise}
+     */
+    get(config) {
```
petey: This does little more than redefine all the code in the base class. Why not utilize the base class code, and do the data wrangling with the result?

```
return super.get(config).then(secret => unseal(secret.value)).then(unsealed => {secret.value = unsealed; return secret; } );
```

lib/secretFactory.js
```diff
@@ -0,0 +1,144 @@
...
+     * @param  {Object}   config.params           Parameters to filter on
+     * @param  {Object}   config.paginate         Pagination parameters
+     * @param  {Number}   config.paginate.count   Number of items per page
+     * @param  {Number}   config.paginate.page    Specific page of the set to return
+     * @return {Promise}
+     */
+    list(config) {
+        config.secret = true;
+
+        return super.list(config);
```
petey: Isn't this forgetting to unseal the secret value?

lib/baseFactory.js
```diff
@@ -171,7 +176,8 @@ class BaseFactory {
                 throw new Error(`No datastore provided to ${className}`);
             }
 
-            if (!config || !config.scmPlugin) {
+            // SecretFactory does not use ScmPlugin
+            if (className !== 'SecretFactory' && (!config || !config.scmPlugin)) {
```
petey: I don't like the idea of tying baseFactory to a specific factory implementation. Maybe this check should get moved into the specific factory instances that actually need it. (Job doesn't actually need it either.)

lib/secretFactory.js
d2lam: ur right! 👍

lib/baseFactory.js
d2lam: Yeah we can do that too :)

d2lam:
Interesting. I don't know why screwdriver build failed at unit test saying job.isPR is not a function.

d2lam:
@stjohn was it some bug in the launcher?!?

stjohnjohnson:
Nah, it was a bug in one of the versions of data-schema.  It's fixed now.

lib/pipeline.js
```diff
@@ -147,6 +147,42 @@ class PipelineModel extends BaseModel {
...
+     * @property secrets
+     * @return Promise
+    */
+    get secrets() {
+        const listConfig = {
+            params: {
+                pipelineId: this.id
+            },
+            paginate: {
+                count: 25, // This limit is set by the matrix restriction
```
stjohnjohnson: Why would this be set by the matrix?  There can be 100+ secrets, unless you put that limit in the config-parser.

stjohnjohnson:
LGTM otherwise.

README.md
```diff
@@ -82,7 +82,10 @@ Example:
 ```js
 'use strict';
 const Model = require('screwdriver-models');
-const factory = Model.PipelineFactory.getInstance({ datastore });
+const factory = Model.PipelineFactory.getInstance({
+    datastore,
+    scmPlugin,          // scm plugin.  For example, screwdriver-scm-github
```
FenrirUnbound: The example is a bit wonky. Simply stating "scm plugin" isn't very descriptive. Can we just require `screwdriver-scm-github` if that's going to be your example anyways?

d2lam: I would need to add how to instantiate scm-github correctly, that's why I didn't want to do it.

lib/pipeline.js
d2lam: Oops copy pasta.

lib/secretFactory.js
```diff
@@ -0,0 +1,126 @@
...
+const hoek = require('hoek');
+// Get symbols for private fields
+const password = Symbol();
+
+let instance;
+
+/**
+ * Seal secret value and return the secret with the sealed value
+ * @method sealSecret
+ * @param  {SecretModel}        secret  Secret to seal
```
FenrirUnbound: `secret` is not a SecretModel

lib/secretFactory.js
```diff
@@ -0,0 +1,126 @@
...
+     * @param  {Object}     config
+     * @param  {String}     config.pipelineId      Pipeline Id
+     * @param  {String}     config.name            Secret name
+     * @param  {String}     config.value           Secret value
+     * @param  {Boolean}    config.allowInPR       Whether this secret can be shown in PR builds
+     * @return {Promise}
+     */
+    create(config) {
+        return sealSecret(config, this[password])
+            .then(secret => super.create(hoek.applyToDefaults(config, { value: secret.value })));
```
FenrirUnbound: I think the `hoek.applyToDefaults` is a redundant call. You already modify the `config` parameter in `sealSecret()`.

lib/userFactory.js
```diff
@@ -69,6 +69,9 @@ class UserFactory extends BaseFactory {
      * @return {UserFactory}
      */
     static getInstance(config) {
+        if (!instance && (!config || !config.scmPlugin)) {
+            throw new Error('No scm plugin provided to UserFactory');
```
FenrirUnbound, line 73: Does this mean that this is a backwards incompatible change?

lib/secretFactory.js
FenrirUnbound: Or at least not based on how you're using it: https://github.com/screwdriver-cd/models/pull/92/files#diff-db62f6a8d7d20845993eb9a470a318dfR81

test/lib/pipeline.test.js
```diff
@@ -299,6 +305,32 @@ describe('Pipeline Model', () => {
...
+            secretFactoryMock.list.resolves(null);
+            // when we fetch secrets it resolves to a promise
+            assert.isFunction(pipeline.secrets.then);
+            // and a factory is called to create that promise
+            assert.calledWith(secretFactoryMock.list, listConfig);
+
+            // When we call pipeline.secrets again it is still a promise
+            assert.isFunction(pipeline.secrets.then);
+            // ...but the factory was not recreated, since the promise is stored
+            // as the model's pipeline property, now
```
FenrirUnbound, line 329: I have no clue why this test is written this way. Looking at it, it seems like it can also be written as:

```
return pipeline.secrets()
    .then(() => {
        assert.calledWith(secretFactoryMock.list, listConfig);

        return pipeline.secrets();
    })
    .then(() => {
        assert.calledOnce(secretFactoryMock.list);
    });
```

Why is this test written this way?

FenrirUnbound, line 329: I looked back and saw that `pipeline.secrets` is an attribute an not a function:

```
const whatever = pipeline.secrets;

return whatever.then(() => {
    assert.calledWith(secretFactoryMock.list, listConfig);

    return pipeline.secrets;
})
.then(() => {
    assert.calledOnce(secretFactoryMock.list);
});
```

README.md
stjohnjohnson: Do we need to put the inline doc here?  We don't for datastore.

lib/secretFactory.js
stjohnjohnson: So is `secret` just an object?

lib/userFactory.js
stjohnjohnson: I don't think so, since it's removed from BaseFactory.

test/lib/pipeline.test.js
d2lam: It's a getter so it's a function. But you access it using `pipeline.secrets`

The test for this getter is written the same way as how we wrote it for other getters. For example here: https://github.com/screwdriver-cd/models/blob/6d31261ea12a0fe2df2ac721d9a074399ba64fc0/test/lib/pipeline.test.js#L260-L274

If we want to change it, we should do it in another PR where we change all of them. But imo, it tests the same thing, just different way of writing it. You can ask @petey why it was written in the previous way, maybe he has better reasons.

lib/userFactory.js
d2lam: @stjohnjohnson is right. We moved it out of BaseFactory since different factories require different things.

lib/secretFactory.js
d2lam: No. The comment is wrong here because of some refactoring & copy pasta. I should just pass in `string` here.

README.md
d2lam: Don't need to but I thought it would be more clear. We can remove it.

test/lib/pipeline.test.js
petey: @FenrirUnbound, your code should work, too. The original tests didn't really care about the output of the Promise, just that the getter returned something that looked like a promise. I agree with @d2lam , if we want to change it to your way, we should change all the getter tests at once and be consistent.

test/lib/secretFactory.test.js
```diff
@@ -0,0 +1,232 @@
...
+        };
+        const datastoreReturnValue = [secret1, secret2];
+        const returnValue = [secret1, secret2];
+
+        returnValue[0].value = 'batman';
+        returnValue[1].value = 'superman';
+
+        it('calls datastore scan and returns correct values', () => {
+            datastore.scan.yieldsAsync(null, datastoreReturnValue);
+            ironMock.unseal.withArgs(secret1.value, password).yieldsAsync(null, 'batman');
```
petey: This may not make a big difference, since it is just the mock data and mock execution, but I'm not convinced by this code that the correct values are being passed to unseal. Please change `secret1.value` to `sealedsecret1value`. Likewise with with `secret2.value`.

lib/userFactory.js
FenrirUnbound: So anything that's using the `userFactory` will still work if we pull in this ~~package~~ update?

lib/secret.js
```diff
@@ -0,0 +1,24 @@
...
+     * @param  {Object}   config                Config object to create the secret with
+     * @param  {Object}   config.datastore      Object that will perform operations on the datastore
+     * @param  {String}   config.pipelineId     Pipeline Id
+     * @param  {String}   config.name           Secret name
+     * @param  {String}   config.value          Secret value
+     * @param  {Boolean}  config.allowInPR      Whether this secret can be shown in PR builds
+     */
+    constructor(config) {
+        super('secret', config);
+        this[password] = config.password;
```
petey, line 23: Why does this have a password? It is never used.

lib/userFactory.js
petey: @FenrirUnbound Yes, they just moved the scmPlugin check from base to userFactory. So the functionality is unchanged.

d2lam:
Hmmm how come screwdriver build doesn't show up?

lib/pipeline.js
```diff
@@ -147,6 +147,42 @@ class PipelineModel extends BaseModel {
...
+     * @property secrets
+     * @return Promise
+    */
+    get secrets() {
+        const listConfig = {
+            params: {
+                pipelineId: this.id
+            },
+            paginate: {
+                count: 25,
```
petey: do we want/need to paginate secrets?

lib/secretFactory.js
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /**
+     * Instantiate a Secret class
+     * @method createClass
+     * @param  config
+     * @return {Secret}
+     */
+    createClass(config) {
+        const c = hoek.applyToDefaults(config, { password: this[password] });
```
petey: Why pass password to secret constructor, it is not used.

d2lam: It was there from earlier. Forgot to remove it.

test/lib/pipeline.test.js
FenrirUnbound: It's written in a non-idiomatic manner; I had to look at it multiple times to understand what it was trying to accomplish. 

My initial comment is an educational comment and not a functional one, especially since it's a test. Tests are low priority and have been treated as such. They'll be copy and pasted throughout the next PRs too. A PR to fix all the tests to be consistent isn't valuable.

lib/userFactory.js
FenrirUnbound: Yeah, it's just a big diff. Found it [here](https://github.com/screwdriver-cd/models/pull/92/files#diff-c6dfbb25b49dd3e823a2aa3a6f797f96L174)

lib/pipeline.js
FenrirUnbound: It's also a magic number. If you keep it, can you make it a constant?

d2lam: Hmmm I can remove this but it looks like dynamo-datastore needs it as required fields:
https://github.com/screwdriver-cd/datastore-dynamodb/blob/master/index.js#L165-L166

d2lam:
@petey  @FenrirUnbound  My response for the pagination on secrets got buried:

```
Hmmm I can remove this but it looks like dynamo-datastore needs it as required fields:
https://github.com/screwdriver-cd/datastore-dynamodb/blob/master/index.js#L165-L166
```

so I kept it and use constant.

lib/pipeline.js
```diff
@@ -147,6 +149,42 @@ class PipelineModel extends BaseModel {
...
+     * @return Promise
+    */
+    get secrets() {
+        const listConfig = {
+            params: {
+                pipelineId: this.id
+            },
+            paginate: {
+                count: PAGINATE_COUNT,
+                page: PAGINATE_PAGE
```
stjohnjohnson, line 163: This is going to lock us in a corner.  Not for this PR< but the next one.  We need to iterate until we've loaded them all.

lib/secret.js
```diff
@@ -0,0 +1,21 @@
...
+     * Construct a SecretModel object
+     * @method constructor
+     * @param  {Object}   config                Config object to create the secret with
+     * @param  {Object}   config.datastore      Object that will perform operations on the datastore
+     * @param  {String}   config.pipelineId     Pipeline Id
+     * @param  {String}   config.name           Secret name
+     * @param  {String}   config.value          Secret value
+     * @param  {Boolean}  config.allowInPR      Whether this secret can be shown in PR builds
+     */
+    constructor(config) {
```
stjohnjohnson, line 21: What about save?  Will that seal the secret before saving?

petey: Ha! Yeah, that's why you needed to pass a password to this model. You need to implement an update method to seal the secret value, especially since the model you pass around will always consider `value` to be `dirty`.

lib/baseFactory.js
```diff
@@ -158,7 +158,8 @@ class BaseFactory {
      * @param  {Object}     instance            The current instance of a factory
      * @param  {Object}     config
      * @param  {Datastore}  config.datastore    A datastore instance
-     * @param  {Datastore}  config.scmPlugin    A scm plugin instance
+     * @param  {ScmPlugin}  [config.scmPlugin]  A scm plugin instance
```
petey: Remove any optional params to baseFactory. These should be documented in the factories that require them.

petey:
👍

FenrirUnbound:
👍 

SHIP IT

giphy_3456

PR #93, opened 2016-09-06
Adding isPR, jobID, and pipelineID to JWT for builds

stjohnjohnson:
This will include additional information into the JWT served to builds


d2lam:
👍

PR #95, opened 2016-09-09
Adding Secrets to Job and Build (fixing bug with falsey values)

stjohnjohnson:
This adds secrets to job and build model.  As well as fixing a bug when updating values to `false`

### DIFF
--- a/lib/baseFactory.js
+++ b/lib/baseFactory.js
@@ -63,7 +63,7 @@ class BaseFactory {
 
         // Filter for valid keys
         this.model.allKeys.forEach((key) => {
-            if (config[key]) {
+            if (config[key] !== undefined) {
                 modelConfig.params.data[key] = config[key];
             }
         });
--- a/lib/job.js
+++ b/lib/job.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const BaseModel = require('./base');
+const hoek = require('hoek');
 
 class Job extends BaseModel {
     /**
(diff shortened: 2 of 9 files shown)


petey:
👍

lib/baseFactory.js
```diff
@@ -63,7 +63,7 @@ class BaseFactory {
 
         // Filter for valid keys
         this.model.allKeys.forEach((key) => {
-            if (config[key]) {
+            if (config[key] !== undefined) {
```
d2lam, line 66: Good catch

lib/job.js
```diff
@@ -43,6 +44,38 @@ class Job extends BaseModel {
...
 
     /**
+     * Lazy load the list of secrets
+     * @property secrets
+     * @return {Promise}    Resolves to a list of secrets
+     */
+    get secrets() {
+        delete this.secrets;
+
+        const secretNames = hoek.reach(this.permutations, '0.secrets', { default: [] });
```
d2lam, line 54: what is `0.secrets`?

d2lam, line 54: Oh I see. It's looking for `this.permutations[0].secrets`
But why? Is it always going to be index 0?

stjohnjohnson: reads the first item in the array and the field secrets

d2lam:
Just some questions, the rest LGTM 👍

lib/job.js
stjohnjohnson: Ahh, that's because secrets cannot be mutated right now, so all permutations will have the same list of secrets.

d2lam: Oh I see. Makes sense 👍

PR #97, opened 2016-09-13
[DNM] Add getters for pipeline's workflow jobs and job's builds

d2lam:
- Add a job.builds, this should replace [getBuildsForJobId](https://github.com/screwdriver-cd/models/blob/master/lib/buildFactory.js#L147-L167). It's still being used by the [api](https://github.com/screwdriver-cd/screwdriver/blob/0e8edac41ee0b8fc1ddce3e93657acc12eb34175/plugins/webhooks/github.js#L80) so not removing it yet
- Add a pipeline.workflowJobs. The API will call this, and go through the jobs in the workflow, get the latest build (using job.builds above and get the first element) and construct the object that @petey  wants in the UI.

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -2,6 +2,8 @@
 
 const BaseModel = require('./base');
 const hoek = require('hoek');
+const PAGINATE_PAGE = 1;
+const PAGINATE_COUNT = 25;
 
 class Job extends BaseModel {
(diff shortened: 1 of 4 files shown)


FenrirUnbound:
👍 

Looks good to me

having a drink - imgur

PR #98, opened 2016-09-14
get and sort builds for a specific job

d2lam:
Add a `getBuilds()` method for job. This optionally takes in a config object that specifies pagination & sorting options. If they are not passed in, use the default config: 

```
paginate: {
   page: 1, 
   count: 50
},
sort: 'descending'
```

If a field is missing, it will use the default value for that field. Sorting is always on primary range key for now, which is `build.number`

This will be use by the API for the route `jobs/id/builds`: https://github.com/screwdriver-cd/screwdriver/pull/186

Waiting on @FenrirUnbound 's implementation of the datastore. 

Solves https://github.com/screwdriver-cd/screwdriver/issues/187

### DIFF
--- a/lib/baseFactory.js
+++ b/lib/baseFactory.js
@@ -120,6 +120,7 @@ class BaseFactory {
      * @param  {Object}   config.paginate         Pagination parameters
      * @param  {Number}   config.paginate.count   Number of items per page
      * @param  {Number}   config.paginate.page    Specific page of the set to return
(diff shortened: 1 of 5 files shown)


FenrirUnbound (approved):
Looks good to me! 

tumblr_o4msxrcmyc1r6ja9oo1_500

FenrirUnbound (approved):
👍 

Still looks good to me

stjohnjohnson (approved):

PR #99, opened 2016-09-16
[4] Get active jobs for a pipeline

d2lam:
- Only lists active jobs (archived = false)
- Lists jobs in the order they appear in workflow, then open PRs.

Solves https://github.com/screwdriver-cd/screwdriver/issues/188

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -4,6 +4,7 @@ const BaseModel = require('./base');
 const hoek = require('hoek');
 const PAGINATE_PAGE = 1;
 const PAGINATE_COUNT = 50;
+const START_INDEX = 3;
 
 class Job extends BaseModel {
     /**
@@ -83,6 +84,14 @@ class Job extends BaseModel {
      */
     isPR() {
         return /^PR\-/.test(this.name);
+    }
+
+    /**
+     * Return PR number. Returns null if this is not a PR job
+     * @return {Number}     PR number
+     */
+    get prNum() {
+        return this.isPR() ? parseInt(this.name.slice(START_INDEX), 10) : null;
     }
 
     /**
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -3,8 +3,10 @@
 const BaseModel = require('./base');
 const parser = require('screwdriver-config-parser');
 const nodeify = require('./nodeify');
+const hoek = require('hoek');
 const PAGINATE_PAGE = 1;
-const PAGINATE_COUNT = 25;
+const PAGINATE_COUNT = 100;
+const START_INDEX = 3;
 
 class PipelineModel extends BaseModel {
     /**
@@ -129,7 +131,7 @@ class PipelineModel extends BaseModel {
                 pipelineId: this.id
             },
             paginate: {
-                count: PAGINATE_COUNT, // This limit is set by the matrix restriction
+                count: PAGINATE_COUNT,
                 page: PAGINATE_PAGE
             }
         };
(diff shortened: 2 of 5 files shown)


lib/pipeline.js
```diff
@@ -191,6 +192,54 @@ class PipelineModel extends BaseModel {
...
+            // sort them by the order that they appear in workflow
+            workflowJobs = workflowJobs.sort((job1, job2) =>
+                workflow.indexOf(job1.name) - workflow.indexOf(job2.name));
+
+            // get PR jobs
+            let prJobs = jobs.filter(job => job.isPR());
+
+            // sort them by name. All PR jobs look like "PR-*"
+            prJobs = prJobs.sort((job1, job2) =>
+                parseInt(job1.name.slice(START_INDEX), 10) -
```
stjohnjohnson: Why not make a method `job.PRnum` or something.

lib/pipeline.js
```diff
@@ -191,6 +192,54 @@ class PipelineModel extends BaseModel {
...
+     * @return {[type]} [description]
+     */
+    getActiveJobs() {
+        const listConfig = {
+            params: {
+                pipelineId: this.id,
+                archived: false
+            },
+            paginate: {
+                count: PAGINATE_COUNT, // This limit is set by the matrix restriction
```
stjohnjohnson: This is not correct, there is no 25 limit when you are asking for jobs + PRs.  You need to either ask for 100+ or iterate on this.

lib/pipeline.js
```diff
@@ -191,6 +192,54 @@ class PipelineModel extends BaseModel {
     }
 
     /**
+     * Fetch all the active jobs that belong to this pipeline
+     * Sorted by workflow order, followed by PR jobs
+     * @return {[type]} [description]
```
stjohnjohnson: Please fill in

lib/pipeline.js
```diff
@@ -191,6 +192,54 @@ class PipelineModel extends BaseModel {
...
+     */
+    getActiveJobs() {
```
stjohnjohnson: Why method vs. getter?  Are you passing info into this?

stjohnjohnson (changes requested):

stjohnjohnson:
You should also update line 62 to archive jobs.  And what about where PRs get archived?

lib/pipeline.js
d2lam: Nothing get passed in so yes I can make it a getter.
Or we can make this take in archived = true or false and pagination. 

What would you prefer?

lib/pipeline.js
d2lam: Sorry this is another copy pasta.

lib/pipeline.js
d2lam: Good idea. I can do that.

d2lam:
@stjohnjohnson  This PR is just for getting active jobs. I was going to have the actual archiving in a separate PR.

lib/pipeline.js
stjohnjohnson: i like option 2, combine that with what we did for `jobs.getBuilds` maybe

d2lam: What do you mean combine with `jobs.getBuilds`?

d2lam:
@stjohnjohnson  Addressed your comments:
- Added a function `job.prNum()`
- Allow `getJobs` to take in a `config` param. It get jobs with `archived = false` by default. 
- And idk what you mean by this: "combine that with what we did for jobs.getBuilds maybe"

lib/job.js
```diff
@@ -86,6 +87,18 @@ class Job extends BaseModel {
     }
 
     /**
+     * Return PR number. Returns -1 if this is not a PR job
+     * @return {Number}     PR number
+     */
+    prNum() {
+        if (!this.isPR()) {
+            return -1;
```
stjohnjohnson: Maybe return `null`?

stjohnjohnson: Can we make this into a getter that is stored on first access?  Otherwise the sort function below will be more expensive then it needs to be.

stjohnjohnson (changes requested):
Just one more piece of feedback

lib/job.js
d2lam: Hmmm I didn't want it to be a getter because not every job is a PR, so it's weird that every job has a get PR number. But it's fine.

d2lam: @stjohnjohnson  I addressed this

stjohnjohnson (approved):
LGTM

lib/job.js
```diff
@@ -86,6 +87,30 @@ class Job extends BaseModel {
     }
 
     /**
+     * Return PR number. Returns null if this is not a PR job
+     * @return {Number}     PR number
+     */
+    get prNum() {
+        delete this.prNum;
```
petey: Why is this so heavy-handed?

The logic in the getter doesn't seem that complicated that it needs to be memoized.

```
return this.isPR() ? parseInt(this.name.slice(START_INDEX), 10) : null;
```

d2lam: @petey addressed.

stjohnjohnson (approved):
Suuure

petey:
👍  
fetch_me_my_hat_jeeves_

PR #102, opened 2016-09-20
support keys inside objects

catto:
I changed the baseFactory so that we can use `myObject.id` style keys in `data-schema`. screwdriver-cd/screwdriver#194

If we specify `myObject.id` to `keys` in a model object, it passes `{'myObject.id': value}` to the hasher.


stjohnjohnson (approved):
So awesome!! 👍

petey (approved):
👍 

200w-1

PR #104, opened 2016-09-21
[prom-4] Switching to Promises for Executor (removing unused code)

stjohnjohnson:
Depends on https://github.com/screwdriver-cd/executor-k8s/pull/37

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "screwdriver-models",
-  "version": "10.4.0",
+  "version": "11.0.0",
   "description": "Screwdriver models",
   "main": "index.js",
   "scripts": {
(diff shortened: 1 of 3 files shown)


d2lam (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "screwdriver-models",
-  "version": "10.4.0",
+  "version": "11.0.0",
```
FenrirUnbound, line 3: Why the major version bump? Modules that consume `models` will still interact with the same API, and the API parameters have not changed

stjohnjohnson: Because it needs a new executor version passed to it.  I can't declare that as a peer dep.

minzcmu (approved):

minzcmu (approved):
promise

PR #106, opened 2016-09-21
[stop:1] Add build stop on done and refresh SCM repo on pipeline save

stjohnjohnson:
Couple changes:
- `build.stop` just stops the job now (doesn't update the build or github status)
- `build.update` will stop the job if the build is done (helps cleanup old resources)
- `pipeline.sync` will update the `scmRepo` object

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -214,38 +214,43 @@ class BuildModel extends BaseModel {
      * @return {Promise}
      */
     update() {
-        const dirty = this.isDirty('status');
+        let prom = Promise.resolve();
 
-        return super.update()
-            .then(() => {
-                if (!dirty) {
-                    return this;
-                }
+        // check if the status is changing
+        if (this.isDirty('status')) {
+            // stop the build if we're done
+            if (this.isDone()) {
+                prom = prom
+                    .then(() => this.stop());
+            }
+            // update scm with status
+            prom = prom
+                .then(() => this.pipeline)
+                .then(pipeline => this.updateCommitStatus(pipeline));
+        }
 
-                // update scm with status
-                return this.pipeline
-                    .then(pipeline => this.updateCommitStatus(pipeline))
-                    .then(() => this);
-            });
+        return prom
+            .then(() => super.update())
+            .then(() => this);
     }
 
     /**
-     * Stop a build and update github status as failure
+     * Stop a build
      * @method stop
      * @return {Promise}
      */
     stop() {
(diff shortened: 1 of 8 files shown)


lib/build.js
```diff
@@ -214,38 +214,43 @@ class BuildModel extends BaseModel {
...
+                .then(() => this.pipeline)
+                .then(pipeline => this.updateCommitStatus(pipeline));
+        }
 
-                // update scm with status
-                return this.pipeline
-                    .then(pipeline => this.updateCommitStatus(pipeline))
-                    .then(() => this);
-            });
+        return prom
```
d2lam, line 232: You need `prom` so that it only returns after the above promises resolve?

lib/build.js
```diff
@@ -214,38 +214,43 @@ class BuildModel extends BaseModel {
...
+            .then(() => super.update())
+            .then(() => this);
     }
 
     /**
-     * Stop a build and update github status as failure
+     * Stop a build
```
d2lam, line 238: What is the reason for not updating github status anymore?

lib/build.js
```diff
@@ -214,38 +214,43 @@ class BuildModel extends BaseModel {
...
-
-        // stop the build
         return this[executor].stop({ buildId: this.id })
-            // update status
-            .then(() => this.update());
+            .then(() => this);
+    }
+
+    /**
+     * Check if a build is doesntMatter
```
d2lam: doesntMatter?

d2lam (changes requested):

lib/build.js
stjohnjohnson: So that stop just stops the build, update is the funnel for updating github status.

lib/build.js
stjohnjohnson: I need it so I can conditionally chain events.

lib/pipeline.js
```diff
@@ -272,6 +274,34 @@ class PipelineModel extends BaseModel {
...
     }
+
+    /**
+     * Update the SCM Repo based on the SCM URL
+     * @method refreshScmRepo
+     * @return {Promise}
+     */
+    refreshScmRepo() {
+        // Skip if no changes
+        if (!this.isDirty('scmUrl')) {
```
d2lam: Wouldn't this skip the very first time?

d2lam (changes requested):

d2lam (approved):
👍  LGTM

Issue #105, opened 2016-09-21
Delete a record


d2lam:
- [x] - Delete a pipeline: should delete all the jobs and builds belong to this pipeline
- [x] - Delete a job: should delete all the builds belong to this job
- [x] - Delete a build
- [x] - Delete a secret
- [ ] - Delete a user: should go through all the pipelines and revoke admin access 

This is all I can think of right now. Please feel free to add more requirements/acceptance criteria.

d2lam:
@stjohnjohnson  So I was just thinking about whether we should allow deleting a build. Would that mess up our system (maybe not now but in the future)? And would a user ever want to delete a build? Deleting a job and deleting a pipeline make sense, but I'm not too sure about the deleting build use case.

stjohnjohnson:
Don't need to delete user right now.  Everything else is done.

PR #107, opened 2016-09-21
[remove:3] Add remove function for baseFactory

d2lam:
Needs https://github.com/screwdriver-cd/datastore-dynamodb/pulls

Related to https://github.com/screwdriver-cd/models/issues/105

### DIFF
--- a/lib/baseFactory.js
+++ b/lib/baseFactory.js
@@ -1,7 +1,6 @@
 'use strict';
 const schema = require('screwdriver-data-schema');
 const hashr = require('screwdriver-hashr');
-const nodeify = require('./nodeify');
 const hoek = require('hoek');
 
 class BaseFactory {
@@ -68,7 +67,7 @@ class BaseFactory {
             }
         });
 
-        return nodeify.withContext(this.datastore, 'save', [modelConfig])
+        return this.datastore.save(modelConfig)
             .then(modelData => this.createClass(hoek.applyToDefaults(config, {
                 datastore: this.datastore,
                 scmPlugin: this.scmPlugin,
@@ -98,7 +97,7 @@ class BaseFactory {
             }
         };
 
-        return nodeify.withContext(this.datastore, 'get', [lookup])
+        return this.datastore.get(lookup)
             .then(data => {
                 // datastore miss, can not applyToDefaults on null
                 if (!data) {
@@ -137,7 +136,7 @@ class BaseFactory {
             scanConfig.sort = config.sort;
         }
 
-        return nodeify.withContext(this.datastore, 'scan', [scanConfig])
+        return this.datastore.scan(scanConfig)
             .then(data => {
                 if (!Array.isArray(data)) {
                     throw new Error('Unexpected response from datastore, ' +
--- a/test/lib/pipelineFactory.test.js
+++ b/test/lib/pipelineFactory.test.js
@@ -110,7 +110,7 @@ describe('Pipeline Factory', () => {
                 scmUrl
             };
 
-            datastore.save.yieldsAsync(null, expected);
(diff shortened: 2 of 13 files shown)


stjohnjohnson (approved):
Yeah!! Promises!

lib/baseFactory.js
```diff
@@ -113,6 +112,31 @@ class BaseFactory {
     }
 
     /**
+     * Remove a Model
+     * @method remove
+     * @param  {Object}    config               Config object
```
minzcmu: The type of `config` should be `mixed`? Since config can be just a string of id instead of an object.

d2lam: Fixed

test/lib/pipelineFactory.test.js
```diff
@@ -85,8 +85,7 @@ describe('Pipeline Factory', () => {
                 data: {
                     admins,
                     createTime: nowTime,
-                    scmUrl,
-                    configUrl: scmUrl
+                    scmUrl
```
minzcmu, line 87: Should also remove `configUrl` here
https://github.com/screwdriver-cd/models/blob/9225a81c0cd949d9f80964a16bbc43ed10626ef8/test/lib/pipelineFactory.test.js#L53

d2lam: This was fixed from St.John's previous PR. I'm curious why it didn't show merge conflicts. 
But I rebased and it automatically fixed itself. :)

d2lam:
Just realized there are some other places that I need to update to promises.

lib/baseFactory.js
```diff
@@ -113,6 +112,31 @@ class BaseFactory {
     }
 
     /**
+     * Remove a Model
+     * @method remove
+     * @param  {Mixed}    config               Config object
+     * @return {Promise}
+     */
+    remove(config) {
```
FenrirUnbound: This just removes a row by id? Does it not do a clean-up of everything related to it?

FenrirUnbound: And why attach to a factory? Wouldn't it be easier just to attach the `remove()` function to each individual model?

d2lam: @FenrirUnbound Yup this removes a row by id. So this will work for models that don't have any child dependencies. For deleting things like pipeline or job, they need to take advantage of this base `remove` and do extra things. Like how we did for `create` for example: https://github.com/screwdriver-cd/models/blob/master/lib/pipelineFactory.js#L41-L46

I put it in a factory because we have a `create` in here so I think it makes sense to have `remove` in here as well. Also it will be nice to be able to pass in `config` where it could be an `object` contains the key (like in `get`) and not having to get the secret first. Either place is fine with me.

FenrirUnbound: Right, this is good for when the model has no child relationships 👍 

Using your example of pipelines, the child relationships are attached to the object (i.e., [jobs](https://github.com/screwdriver-cd/models/blob/master/lib/pipeline.js#L130)). When constructing a model, it's not easy to construct an object in an asynchronous fashion. The factories help address that concern by using them as an intermediary while the respective model is loaded with data asynchronously. 

Once the model is loaded with data, we shouldn't need the factory for anything else; the remaining CRUD operations can be handled by the models themselves.

d2lam: @FenrirUnbound  I moved it to model

d2lam: Please squash/merge for me when you merge it 😄

stjohnjohnson (approved):
Looks good!

FenrirUnbound (approved):
👍 

Looks good!

a diglett doing a twirl in pokemon go x-post from rpokemongo - imgur

PR #108, opened 2016-09-23
Skip unseal if secret does not exist

d2lam:
This is a bug. If secret does not exist, we should skip the unsealing of secret. Otherwise, it will try to read `secret.value` where secret is `null`

### DIFF
--- a/test/lib/secretFactory.test.js
+++ b/test/lib/secretFactory.test.js
@@ -163,6 +163,22 @@ describe('Secret Factory', () => {
                     });
                 })
         );
+
+        it('skip unseal when secret not found', () => {
+            datastore.get.withArgs({
+                table: 'secrets',
+                params: {
+                    id
+                }
+            }).resolves(null);
+
+            return factory.get({ pipelineId, name })
+                .then(model => {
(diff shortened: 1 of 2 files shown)


test/lib/secretFactory.test.js
```diff
@@ -163,6 +163,22 @@ describe('Secret Factory', () => {
...
+                params: {
+                    id
+                }
+            }).resolves(null);
+
+            return factory.get({ pipelineId, name })
+                .then(model => {
+                    assert.isTrue(datastore.get.calledOnce);
+                    assert.notCalled(ironMock.unseal);
+                    assert.isNull(model);
```
FenrirUnbound, line 179: That's an odd assertion. Are you checking for `null` because your mock returns null? 

If you're looking to see if data came back from your mock, it would be more deterministic if you had your mock return something that only could come back from the mock.

d2lam:
I mocked the datastore, but assert what's get back from factory, which should be null.

FenrirUnbound (approved):
👍 

Looks fine to me

myspk5nqr5nx

tkyi (approved):
👍 LGTM

PR #109, opened 2016-09-24
[1] Remove pipeline and job

d2lam:
- Removing pipeline removes all the jobs and builds associated with it. 
- Removing job removes all the builds associated with it. 

Relates to https://github.com/screwdriver-cd/models/issues/105

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -131,6 +131,27 @@ class Job extends BaseModel {
 
         return factory.list(listConfig);
     }
+
+    /**
+     * Remove all builds associated with this job and the job itself
+     * @return {Promise}        Resolves to null if remove successfully
+     */
+    remove() {
+        const removeBuilds = (() =>
+            // Iterate through the builds and remove them
+            this.getBuilds().then(builds => {
+                if (builds.length === 0) {
+                    // Done removing builds
+                    return null;
+                }
+
+                return Promise.all(builds.map(build => build.remove()))
+                    .then(() => removeBuilds());
+            }));
+
+        // Remove the job
+        return removeBuilds().then(() => super.remove());
+    }
 }
 
 module.exports = Job;
--- a/test/lib/job.test.js
+++ b/test/lib/job.test.js
@@ -3,6 +3,7 @@ const assert = require('chai').assert;
 const mockery = require('mockery');
 const sinon = require('sinon');
 const schema = require('screwdriver-data-schema');
+const hoek = require('hoek');
 
 sinon.assert.expose(assert, { prefix: '' });
 require('sinon-as-promised');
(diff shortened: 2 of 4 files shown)


lib/job.js
```diff
@@ -131,6 +131,28 @@ class Job extends BaseModel {
...
+    /**
+     * Remove all builds associated with this job and the job itself
+     * @return {Promise}        Resolves to null if remove successfully
+     */
+    remove() {
+        const removeBuilds = (() =>
+            // Iterate through the builds and remove them
+            this.getBuilds().then(builds => {
+                if (builds.length > 0) {
+                    builds.forEach(build => Promise.all(build.remove()));
```
stjohnjohnson: I think you want a map here instead.  Something like:

```
this.getBuilds().then(builds => Promise.all(builds.map(build => build.remove()))).then(() => super.remove());
```

stjohnjohnson (changes requested):

lib/job.js
d2lam: Wouldn't that just call getBuilds() once? I wanted to account for the scenario when there are like 200 builds, but getBuilds would just get the first 50 builds and remove them.

stjohnjohnson: Oh... I'm not sure how promises will work with the way you wrote this function.  I can tell you the `Promise.all` code concerns me.

d2lam: What concerns you in particular? This would only resolve when all the builds are removed. Do you have any suggestions? 

@petey  Any better way to write this?

petey: You are both right, in a fashion. `Promise.all` takes an array of promises and waits for them all to be resolved.

```
this.getBuilds().then(builds => {
    if (builds.length === 0) {
        return null; // or super.remove();
    }
    return Promise.all(builds.map(build => build.remove())))
        .then(() => removeBuilds());
```

Same sort of code in pipeline for jobs.

petey (changes requested):

lib/job.js
d2lam: @stjohnjohnson  @petey  Done!

stjohnjohnson (approved):
LGTM

package.json
```diff
@@ -30,6 +30,7 @@
   ],
   "devDependencies": {
     "chai": "^3.5.0",
+    "clone": "^1.0.2",
```
petey: Why do we pull in clone, instead of using `hoek.clone()`?

test/lib/pipeline.test.js
```diff
@@ -1,5 +1,6 @@
 'use strict';
 const assert = require('chai').assert;
+const clone = require('clone');
```
petey: `hoek.clone()`?

test/lib/pipeline.test.js
```diff
@@ -562,4 +564,66 @@ describe('Pipeline Model', () => {
...
+
+        afterEach(() => {
+            jobFactoryMock.list.reset();
+            publishJob.remove.reset();
+            mainJob.remove.reset();
+            blahJob.remove.reset();
+        });
+
+        it('remove jobs recursively', () => {
+            const nonArchived = clone(archived);
```
petey: `hoek.clone`

test/lib/pipeline.test.js
```diff
@@ -562,4 +564,66 @@ describe('Pipeline Model', () => {
...
+                assert.callCount(publishJob.remove, 4);
+                assert.callCount(mainJob.remove, 4);
+                assert.callCount(blahJob.remove, 2);
+
+                // Delete the pipeline
+                assert.calledOnce(datastore.remove);
+            });
+        });
+
+        it('fail if getJobs returns error', () => {
```
petey, line 617: I feel like there should be a test for a `job.remove()` rejects.

test/lib/job.test.js
```diff
@@ -236,4 +263,40 @@ describe('Job Model', () => {
...
+
+            return job.remove().then(() => {
+                assert.callCount(buildFactoryMock.list, 5);
+                assert.callCount(build1.remove, 4);     // remove builds recursively
+                assert.callCount(build2.remove, 4);
+                assert.calledOnce(datastore.remove);    // remove the job
+            });
+        });
+
+        it('fail if getBuilds returns error', () => {
```
petey, line 291: maybe a test for `build.remove()` rejects.

petey (changes requested):

d2lam:
@petey I added more tests.

petey (approved):

petey (approved):
👍

PR #110, opened 2016-09-24
[WIP] Add scmRepo to pipeline

catto:
I added a `scmRepo` to a modelConfig for a pipeline. screwdriver-cd/screwdriver#160
I'm not sure this is correct but I think I shouldn't call getRepoId at [here](https://github.com/screwdriver-cd/screwdriver/pull/204/files#diff-1eb6375f360812e82460d42f926123b1L77).

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -1,6 +1,7 @@
 /* eslint no-param-reassign: ["error", { "props": false }] */
 'use strict';
 const BaseModel = require('./base');
+const PipelineFactory = require('./pipelineFactory');
 const parser = require('screwdriver-config-parser');
 const nodeify = require('./nodeify');
 const hoek = require('hoek');
@@ -281,23 +282,15 @@ class PipelineModel extends BaseModel {
      * @return {Promise}
      */
     refreshScmRepo() {
+        const factory = PipelineFactory.getInstance();
+
         // Skip if no changes
         if (!this.isDirty('scmUrl') && this.scmRepo) {
             return Promise.resolve();
         }
 
-        // get the admin of the pipeline
         return this.admin
-            // and their token
-            .then(user => user.unsealToken())
-            // get the expanded repo information
-            .then(token =>
-                this.scm.getRepoId({
-                    scmUrl: this.scmUrl,
-                    token
-                })
-            )
-            // and store it
+            .then(user => factory.getScmRepo(user, this.scmUrl))
             .then(scmRepo => {
                 this.scmRepo = scmRepo;
             });
--- a/lib/pipelineFactory.js
+++ b/lib/pipelineFactory.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const BaseFactory = require('./baseFactory');
+const UserFactory = require('./userFactory');
 const Pipeline = require('./pipeline');
 const hoek = require('hoek');
 let instance;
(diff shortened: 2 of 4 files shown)


stjohnjohnson:
Why not use the new function added here: https://github.com/screwdriver-cd/models/pull/106/files

catto:
Ahh.. I completely misunderstood.
I'll change these to use `refreshScmRepo` and add duplicate checking instead of it.

catto:
Hmm.. we can call `refreshScmRepo` if a pipeline already exists. But `id` for new pipeline will be generated using `scmRepo.id` before a creation of new pipeline, so we can't call `refreshScmRepo`. We need to get scmRepo and call `super.create` with scmRepo, don't we?

stjohnjohnson:
@catto where can we move that function then?  Maybe make it static somehow?

catto:
@stjohnjohnson 🙆 I'll fix them.
I was just wondering if you had design discussions and how much code I can change.

catto:
I made it static and added duplication checks. If they are okay, I will change pipeline plugin to catch the error and return a conflict error.

lib/pipeline.js
```diff
@@ -123,6 +115,35 @@ class PipelineModel extends BaseModel {
     }
 
     /**
+     * Get the first admin username of the pipeline.
+     * @method getFirstAdminName
+     * @return {String}
+     */
+    // TODO: Get the first admin. Validate if it's valid. If not, remove from admins field,
+    //       and retry with the next admin.
+    getFirstAdminName() {
```
petey: `getFirstAdminName`, and `getAdmin` should appear in source order before they are used (in `get admin`)

lib/pipeline.js
```diff
@@ -287,20 +308,31 @@ class PipelineModel extends BaseModel {
...
+    }
+
+    /**
+     * Get the SCM Repo based on the SCM URL
+     * @method getScmRepo
+     * @param  {String}  username The admin username for the repository of the scmUrl
+     * @param  {String}  scmUrl   The scmUrl for the SCM Repo object
+     * @return {Promise}
+     */
+    static getScmRepo(username, scm, scmUrl) {
```
petey: `getScmRepo` should appear in source order before it is used

lib/pipelineFactory.js
```diff
@@ -42,8 +42,22 @@ class PipelineFactory extends BaseFactory {
         const modelConfig = hoek.applyToDefaults({
             createTime: (new Date()).toISOString()
         }, config);
+        const admin = config.admins[0];
 
-        return super.create(modelConfig);
+        return Pipeline.getScmRepo(admin, this.scmPlugin, config.scmUrl)
```
petey: I don't really understand why this is something the pipeline model has to be a part of, other than it is a convenient place to have static methods to get a user model. 

On the other hand, the pipelineFactory could have a method that fetches a user model, and scmPlugin could have a method `getRepoData(userModel, scmUrl)` or whatever that does the lookup.

petey (changes requested):

lib/pipelineFactory.js
catto: That's reasonable. I have moved `getScmRepo` to the pipelineFactory. I think `getRepoData` in the scmPlugin is not necessary in this case.

petey:
Closing in favor of work in #120

PR #111, opened 2016-09-28
Removing pipeline also removes its secrets

d2lam:
Just realized we should probably remove the pipeline's secrets too. 

Relates to https://github.com/screwdriver-cd/models/issues/105

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -322,7 +322,9 @@ class PipelineModel extends BaseModel {
                     .then(() => removeJobs(archived));
             }));
 
-        return removeJobs(true)                 // remove archived jobs
+        return this.secrets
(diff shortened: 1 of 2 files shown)


stjohnjohnson (approved):
Good catch

lib/pipeline.js
```diff
@@ -322,7 +322,9 @@ class PipelineModel extends BaseModel {
                     .then(() => removeJobs(archived));
             }));
 
-        return removeJobs(true)                 // remove archived jobs
+        return this.secrets
+            .then(secrets => Promise.all(secrets.map(secret => secret.remove())))   // remove secrets
```
petey, line 326: This only works as long as there are 50 or fewer secrets.

petey (approved):
:+1: Should be ok for most pipelines.

PR #112, opened 2016-09-30
PR sync reads screwdriver yaml from its branch instead of master branch

d2lam:
Currently, PR build reads screwdriver.yaml from master branch. I think we can add this, then in webhooks, pass in the scmUrl: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/webhooks/github.js#L208

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -26,9 +26,10 @@ class PipelineModel extends BaseModel {
      * Create, update, or disable jobs if necessary.
      * Store/update the pipeline workflow
      * @method sync
+     * @param  [String]   scmUrl                Optional scmUrl to sync
      * @return {Promise}
      */
-    sync() {
+    sync(scmUrl) {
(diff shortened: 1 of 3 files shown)


stjohnjohnson:
@d2lam if you do this, then will the sync update all the jobs (even main job) to be based on that PR?

d2lam:
@stjohnjohnson  That's true. What can we do so that it won't override but still works as expected? I can't think of a way unless we have a separate pipeline for it.

stjohnjohnson (changes requested):
Will damage things. 👎

d2lam:
@stjohnjohnson  We should discuss how to fix this. One way is when there is a change of screwdriver.yaml for a PR, then we create a separate pipeline for it and load from there instead. But that means the PR job will be in 2 pipelines (the master branch and its own branch). When the PR is closed then we delete that pipeline. But there are edge cases where we don't want to delete it. We need to go over this carefully but it seems complicated :(

PR #113, opened 2016-10-03
Remove getBuildsForJobId

d2lam:
- Remove getBuildsForJobId
- Add a function to get all running builds. This is to simplify the actions required in the API for PR sync. We won't need to call getBuilds recursively when pagination is added.

Relates to https://github.com/screwdriver-cd/screwdriver/issues/241

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -133,6 +133,28 @@ class Job extends BaseModel {
     }
 
     /**
+     * Return all running builds that belong to this job
+     * @return {Promise}        List of running builds
+     */
+    getRunningBuilds() {
+        const listConfig = {
+            params: {
+                jobId: this.id,
+                status: ['RUNNING', 'QUEUED']
+            }
+        };
+
+        // Lazy load factory dependency to prevent circular dependency issues
+        // https://nodejs.org/api/modules.html#modules_cycles
+        /* eslint-disable global-require */
+        const BuildFactory = require('./buildFactory');
+        /* eslint-enable global-require */
+        const factory = BuildFactory.getInstance();
+
+        return factory.list(listConfig);
+    }
+
+    /**
      * Remove all builds associated with this job and the job itself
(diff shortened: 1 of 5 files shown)


stjohnjohnson (approved):
Awesome!!

lib/job.js
```diff
@@ -133,6 +133,28 @@ class Job extends BaseModel {
...
 
     /**
+     * Return all running builds that belong to this job
+     * @return {Promise}        List of running builds
+     */
+    getRunningBuilds() {
+        const listConfig = {
+            params: {
+                jobId: this.id,
+                status: 'RUNNING'
```
petey: Aren't `QUEUED` builds also sort of like `RUNNING` builds?

petey (approved):
👍

lib/job.js
d2lam: That is true! I guess I can put RUNNING || QUEUED here but I would need to modify the datastore dynamo to support something like this.

stjohnjohnson: +1

lib/job.js
```diff
@@ -133,6 +133,28 @@ class Job extends BaseModel {
...
+        };
+
+        // Lazy load factory dependency to prevent circular dependency issues
+        // https://nodejs.org/api/modules.html#modules_cycles
+        /* eslint-disable global-require */
+        const BuildFactory = require('./buildFactory');
+        /* eslint-enable global-require */
+        const factory = BuildFactory.getInstance();
+
+        return factory.list(listConfig);
```
tkyi, line 154: Will the new function return sorted builds?

d2lam: It's descending sort by default.

lib/job.js
d2lam: @petey Addressed this.

tkyi (approved):
👍

PR #116, opened 2016-10-07
Set code on unfinished steps when build is updated to done

petey:
Builds that are aborted by the user or launcher do not result in the step codes being updated. This causes odd behavior in the API and UI related to logging like: https://github.com/screwdriver-cd/screwdriver/issues/248

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -6,6 +6,7 @@ const executor = Symbol('executor');
 const apiUri = Symbol('apiUri');
 const tokenGen = Symbol('tokenGen');
 const uiUri = Symbol('uiUri');
+const ABORT_CODE = 130; // 128 + SIGINT 2 (^C)
 
 /**
  * Update status to SCM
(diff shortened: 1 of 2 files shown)


lib/build.js
```diff
@@ -209,6 +210,34 @@ class BuildModel extends BaseModel {
...
+     */
+    abortSteps() {
+        const now = (new Date()).toISOString();
+        let steps = this.steps;
+
+        // fail first step if never started
+        if (!steps[0].startTime) {
+            steps[0].startTime = now;
+            steps[0].endTime = now;
+            steps[0].code = ABORT_CODE;
```
d2lam: Don't we want to set status for ALL the steps, not just the first step?

d2lam (changes requested):

lib/build.js
petey: Yeah. I made a bad assumption on how the log api worked.

stjohnjohnson: Wait, don't you want it just for the current running step?

stjohnjohnson (changes requested):
Shouldn't we only fail steps that are currently in progress?

lib/build.js
petey: I made that assumption at first, but we've clearly seen cases where it looks like the build failed when no steps ever started.
The log api doesn't care what start/end times are. It only looks for an exit code to determine if the step is finished, so every step needs to have an exit code in order to prevent the log api from returning x-more-data=true. While the UI is nice and doesn't allow you to open logs views for steps that don't appear to have started, nothing prevents someone from asking the api directly.
I'm happy to back some of this out, but https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/builds/steps/logs.js would need to be smarter.

stjohnjohnson: I think we should fix both, have this fail any active steps and then update the log api to check that the step has even started.

petey: updated.

stjohnjohnson (approved):
Ship it!

FenrirUnbound (approved):
Looks good to me

handsomefastkitty-size_restricted

d2lam (approved):

PR #117, opened 2016-10-07
List config expects pagination to be defined

d2lam:
We don't have pagination, but baseFactory's list method still expects pagination to be passed in: https://github.com/screwdriver-cd/models/blob/master/lib/baseFactory.js#L125-L133

This is causing:

```
161007/234620.188, [request,server,error] data: TypeError: Cannot read property 'count' of undefined
    at BuildFactory.list (/usr/src/app/node_modules/screwdriver-models/lib/baseFactory.js:130:39)
    at Job.getRunningBuilds (/usr/src/app/node_modules/screwdriver-models/lib/job.js:154:24)
    at stopJob (/usr/src/app/node_modules/screwdriver-api/plugins/webhooks/github.js:28:16)
    at jobFactory.get.then.job (/usr/src/app/node_modules/screwdriver-api/plugins/webhooks/github.js:115:13)
    at process._tickDomainCallback (internal/process/next_tick.js:129:7)
```

### DIFF
--- a/lib/baseFactory.js
+++ b/lib/baseFactory.js
@@ -2,6 +2,8 @@
 const schema = require('screwdriver-data-schema');
 const hashr = require('screwdriver-hashr');
 const hoek = require('hoek');
+const PAGINATE_PAGE = 1;
+const PAGINATE_COUNT = 50;
 
 class BaseFactory {
     /**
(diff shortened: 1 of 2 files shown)


tkyi:
Should we change the expectations in the baseFactory or in models?

minzcmu (approved):
👍

PR #119, opened 2016-10-10
Remove hoek.applyToDefaults on complex objects

petey:
`hoek.applyToDefaults` does numerous clones of input objects over the course of merging into a single output. This has negative effects when dealing with the sequalize module as an SCM ORM. This PR removes the use of that method for all complex objects.

### DIFF
--- a/lib/baseFactory.js
+++ b/lib/baseFactory.js
@@ -70,11 +70,15 @@ class BaseFactory {
         });
 
         return this.datastore.save(modelConfig)
-            .then(modelData => this.createClass(hoek.applyToDefaults(config, {
-                datastore: this.datastore,
-                scm: this.scm,
-                id: modelData.id
-            })));
+            .then(modelData => {
+                const c = config;
+
+                c.datastore = this.datastore;
+                c.scm = this.scm;
+                c.id = modelData.id;
+
+                return this.createClass(c);
+            });
     }
 
     /**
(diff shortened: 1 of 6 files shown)


stjohnjohnson (approved):
LGTM.  I hope we don't run into more issues.

lib/baseFactory.js
```diff
@@ -70,11 +70,15 @@ class BaseFactory {
...
         return this.datastore.save(modelConfig)
-            .then(modelData => this.createClass(hoek.applyToDefaults(config, {
-                datastore: this.datastore,
-                scm: this.scm,
-                id: modelData.id
-            })));
+            .then(modelData => {
+                const c = config;
+
+                c.datastore = this.datastore;
```
d2lam, line 76: For this to do the exact same thing as before, shouldn't this block be:

```
c.datastore = this.datastore || config.datastore;
c.scm = this.scm || config.scm;
c.id = modelData.id || config.id;
```

d2lam (changes requested):

lib/baseFactory.js
petey: Since `this.datastore`, `this.scm`, and `modelData.id` are always defined, I don't really see the difference.

d2lam: True. I'm fine with it.

d2lam (approved):

PR #120, opened 2016-10-12
Refactor models to use new scm-base functions

tkyi:
- Use `scmUri` over `scmUrl`
- Call `parseUrl` to get `scmUri`
- Call `decorateCommit`
- Make changes to conform to new eslint rules
- Bump data-schema version to 14.0.0
- Bump models version to 16.0.0

Based on changes made in https://github.com/screwdriver-cd/scm-base/pull/11
Related to https://github.com/screwdriver-cd/screwdriver/issues/160

### DIFF
--- a/lib/buildFactory.js
+++ b/lib/buildFactory.js
@@ -6,7 +6,7 @@ let instance;
 
 /**
  * Gathers a sha for the build
- * @method getCommitSha
+ * @method getCommitInfo
  * @param  {Object}         config
  * @param  {JobModel}       config.job                  Instance of Job Model
  * @param  {Scm}            config.scm                  Instance of SCM
@@ -15,12 +15,7 @@ let instance;
  * @param  {String}         config.modelConfig.username The name of the user
  * @return {Promise}
  */
-function getCommitSha(config) {
-    // Short circuit if sha already defined
-    if (config.modelConfig.sha) {
-        return Promise.resolve(config.modelConfig.sha);
-    }
-
+function getCommitInfo(config) {
     // Lazy load factory dependency to prevent circular dependency issues
     // https://nodejs.org/api/modules.html#modules_cycles
     /* eslint-disable global-require */
@@ -44,9 +39,21 @@ function getCommitSha(config) {
 
         return user.unsealToken()
             .then(token => config.scm.getCommitSha({
-                scmUrl: pipeline.scmUrl,
+                scmUri: pipeline.scmUri,
                 token
-            }));
+            })
+            .then(sha =>
+                config.scm.decorateCommit({
+                    scmUri: pipeline.scmUri,
+                    sha,
+                    token
+                })
+                .then(decoratedCommit => Promise.resolve({
+                    sha,
+                    decoratedCommit
+                }))
+            )
+        );
     });
 }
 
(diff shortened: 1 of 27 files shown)


FenrirUnbound:
Thing to note is that `scm-github` hasn't adopted the new `scm-base@2.x` interface

tkyi:
@FenrirUnbound Should be okay if I bumped major versions for `models`, right?

lib/buildFactory.js
```diff
@@ -117,20 +122,21 @@ class BuildFactory extends BaseFactory {
...
-                    .then(sha => {
+                return getCommitInfo({ job, scm: this.scm, modelConfig })
+                    .then((data) => {
                         // TODO: support matrix jobs
                         const index = number.toString().split('.')[1] || 0;
                         const permutation = job.permutations[index];
 
-                        modelConfig.sha = sha;
+                        modelConfig.commit = data;
+                        console.log('commit info: ', modelConfig.commit);
```
petey: console.log

petey (changes requested):

lib/buildFactory.js
```diff
@@ -44,9 +39,19 @@ function getCommitSha(config) {
 
         return user.unsealToken()
             .then(token => config.scm.getCommitSha({
-                scmUrl: pipeline.scmUrl,
+                scmUri: pipeline.scmUri,
                 token
-            }));
+            })
+            .then((sha) => {
+                config.modelConfig.sha = sha;
```
d2lam: What is this for? It's not being used.

lib/pipeline.js
```diff
@@ -276,44 +278,50 @@ class PipelineModel extends BaseModel {
...
-        // get the admin of the pipeline
-        return this.admin
-            // and their token
-            .then(user => user.unsealToken())
+        return Promise.all([
+            this.admin,                          // get the admin of the pipeline
+            this.scm.parseUrl(this.checkoutUrl)  // get the scmUri
+        ]).then(([user, scmUri]) => {
+            this.scmUri = scmUri;
+            this.checkoutUrl = null; // TODO: delete checkoutUrl?
```
d2lam: Why?

d2lam (changes requested):

lib/buildFactory.js
tkyi: Do we not want to return the `sha` in the `build` blob?

lib/pipeline.js
tkyi: I don't think we want to save `checkoutUrl` permanently in the `pipeline` blob

lib/buildFactory.js
d2lam: What do you mean? You're just setting the config.modelConfig.sha here. You're not returning or doing anything with it in this function.

FenrirUnbound: It looks like we're modifying a parameter, which is a code smell. The function already passes back the SHA, as well as additional data regarding the SHA. 

Why do we want the function to also mutate a given parameter?

tkyi: The `decorateCommit` function does not include a SHA in its return.
Changed the `getCommitInfo` function to return an object with the decorated commit and the SHA.

d2lam (approved):
LGTM 👍

petey (approved):

PR #121, opened 2016-10-13
Update config-parser

petey:
The new config parser is promised based, and should always resolve with a valid config


tkyi (approved):

tkyi:
There's a bunch of eslint errors but I fixed all those in my PR already
https://github.com/screwdriver-cd/models/pull/120
So we could merge that one first

PR #122, opened 2016-10-14
Add scmRepo when create pipeline and remove refreshScmRepo

minzcmu:
- Add `scmRepo` blob when creating pipeline
- Remove `refreshScmRepo`  in `sync` 
- Because we do not have`scmUri` update endpoint now, no need to refresh the `scmRepo`

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -23,6 +23,28 @@ class PipelineModel extends BaseModel {
     }
 
     /**
+     * Get the screwdriver configuration for the pipeline at the given ref
+     * @method getConfiguration
+     * @param  {String}  [ref=scmUri]   Reference to the repo
+     * @return {Promise}                Resolves to parsed and flattened configuration
+     */
+    getConfiguration(ref) {
+        const uri = ref || this.scmUri;
+
+        return this.admin.then(user => user.unsealToken())
+            // fetch the screwdriver.yaml file
+            .then(token =>
+                this.scm.getFile({
+                    scmUri: uri,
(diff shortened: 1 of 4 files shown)


tkyi (approved):

lib/pipeline.js
```diff
@@ -38,10 +38,8 @@ class PipelineModel extends BaseModel {
 
         const factory = JobFactory.getInstance();
 
-        // update the generated scm repo
-        return this.refreshScmRepo()
-            // get the pipeline configuration
-            .then(() => this.getConfiguration())
+        // get the pipeline configuration
+        return this.getConfiguration()
```
petey, line 64: getConfiguration should appear in source order before it is used.

lib/pipelineFactory.js
```diff
@@ -34,21 +34,29 @@ class PipelineFactory extends BaseFactory {
...
-            modelConfig.createTime = (new Date()).toISOString();
+        const userFactory = UserFactory.getInstance();
+        const modelConfig = config;
 
-            return super.create(modelConfig);
-        });
+        modelConfig.createTime = (new Date()).toISOString();
+
+        return userFactory.get({ username: Object.keys(config.admins)[0] })
+                .then(user => user.unsealToken())
```
petey: Why is this double indented?

petey (changes requested):

lib/pipeline.js
minzcmu: fixed :)

petey (approved):
going_right_round

PR #124, opened 2016-10-21
Determine scmUri key for scm.getFile() dynamically

minzcmu:
- Determine scmUri key for scm.getFile() dynamically since we have 2 sets of schema for the method now
- Related https://github.com/screwdriver-cd/data-schema/pull/78


d2lam (approved):
Good catch! 👍

PR #125, opened 2016-10-21
[1] Take in ref instead whole scmUrl

minzcmu:


lib/pipeline.js
```diff
@@ -29,18 +29,21 @@ class PipelineModel extends BaseModel {
      * @return {Promise}                Resolves to parsed and flattened configuration
```
tkyi: Can we make the jsdoc more clear?

tkyi (approved):
👍 LGTM just need to fix jsDoc

PR #132, opened 2016-11-11
[5] Add checkout-code step to build

minzcmu:
- Add checkout-code step to build

Blocked by https://github.com/screwdriver-cd/screwdriver/pull/356
Related to https://github.com/screwdriver-cd/screwdriver/issues/312

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # CHANGELOG
 
+## 18.0.0
+
+Breaking changes:
+  * Scm plugins need to have `getCheckoutCommand` implemented.
+
+Features:
+  * Add `sd-checkout-code` step to `steps` of build object.
+
 ## 17.1.0
 
 Features:
--- a/README.md
+++ b/README.md
@@ -283,6 +283,8 @@ factory.create(config).then(model => {
 | config.username | String | Yes | User who made the change to kick off the build |
 | config.container | String | No | Container for the build to run in |
 | config.sha | String | No | SHA used to kick off the build |
+| config.prRef | String | No | PR branch or reference; required for PR jobs |
+| config.eventId | String | No | Id of the event this build belongs to |
 
 #### Get
 Get a build based on id. Can pass the generatedId for the build, or the unique keys for the model, and the id will be determined automatically.
--- a/lib/buildFactory.js
+++ b/lib/buildFactory.js
@@ -112,6 +112,7 @@ class BuildFactory extends BaseFactory {
      * @param  {String}    config.jobId          The job associated with this build
      * @param  {String}    config.username       The user that created this build
      * @param  {String}    [config.sha]          The sha of the build
+     * @param  {String}    [config.prRef]        The PR branch or reference
      * @return {Promise}
      */
     create(config) {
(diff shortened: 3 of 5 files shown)


README.md
```diff
@@ -283,6 +283,8 @@ factory.create(config).then(model => {
 | config.username | String | Yes | User who made the change to kick off the build |
 | config.container | String | No | Container for the build to run in |
 | config.sha | String | No | SHA used to kick off the build |
+| config.prRef | String | No | PR branch or reference; required for PR jobs |
+| config.eventId | String | No | Id of the event this build belongs to |
```
tkyi, line 287: 👍  nice!

petey:
Why are we renaming sd-setup?

test/lib/buildFactory.test.js
```diff
@@ -156,6 +161,7 @@ describe('Build Factory', () => {
         beforeEach(() => {
             scmMock.getCommitSha.resolves(sha);
             scmMock.decorateCommit.resolves(commit);
+            scmMock.getCheckoutCommand.resolves({ name: 'checkout-code', command: 'git clone' });
```
tkyi: could you use `resolves(steps[0])`?

tkyi (approved):
Nice!

minzcmu:
@petey  In the launcher, we are doing `checkout`, `set up workspace`, and `set up environments` in `sd-setup` step. We just move the `checkout` part out here, so we think `checkout-code` is more accurate. Can we still keep `sd-setup` and log something about  `set up workspace`, and `set up environments` in that step? Is it better? Then `sd-setup` will be the first step, and `checkout-code` will be the second one.

lib/buildFactory.js
```diff
@@ -136,29 +137,48 @@ class BuildFactory extends BaseFactory {
...
+                    }
+
+                    modelConfig.container = permutation.image;
+                    modelConfig.environment = permutation.environment;
+                    modelConfig.steps = permutation.commands.map(command => ({
+                        name: command.name,
+                        command: command.command
+                    }));
+
+                    const scmParts = pipeline.scmUri.split(':');
```
d2lam: You can do this
`const [host,,branch] = pipeline.scmUri.split(':');`

lib/buildFactory.js
```diff
@@ -136,29 +137,48 @@ class BuildFactory extends BaseFactory {
...
+                    const orgRepo = pipeline.scmRepo.name.split('/');
```
d2lam: Same thing here:
`const [org,repo] = pipeline.scmUri.split('/');`

So there aren't too many magic numbers below.

lib/buildFactory.js
```diff
@@ -136,29 +137,48 @@ class BuildFactory extends BaseFactory {
...
+                        repo: orgRepo[1],
+                        sha: modelConfig.sha
+                    };
+
+                    if (modelConfig.prRef) {
+                        checkoutConfig.prRef = modelConfig.prRef;
+                    }
+
+                    return this.scm.getCheckoutCommand(checkoutConfig).then((command) => {
+                        modelConfig.steps.unshift(command);
```
d2lam, line 176: Don't we still need the `sd-setup` for other things?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "screwdriver-models",
-  "version": "17.1.0",
+  "version": "17.2.0",
```
d2lam: I feel like this should be a major bump.

d2lam (changes requested):

package.json
petey: I agree that this is a major version bump.

test/lib/buildFactory.test.js
```diff
@@ -116,12 +117,16 @@ describe('Build Factory', () => {
...
+            name: 'screwdriver-cd/models'
+        };
+        const prRef = 'pull/3/merge';
         const username = 'i_made_the_request';
         const dateNow = Date.now();
         const isoTime = (new Date(dateNow)).toISOString();
         const container = 'node:4';
         const steps = [
-            { name: 'sd-setup' },
+            { command: 'git clone', name: 'checkout-code' },
```
petey: If this step is named something other than `sd-setup`, then it needs to start with `sd-` to keep from stepping on user configs. 

If the launcher is still doing other special commands in `sd-setup` maybe those should also be moved out of launcher and inserted by the models. Maybe that should be done in another PR. We want to keep the `sd-` steps to a minimum.

lib/buildFactory.js
petey: I wasn't aware of this feature of destructuring. Thanks!

petey (changes requested):

lib/buildFactory.js
minzcmu: Add it back now.

test/lib/buildFactory.test.js
minzcmu: Makes sense to add `sd-` as the prefix! 👍 Keep the `sd-setup` for now and will see if we can move those commands out in another PR. Rename this to  `sd-checkout-code`. @tkyi We need to change scm-base, scm-bitbucket and scm-github with the step name  `sd-checkout-code`.

minzcmu: @tkyi Or can we just add a prefix `sd-` to command.name here?

d2lam: We can set them to `sd-checkout-code` to be consistent across. So we can tell user that those steps are reserved. If we just add prefix, it's harder to document which steps are reserved and should not be inside the user config.

tkyi: @minz1027 okay I'll start opening pull requests for those 👍

lib/buildFactory.js
```diff
@@ -136,29 +137,49 @@ class BuildFactory extends BaseFactory {
...
-                        modelConfig.steps = permutation.commands.map(command => ({
-                            name: command.name,
-                            command: command.command
-                        }));
+                return Promise.all([
+                    job.pipeline,
+                    getCommitInfo({ job, scm: this.scm, modelConfig })
+                ]).then(([pipeline, data]) => {
+                    // TODO: support matrix jobs
+                    const index = number.toString().split('.')[1] || 0;
```
FenrirUnbound, line 145: Does `Date.now()` return a number with a decimal?

tkyi: No. We're not supporting matrix jobs yet, so it'll always use index 0 right now and only use the first permutation object.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## 18.0.0
```
FenrirUnbound, line 3: It's a major version bump, so this would indicate that there is a backwards-incompatible change. Could we point out what's the breaking change in this release?

FenrirUnbound (changes requested):

CHANGELOG.md
tkyi: Added a `breaking changes` section

FenrirUnbound (approved):
👍 

insane teeterboard skills - imgur

d2lam (approved):
Looks good 👍

PR #134, opened 2016-11-22
fix: PR event fails to create when author not found

petey:
Should fix the error that greenkeeper opens a PR in github and the webhook fails to create an event.


stjohnjohnson:
Eh, we should just be using the pipeline user, not the person who created the PR.

petey:
Updated to use pipeline admin's token.

tkyi (approved):

stjohnjohnson (approved):
So much 👍

PR #135, opened 2016-11-23
fix: launcher needs an sd-setup command

petey:
* Launcher is hardcoded to use an sd-setup command for certain operations that are logged.


FenrirUnbound (approved):
👍 

4107q04wh9zx

d2lam (approved):

PR #141, opened 2016-12-09
feat(371): update repo webhooks during sync operation

FenrirUnbound:
Looking for feedback on how to refactor this section. 

We discussed about how to break down this beast of a method. So this is what it boils down to:

1. There are now 2 unique operations that occur: `syncPipeline` and `addWebhook`. We want to separate those into their own methods, and shrink the `sync` function to just [the Promise.all](https://github.com/screwdriver-cd/models/compare/repoWebhooks?expand=1#diff-679e90b5b15ab1b73ddb7ed2377e7413R131). Does this structure make sense, knowing how they look like now?

1. The current implementation does not find `addWebhook` to be a critical operation. If it fails to add a webhook, we keep going as if it didn't. Is this behavior reasonable?

Related to screwdriver-cd/screwdriver#379

EDIT1: Link to working issue

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -66,7 +66,7 @@ class PipelineModel extends BaseModel {
         const factory = JobFactory.getInstance();
 
         // get the pipeline configuration
-        return this.getConfiguration()
+        const syncPipeline = this.getConfiguration()
             // get list of jobs to create
             .then((parsedConfig) => {
                 this.workflow = parsedConfig.workflow;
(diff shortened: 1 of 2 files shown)


petey (approved):
I like this approach, but I'd like to see the syncPipeline, and addWebhook stuff refactored into separate functions later.

minzcmu:
In the `/webhook` routes, we call `pipeline.sync()` for PR and Push events. Add webhooks here will be redundant since it will not reach this route if the webhook is not set up.
 https://github.com/screwdriver-cd/screwdriver/blob/ae8fce5fade10fa788cddb3d04fe918d79e8e2c0/plugins/webhooks/index.js
Agree @petey with a separate function for addWebhook and should call it only when we need it.

FenrirUnbound:
>@minz1027
>In the /webhook routes, we call pipeline.sync() for PR and Push events. Add webhooks here will be redundant since it will not reach this route if the webhook is not set up.

The `pipeline.sync` operation also happens during pipeline creation: https://github.com/screwdriver-cd/screwdriver/blob/a65ce51af6f340396c35732ac122f9c684d5f742/plugins/pipelines/create.js#L92

The initial design behind doing it in the sync was to ensure that the repository is configured correctly. Some scenarios where we may want to update the repository would be when the secret is rotated or we need to update the list of events we care about.

I think we all agree that there's some level of redundancy here. Petey suggested to refactor them into separate functions later on. Do you believe we should refactor meow?

d2lam:
We were thinking have another function called `addWebhooks` and call it right before this line: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/pipelines/create.js#L92

I forgot why we decided to put in sync? There will be a lot of overhead since `sync()` is called at a lot of places and I feel like this is should only need to be called once, right before the pipeline is created.

stjohnjohnson (approved):
🐑 🇮🇹 
mhmm

Looks good, let's refactor it out to create and a separate API route later.

PR #144, opened 2016-12-20
feat: use numeric ids instead of a hash

petey:
BREAKING CHANGE: removes hash based unique keys in favor of numeric ids. No longer generates hashes based on unique properties.

Related Epic: screwdriver-cd/screwdriver#399
Blocked By: screwdriver-cd/data-schema#98

### DIFF
--- a/lib/base.js
+++ b/lib/base.js
@@ -78,14 +78,11 @@ class BaseModel {
         this[dirty].forEach((key) => {
             data[key] = this[rowData][key];
         });
-        delete data.id;
+        data.id = this.id;
 
         const datastoreConfig = {
             table: this[table],
-            params: {
-                id: this.id,
-                data
-            }
+            params: data
         };
 
         // TODO: sync `this` with db response?
--- a/lib/baseFactory.js
+++ b/lib/baseFactory.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const schema = require('screwdriver-data-schema');
-const hashr = require('screwdriver-hashr');
 const hoek = require('hoek');
 const PAGINATE_PAGE = 1;
 const PAGINATE_COUNT = 50;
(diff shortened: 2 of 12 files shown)


lib/baseFactory.js
```diff
@@ -89,19 +68,17 @@ class BaseFactory {
...
-        let id = config;
+        let params = {};
 
-        if (typeof config === 'object' && config.id) {
-            id = config.id;
+        if (typeof config === 'number' || typeof config.id === 'number') {
+            params.id = config.id || config;
         } else if (typeof config === 'object') {
-            id = this.generateId(config);
+            params = config;
```
petey: I'm not sure if we need to filter for valid keys here like we do in create.

lib/pipeline.js
```diff
@@ -120,11 +120,11 @@ class PipelineModel extends BaseModel {
                 scmUri: this.scmUri,
                 token,
                 webhookUrl: 'https://api.screwdriver.cd/v4/webhooks'
-            }).catch((err) => {
+            }).catch(() => {
                 // Disregard errors since they're non-critical
 
-                console.error('An error was encountered & discarded while updating webhooks.');
-                console.error(err);
+                // console.error('An error was encountered & discarded while updating webhooks.');
```
stjohnjohnson: Should we just remove these?

lib/baseFactory.js
stjohnjohnson: I would filter on unique keys

stjohnjohnson (approved):
oh boy

lib/base.js
```diff
@@ -78,14 +78,11 @@ class BaseModel {
...
-        delete data.id;
+        data.id = this.id;
 
         const datastoreConfig = {
             table: this[table],
-            params: {
-                id: this.id,
-                data
-            }
+            params: data
```
d2lam, line 85: What's the expected behavior if nothing needs to be updated? Right now it will fail validation since there is only 1 key in `params`: https://github.com/screwdriver-cd/data-schema/blob/a5e3894079a6284cc441e72f9f6918e991a0e423/plugins/datastore.js#L16

stjohnjohnson: See line 74, it shortcuts out.

d2lam (approved):
Cool looks good 👍

PR #145, opened 2016-12-20
fix(399): convert string ids to numbers

petey:
The api pulls params as strings, and does not try to convert them. This PR has the model attempt to convert ids to a number before using them in get requests.


stjohnjohnson (approved):
👍 
woah

PR #148, opened 2017-01-07
feat: Add syncPRs() for pipeline

d2lam:
There are 3 cases:
- `PR-1` is closed, but it's still in the current list of jobs, so archive it. Related: https://github.com/screwdriver-cd/screwdriver/issues/389
- `PR-1` is opened, but it's not in the current list of jobs, so create it. 
- `PR-1` is opened, it's in the current list of jobs, but it was archived, so update it. 

Mocking `this.scm.getOpenedPRs`. This should return the list of opened PR jobs, for example: `['PR-12', PR-20']`

This is a major version bump. The `this.scm` passed in needs to be the newest version which has `getOpenedPRs` implemented. Otherwise this will break. 

Related Epic https://github.com/screwdriver-cd/screwdriver/issues/417

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -50,6 +50,99 @@ class PipelineModel extends BaseModel {
     }
 
     /**
+     * Get a list of PR jobs to create or update
+     * @method _checkPRState
+     * @param  {Array}    existingJobs      List pipeline's existing jobs
+     * @param  {Array}    openedPRs         List of opened PRs coming from SCM
+     * @return {Promise}                    Resolves to the list of jobs to archive, unarchive and create
+     * Note: toArchive and toUnarchive is an array of job objects; toCreate is an array of objects with name and ref.
+     */
+    _checkPRState(existingJobs, openedPRs) {
+        const jobList = {
+            toCreate: [],
+            toArchive: [],
+            toUnarchive: []
+        };
+        const existingPRs = existingJobs.filter(j => j.isPR());   // list of PRs according to SD
+        const existingPRsNames = existingPRs.map(j => j.name);
+        const openedPRsNames = openedPRs.map(j => j.name);
+        const openedPRsRef = openedPRs.map(j => j.ref);
+
+        existingPRs.forEach((job) => {
+            // if PR is closed, add it to archive list
+            if (openedPRsNames.indexOf(job.name) < 0 && job.archived === false) {
+                jobList.toArchive.push(job);
+            }
+        });
+
+        openedPRsNames.forEach((name, i) => {
+            const index = existingPRsNames.indexOf(name);
+
+            if (index < 0) {
+                // if opened PR is not in the list of existingPRs, create it
+                jobList.toCreate.push({ name, ref: openedPRsRef[i] });
+            } else {
+                const job = existingPRs[index];
+
+                // if opened PR was previously archived, unarchive it
+                if (job.archived) {
+                    jobList.toUnarchive.push(existingPRs[index]);
+                }
+            }
+        });
+
+        return jobList;
+    }
+
+    /**
+     * Go through the job list and archive/unarchive it
+     * @method _updateJobArchive
+     * @param  {Array}        jobList     List of job objects
+     * @param  {boolean}      archived    Archived value to update to
+     * @return {Promise}
+     */
+    _updateJobArchive(jobList, archived) {
+        const jobsToUpdate = [];
+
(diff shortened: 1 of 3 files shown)


FenrirUnbound:
Looking over the logic, it seems like the flow goes something like:

1. Sort the PRs
2. Do stuff with the PR
  2.a. Archive
  2.b Create
  2.c Update

One approach you could take is to map-reduce. Have a function that accomplishes *Task 1* and splits the PRs into 3 lists. Each of the operations in *Task 2* could be a separate function that, given a set of jobs, performs that operation.

d2lam:
@FenrirUnbound  I'm not sure what you meant.
It's currently 2 parts:
- Get a list of actual opened PRs (by contacting SCM). For example: `['PR-2']` --- list A
- Get a list of PRs that Screwdriver thinks are opened. For example: `['PR-1', 'PR-3']` --- list B

Then there are 3 scenarios like I mentioned:
- `PR-2` is already closed. If it's in list B, then archive it. If not, then it's fine. 
- `PR-2` is opened. If it's not in list B, then create it. If it's in list B and archived, then unarchive it. If it's in list B and not archived, then it's fine.

FenrirUnbound:
Right. The initial promise chain starts by fetching all of the PRs: which PRs are actually open and which PRs we previously thought were open. That's a prerequisite to the remaining scenarios.

My suggested Task 1 would -- given a PR's state -- determine if it needs to be archived, created, or updated. It wouldn't actually do those actions; it would just sort the PRs into different lists.

Task 2 would be where all those PRs would be processed. This would be the area where the archiving happens, or creations occurs, or when the information is updated.

FenrirUnbound:
Pseudo-code:

```
Promise.all([
  this.jobs,
  this.scm.getOpenedPRs({
    scmUri: this.scmUri,
    token
  }),
  this.getConfiguration()
]).then(([existingJobs, openPrs, config]) => {
  const [toArchive, toCreate, toUpdate] = checkPrState(existingJobs, openPrs, config)

  return Promise.all([
    archivePr(toArchive),
    createPr(toCreate),
    updatePr(toUpdate)
  ]);
});

```

d2lam:
@FenrirUnbound  Got it! 👍

d2lam:
@FenrirUnbound  I refactored the code in a separate commit. Will squash later. Please help re-review! :)

lib/pipeline.js
```diff
@@ -64,6 +156,33 @@ class PipelineModel extends BaseModel {
...
+        return this.token.then(token =>
+            Promise.all([
+                this.jobs,
+                this.scm.getOpenedPRs({
+                    scmUri: this.scmUri,
+                    token
+                }),
+                this.getConfiguration()
+            ]).then(([existingJobs, openedPRs, config]) => {
+                const jobList = checkPrState(existingJobs, openedPRs, config);
```
FenrirUnbound: You're not using `config` in the `checkPrState` function

FenrirUnbound (changes requested):

lib/pipeline.js
```diff
@@ -8,6 +8,98 @@ const hoek = require('hoek');
...
+}
+
+/**
+ * Go through the job list and archive/unarchive it
+ * @method updateJob
+ * @param  {Array}        jobList     List of job objects
+ * @param  {boolean}      archived    Archived value to update to
+ * @return {Promise}
+ */
+function updateJob(jobList, archived) {
```
FenrirUnbound: Minor: `updateJob` doesn't update the whole job. Maybe "updateJobArchive"?

lib/pipeline.js
d2lam: Good catch! 👍

d2lam:
@FenrirUnbound  Thanks for your suggestions! :) Please re-review! :D

FenrirUnbound (approved):
👍 

Looks good!

giphy 39

FenrirUnbound:
Are you planning on waiting for the SCM modules to catch up and implement the extra endpoint?

d2lam:
@FenrirUnbound  Just rewrote some parts due to the extra requirements from scm-base. Please help re-review :) 

And yeah want to merge in this order: scm-base -> scm-github/scm-bitbucket -> models -> API

FenrirUnbound (approved):
👍 

Still looks good to me! 😸

minzcmu (approved):
👍 LGTM

PR #151, opened 2017-03-14
feat(470): add getTemplate function in templateFactory

minzcmu:
- add getTemplate function in templateFactory
- will return latest version of template based on name, version, label provided

### DIFF
--- a/lib/templateFactory.js
+++ b/lib/templateFactory.js
@@ -59,7 +59,7 @@ class TemplateFactory extends BaseFactory {
      * @param  {String}     config.name          The template name
      * @param  {String}     config.version       Version of the template
      * @param  {String}     [config.label]       Label of template
-     * @return {Promise}
+     * @return {Promise}                         Resolves template model or undefined if not found
      */
     getTemplate(config) {
         // get all templates with the same name
@@ -73,7 +73,7 @@ class TemplateFactory extends BaseFactory {
                 filtered.sort((a, b) => compareVersions(b.version, a.version));
 
                 // If no config.label, return latest one
-                if (filtered && filtered.length > 0 && !config.label) {
+                if (filtered.length > 0 && !config.label) {
                     return filtered[0];
                 }
 
--- a/lib/templateFactory.js
+++ b/lib/templateFactory.js
@@ -2,6 +2,7 @@
 
 const BaseFactory = require('./baseFactory');
 const Template = require('./template');
+const compareVersions = require('compare-versions');
 let instance;
 
 class TemplateFactory extends BaseFactory {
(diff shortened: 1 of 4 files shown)


test/lib/templateFactory.test.js
```diff
@@ -122,4 +128,63 @@ describe('Template Factory', () => {
...
+                    name: 'testTemplateName',
+                    version: '1.0.2',
+                    labels: ['testLabel', 'otherLabel']
+                },
+                {
+                    id: 'd398fb192747c9a0124e9e5b4e6e8e841cf8c71c',
+                    name: 'testTemplateName',
+                    version: '1.0.3',
+                    labels: ['otherLabel']
+                }
```
d2lam: Can we add an entry that is version `2`, to see if it filters it out correctly?
Also, let's make this in random order, since the `rangeKey` doesn't work and might not return them in this ascending order.

d2lam (changes requested):

test/lib/templateFactory.test.js
minzcmu: addressed

lib/templateFactory.js
```diff
@@ -52,6 +53,36 @@ class TemplateFactory extends BaseFactory {
...
+            .then((templates) => {
+                // get templates that have version prefix as config.version
+                const filtered = templates.filter(template =>
+                    template.version.startsWith(config.version));
+
+                // Sort templates by descending order
+                filtered.sort((a, b) => compareVersions(b.version, a.version));
+
+                // If no config.label, return latest one
+                if (filtered.length > 0 && !config.label) {
```
FenrirUnbound, line 76: What if `config.label` is empty?

test/lib/templateFactory.test.js
```diff
@@ -122,4 +123,69 @@ describe('Template Factory', () => {
...
+
+            return factory.getTemplate(config).then((model) => {
+                assert.instanceOf(model, Template);
+                Object.keys(expected).forEach((key) => {
+                    assert.strictEqual(model[key], expected[key]);
+                });
+            });
+        });
+
+        it('should get the correct template for a given config without label', () => {
```
FenrirUnbound, line 179: There are no error cases described

FenrirUnbound (changes requested):

lib/templateFactory.js
FenrirUnbound: For some odd reason, I misinterpreted config.label to be an array. As a String, my concern isn't valid.

This is cool 👍

test/lib/templateFactory.test.js
```diff
@@ -122,4 +123,69 @@ describe('Template Factory', () => {
...
+                {
+                    id: '151c9b11e4a9a27e9e374daca6e59df37d8cf00f',
+                    name: 'testTemplateName',
+                    version: '2.0.1',
+                    labels: ['testLabel', 'otherLabel']
+                }
+            ];
+        });
+
+        it('should get the correct template for a given config with label', () => {
```
FenrirUnbound, line 166: What happens in the case where I give you a label that doesn't match?

FenrirUnbound: What happens in the edge case where `super.list` returns nothing?

minzcmu: It will return undefined. According to the doc for find: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

test/lib/templateFactory.test.js
minzcmu: Added a test case when list resolves an empty array.

d2lam (approved):

FenrirUnbound (approved):
👍 

source 5

PR #153, opened 2017-03-20
fix(470): Pass templateFactory to config-parser

d2lam:
For template to work, we need to pass templateFactory to the `config-parser`: https://github.com/screwdriver-cd/config-parser/blob/master/index.js#L26

This change is backward-compatible.


minzcmu (approved):
👍 LGTM

PR #155, opened 2017-04-17
feat(473): Add alwaysRun flag to steps in builds

jerryzhang222:
Allows users to define a step verbosely in their `screwdriver.yaml`, to make it possible for an `alwaysRun` flag to exist:

```yaml
jobs:
    main:
        steps:
            - name: hello
               command: echo hello
               alwaysRun: true
```

Related to: https://github.com/screwdriver-cd/screwdriver/issues/473

### DIFF
--- a/lib/buildFactory.js
+++ b/lib/buildFactory.js
@@ -157,7 +157,8 @@ class BuildFactory extends BaseFactory {
                     modelConfig.environment = permutation.environment;
                     modelConfig.steps = permutation.commands.map(command => ({
                         name: command.name,
-                        command: command.command
+                        command: command.command,
+                        alwaysRun: command.alwaysRun || false
                     }));
 
                     const bookendConfig = { pipeline, job, build: modelConfig };
@@ -168,7 +169,7 @@ class BuildFactory extends BaseFactory {
                     ]).then(([setup, teardown]) => {
                         modelConfig.steps = setup.concat(modelConfig.steps, teardown);
                         // Launcher is hardcoded to do some business in sd-setup-launcher
-                        modelConfig.steps.unshift({ name: 'sd-setup-launcher' });
+                        modelConfig.steps.unshift({ name: 'sd-setup-launcher', alwaysRun: true });
 
                         return super.create(modelConfig);
                     });
--- a/test/lib/buildFactory.test.js
+++ b/test/lib/buildFactory.test.js
@@ -134,30 +134,30 @@ describe('Build Factory', () => {
         const isoTime = (new Date(dateNow)).toISOString();
         const container = 'node:4';
         const steps = [
-            { name: 'sd-setup-launcher' },
-            { name: 'sd-setup-scm', command: 'git clone' },
-            { command: 'npm install', name: 'init' },
-            { command: 'npm test', name: 'test' }
+            { name: 'sd-setup-launcher', alwaysRun: true },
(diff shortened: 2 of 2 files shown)


lib/buildFactory.js
```diff
@@ -168,7 +169,7 @@ class BuildFactory extends BaseFactory {
                     ]).then(([setup, teardown]) => {
                         modelConfig.steps = setup.concat(modelConfig.steps, teardown);
                         // Launcher is hardcoded to do some business in sd-setup-launcher
-                        modelConfig.steps.unshift({ name: 'sd-setup-launcher' });
+                        modelConfig.steps.unshift({ name: 'sd-setup-launcher', alwaysRun: true });
```
minzcmu, line 172: This is the first step, so I guess it will alwaysRun without the flag.. And I think we are missing the `alwaysRun: true` flag for the teardown bookends?

minzcmu (changes requested):

test/lib/buildFactory.test.js
```diff
@@ -134,30 +134,30 @@ describe('Build Factory', () => {
         const isoTime = (new Date(dateNow)).toISOString();
         const container = 'node:4';
         const steps = [
-            { name: 'sd-setup-launcher' },
-            { name: 'sd-setup-scm', command: 'git clone' },
-            { command: 'npm install', name: 'init' },
-            { command: 'npm test', name: 'test' }
+            { name: 'sd-setup-launcher', alwaysRun: true },
+            { name: 'sd-setup-scm', command: 'git clone', alwaysRun: true },
+            { command: 'npm install', name: 'init', alwaysRun: false },
```
tkyi, line 139: Would probably be nicer if `alwaysRun` defaults to `false` so we don't need to define it everywhere

test/lib/templateFactory.test.js
```diff
@@ -101,6 +101,11 @@ describe('Template Factory', () => {
                 scmUri
             }).then((model) => {
                 assert.instanceOf(model, Template);
+                console.log('HELLOOOO');
```
tkyi: Extra `console.log`s

lib/buildFactory.js
jerryzhang222: I'm planning to add `alwaysRun: true` to all bookends here: https://github.com/screwdriver-cd/build-bookend/pull/8

test/lib/templateFactory.test.js
jerryzhang222: Sorry just debugging something right now - this isn't quite ready for review!

test/lib/buildFactory.test.js
jerryzhang222: It does actually default to `false` in `buildFactory.js` itself (https://github.com/screwdriver-cd/models/blob/alwaysrun/lib/buildFactory.js#L161), but these are just the expected steps (test data).

lib/buildFactory.js
jerryzhang222: And my thought on the first step alwaysRun flag is that I just want things to be standardized. So I want this step to have an `alwaysRun` key and I figure that the value makes more sense `true` than `false`. Not sure if that makes sense.

d2lam: I think it's ok to have it here. In case the launcher logic changes, this is still true.

tkyi (approved):
LGTM

d2lam (approved):

minzcmu (approved):

PR #157, opened 2017-04-18
fix: Fix tests + comments after changing scmUri to pipelineId in templates

jerryzhang222:
Unit tests were failing because `scmUri` was no longer a key. It was replaced by `pipelineId` as a way of identifying which pipeline a template was associated to. 

This PR fixes tests and comments.

### DIFF
--- a/lib/templateFactory.js
+++ b/lib/templateFactory.js
@@ -32,8 +32,8 @@ class TemplateFactory extends BaseFactory {
      * @param  {String}     config.description   Description of the template
      * @param  {String}     config.maintainer    Maintainer's email
      * @param  {Object}     config.config        Config of the screwdriver-template.yaml
-     * @param  {String}     config.scmUri        ScmUri of the template pipeline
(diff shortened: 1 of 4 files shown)


d2lam (approved):

lib/templateFactory.js
```diff
@@ -32,8 +32,8 @@ class TemplateFactory extends BaseFactory {
      * @param  {String}     config.description   Description of the template
      * @param  {String}     config.maintainer    Maintainer's email
      * @param  {Object}     config.config        Config of the screwdriver-template.yaml
-     * @param  {String}     config.scmUri        ScmUri of the template pipeline
-     * @param  {Array}      [config.labels]        Labels attached to the template
+     * @param  {String}     config.pipelineId    pipelineId of the template pipeline
```
d2lam: of the template?

minzcmu: The pipelineId should be Integer actually, need to fix the tests too...

PR #160, opened 2017-05-12
feat: Add token models and tests

ian-fox:
* add Token and TokenFactory models
* add tests for those models

Notably with the current plan for authenticating, we don't actually need to be able to get a single token, just getting a list of the tokens associated with a user. It'd be pretty easy to implement though.

### DIFF
--- a/lib/token.js
+++ b/lib/token.js
@@ -1,0 +1,44 @@
+'use strict';
+
+const BaseModel = require('./base');
+const nodeify = require('./nodeify');
+const iron = require('iron');
+
+// Symbols for private members
+const password = Symbol('password');
+
+class TokenModel extends BaseModel {
+    /**
+     * Construct a TokenModel object
+     * @method constructor
+     * @param  {Object}    config
+     * @param  {Object}    config.datastore         Object that will perform operations on the datastore
+     * @param  {Number}    config.userId            The ID of the associated user
+     * @param  {String}    config.value             The token value
+     * @param  {String}    config.tokenId           The token ID
+     * @param  {String}    config.name              The token name
+     * @param  {String}    config.description       The token description
+     * @param  {Date}      config.lastUsed          The last time the token was accessed
+     * @param  {String}    config.password          The encryption password
+     */
+    constructor(config) {
+        super('token', config);
+        this[password] = config.password;
+    }
+
+    /**
+     * Update a token with sealed token value
+     * @method update
+     * @return {Promise}
+     */
+    update() {
+        return nodeify.withContext(iron, 'seal', [this.value, this[password], iron.defaults])
+            .then((sealed) => {
+                this.value = sealed;
+
+                return super.update();
+            });
+    }
+}
+
+module.exports = TokenModel;
--- a/lib/tokenFactory.js
+++ b/lib/tokenFactory.js
@@ -1,78 +1,63 @@
 'use strict';
 
 const BaseFactory = require('./baseFactory');
-const Secret = require('./secret');
+const Token = require('./token');
 const Lib = require('./library');
-// Get symbols for private fields
+// Symbols for private fields
 const password = Symbol('password');
 
 let instance;
 
-class SecretFactory extends BaseFactory {
+class TokenFactory extends BaseFactory {
     /**
-     * Construct a SecretFactory object
+     * Construct a TokenFactory object
      * @method constructor
      * @param  {Object}    config
      * @param  {Object}    config.datastore     Object that will perform operations on the datastore
      * @param  {String}    config.password      Password for encryption operations
      */
     constructor(config) {
-        super('secret', config);
+        super('token', config);
         this[password] = config.password;
     }
 
     /**
-     * Instantiate a Secret class
+     * Instantiate a Token class
      * @method createClass
      * @param  config
-     * @return {Secret}
+     * @return {Token}
      */
     createClass(config) {
         const c = config;
 
         c.password = this[password];
 
-        return new Secret(c);
+        return new Token(c);
     }
 
     /**
-     * Create a secret model
-     * Need to seal the secret before saving
+     * Create a token model
+     * Need to seal the token before saving
      * @method create
      * @param  {Object}     config
-     * @param  {String}     config.pipelineId      Pipeline Id
-     * @param  {String}     config.name            Secret name
-     * @param  {String}     config.value           Secret value
-     * @param  {Boolean}    config.allowInPR       Whether this secret can be shown in PR builds
+     * @param  {String}     config.userId            The ID of the associated user
+     * @param  {String}     config.value             The token value
+     * @param  {String}     config.name              The token name
+     * @param  {String}     config.description       The token description
      * @return {Promise}
      */
     create(config) {
         return Lib.sealValue(config.value, this[password])
             .then((sealed) => {
                 config.value = sealed;
+                config.lastUsed = null;
 
                 return super.create(config);
             });
     }
 
     /**
-     * Get a secret based on id
-     * @method get
-     * @param  {Mixed}   config    The configuration from which an id is generated or the actual id
-     * @return {Promise}
-     */
-    get(config) {
-        return super.get(config).then((secret) => {
-            if (!secret) {
-                return secret;
-            }
-
-            return Lib.unsealValue(secret, this[password]);
-        });
-    }
-
-    /**
-     * List secrets with pagination and filter options
+     * List tokens with pagination and filter options
      * @method list
      * @param  {Object}   config                  Config object
      * @param  {Object}   config.params           Parameters to filter on
@@ -82,23 +67,23 @@ class SecretFactory extends BaseFactory {
      * @return {Promise}
      */
     list(config) {
-        return super.list(config).then(secrets =>
-            Promise.all(secrets.map(secret => Lib.unsealValue(secret, this[password]))));
+        return super.list(config).then(tokens =>
+            Promise.all(tokens.map(token => Lib.unsealValue(token, this[password]))));
     }
 
     /**
-     * Get an instance of the SecretFactory
+     * Get an instance of the TokenFactory
      * @method getInstance
      * @param  {Object}     config
      * @param  {Datastore}  config.datastore    A datastore instance
      * @param  {String}     config.password     Password for encryption operations
-     * @return {SecretFactory}
+     * @return {TokenFactory}
      */
     static getInstance(config) {
-        instance = BaseFactory.getInstance(SecretFactory, instance, config);
+        instance = BaseFactory.getInstance(TokenFactory, instance, config);
 
         return instance;
     }
 }
 
-module.exports = SecretFactory;
+module.exports = TokenFactory;
--- a/test/lib/token.test.js
+++ b/test/lib/token.test.js
@@ -1,0 +1,101 @@
+'use strict';
+
+const assert = require('chai').assert;
+const sinon = require('sinon');
+const mockery = require('mockery');
+const schema = require('screwdriver-data-schema');
+
+sinon.assert.expose(assert, { prefix: '' });
+
+describe('Token Model', () => {
+    const password = 'password';
+    let BaseModel;
+    let TokenModel;
+    let ironMock;
+    let datastore;
+    let createConfig;
+    let token;
+
+    before(() => {
+        mockery.enable({
+            useCleanCache: true,
+            warnOnUnregistered: false
+        });
+    });
+
+    beforeEach(() => {
+        datastore = {
+            update: sinon.stub()
+        };
+        datastore.update.resolves({});
+
+        ironMock = {
+            seal: sinon.stub(),
+            unseal: sinon.stub(),
+            defaults: {}
+        };
+        mockery.registerMock('iron', ironMock);
+
+        /* eslint-disable global-require */
+        BaseModel = require('../../lib/base');
+        TokenModel = require('../../lib/token');
+        /* eslint-enable global-require */
+
+        createConfig = {
+            datastore,
+            userId: 12345,
+            value: 'A_SECRET_TOKEN_VALUE',
+            id: 6789,
+            name: 'Mobile client auth token',
+            description: 'For the mobile app',
+            lastUsed: '2017-05-10T01:49:59.327Z',
+            password
+        };
+        token = new TokenModel(createConfig);
+    });
+
+    afterEach(() => {
+        datastore = null;
+        mockery.deregisterAll();
+        mockery.resetCache();
+    });
+
+    after(() => {
+        mockery.disable();
+    });
+
+    it('is constructed properly', () => {
+        assert.instanceOf(token, TokenModel);
+        assert.instanceOf(token, BaseModel);
+        schema.models.token.allKeys.forEach((key) => {
+            assert.strictEqual(token[key], createConfig[key]);
+        });
+    });
+
+    describe('update', () => {
+        beforeEach(() => {
+            ironMock.seal.yieldsAsync(null, 'sealed');
+        });
+
+        it('promises to update a token and seal the value before datastore saves it', () => {
+            const newTimestamp = '2017-05-13T02:01:17.588Z';
+
+            token.value = 'unsealed';
+            token.lastUsed = newTimestamp;
+
+            return token.update()
+                .then(() => {
+                    assert.calledWith(ironMock.seal, 'unsealed', password, ironMock.defaults);
+
+                    assert.isTrue(datastore.update.calledWith({
+                        table: 'tokens',
+                        params: {
+                            id: 6789,
+                            value: 'sealed',
+                            lastUsed: newTimestamp
+                        }
+                    }));
+                });
+        });
+    });
+});
--- a/test/lib/tokenFactory.test.js
+++ b/test/lib/tokenFactory.test.js
@@ -1,0 +1,189 @@
+'use strict';
+
+const assert = require('chai').assert;
+const mockery = require('mockery');
+const sinon = require('sinon');
+
+sinon.assert.expose(assert, { prefix: '' });
+
+describe('Token Factory', () => {
+    const password = 'super_secure_password';
+    const name = 'mobile_token';
+    const description = 'a token for a mobile app';
+    const sealed = 'abcd';
+    const unsealed = 'efgh';
+    const userId = 6789;
+    const tokenId = 12345;
+    const tokenData = {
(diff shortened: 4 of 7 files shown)


lib/token.js
```diff
@@ -0,0 +1,45 @@
...
+const BaseModel = require('./base');
+const nodeify = require('./nodeify');
+const iron = require('iron');
+
+// Symbols for private members
+const password = Symbol('password');
+
+class TokenModel extends BaseModel {
+    /**
+     * Construct a BuildModel object
```
tkyi: **Token**Model object

tkyi (approved):

lib/token.js
```diff
@@ -0,0 +1,45 @@
...
+        super('token', config);
+        this[password] = config.password;
+    }
+
+    /**
+     * Update a token with sealed token value
+     * @method update
+     * @return {Promise}
+     */
+    update() {
```
r3rastogi, line 34: Why are Token and TokenFactory divided/broken up the way they are? For example, why isn't update() defined in the same module as sealToken()?

d2lam: `update()` is performed on the specific token.
TokenFactory is to perform things like LIST all the tokens, etc.

ian-fox: My understanding is that any time we go to store something like a token value, we want it to be encrypted. That happens either when creating or updating a model.

The factory is what handles creating new models (and getting existing ones), whereas once you have a model you call update() on the model itself, which means both of them need to be able to encrypt the token values. @d2lam can you confirm and/or elaborate please?

lib/tokenFactory.js
```diff
@@ -0,0 +1,116 @@
...
+     * List tokens with pagination and filter options
+     * @method list
+     * @param  {Object}   config                  Config object
+     * @param  {Object}   config.params           Parameters to filter on
+     * @param  {Object}   config.paginate         Pagination parameters
+     * @param  {Number}   config.paginate.count   Number of items per page
+     * @param  {Number}   config.paginate.page    Specific page of the set to return
+     * @return {Promise}
+     */
+    list(config) {
```
r3rastogi, line 69: From a quick in-person chat with @ian-fox , this is essentially an API that the API calls. @petey (or someone else who might know), can we assume that the tokens remain secure as long as we don't add a user-facing API route to all this?

ian-fox: Git says @d2lam wrote the list bit for secrets, so she would probably also be a good person to ask.

lib/tokenFactory.js
```diff
@@ -0,0 +1,116 @@
...
+
+    /**
+     * Get an instance of the TokenFactory
+     * @method getInstance
+     * @param  {Object}     config
+     * @param  {Datastore}  config.datastore    A datastore instance
+     * @param  {String}     config.password     Password for encryption operations
+     * @return {TokenFactory}
+     */
+    static getInstance(config) {
```
r3rastogi, line 82: What's the point of this? When is it used?

test/lib/token.test.js
```diff
@@ -0,0 +1,117 @@
...
+            defaults: {}
+        };
+        mockery.registerMock('iron', ironMock);
+
+        // eslint-disable-next-line no-global-assign
+        Date = function Date() {
+            this.toISOString = () => 'updated timestamp';
+        };
+
+        // eslint-disable-next-line global-require
```
r3rastogi: Can we just `eslint-disable global-require` and then re-enable it after both requires are declared? Instead of disabling each individual line

test/lib/token.test.js
```diff
@@ -0,0 +1,117 @@
...
+                    }));
+                });
+        });
+
+        it('updates the lastUsed time when a token is updated with no fields changed', () => {
+            assert.isFalse(token.isDirty());
+
+            // eslint-disable-next-line no-global-assign
+            Date = function Date() {
+                this.toISOString = () => 'a more different timestamp';
```
r3rastogi: "a different timestamp"? "a new timestamp"? I don't like the current wording here 😛

test/lib/tokenFactory.test.js
```diff
@@ -0,0 +1,187 @@
...
+        };
+        ironMock = {
+            seal: sinon.stub(),
+            unseal: sinon.stub(),
+            defaults: 'defaults'
+        };
+
+        mockery.registerMock('iron', ironMock);
+
+        // eslint-disable-next-line global-require
```
r3rastogi: Same comment here as token.test.js re: disabling eslint rules

lib/tokenFactory.js
d2lam: The list route would need to determine who calls the API. If it's a user, it shouldn't show the value. If it's a build, it should expose the value. Something like this: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/secrets/index.js#L42-L57
And https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/secrets/get.js#L39-L41

test/lib/tokenFactory.test.js
```diff
@@ -0,0 +1,187 @@
...
+            const f1 = TokenFactory.getInstance(config);
+            const f2 = TokenFactory.getInstance(config);
+
+            assert.instanceOf(f1, TokenFactory);
+            assert.instanceOf(f2, TokenFactory);
+
+            assert.equal(f1, f2);
+        });
+
+        it('should throw when config not supplied', () => {
```
r3rastogi: "should throw an error when ..."

lib/token.js
r3rastogi: @d2lam but isn't sealToken also performed on a specific token?

test/lib/token.test.js
ian-fox: Was a reference to an internet video, not great for readability though I guess :P

lib/tokenFactory.js
r3rastogi: @d2lam cool, that makes sense. Thanks!
We'd still never expose the unsealed values though right?

test/lib/token.test.js
ian-fox: We certainly could. Do you think that'd be cleaner? Both ways require two eslint config comments.

r3rastogi: Yeah, I think it would, especially if we need to add in more requires later on

test/lib/token.test.js
r3rastogi: Hahah gotcha! Feel free to leave it then :) Someone with better scifi/pop culture knowledge would enjoy it

lib/token.js
```diff
@@ -0,0 +1,45 @@
...
+        return nodeify.withContext(iron, 'seal', [this.value, this[password], iron.defaults])
```
petey, line 35: Once created, what would cause the token's `value` to change? Is it even necessary to unseal the value of the token at any time except when specifically requested by internal endpoints?

Shouldn't we only keep the sealed value of the token in `this.value`?

Also, TokenFactory.create gives you a token where `value` is already sealed. So, if you were to update some metadata of the token and called `update`, this would result in the value being sealed twice.

lib/tokenFactory.js
```diff
@@ -0,0 +1,116 @@
...
+let instance;
+
+/**
+ * Seal token value
+ * @method sealToken
+ * @param  {String}         tokenValue     Token value to seal
+ * @param  {String}         pw             Password to seal with
+ * @return {Promise}
+ */
+function sealToken(tokenValue, pw) {
```
petey: This code is identical to that in `secretFactory`. Seems like an opportunity for refactoring.

lib/tokenFactory.js
```diff
@@ -0,0 +1,116 @@
...
+}
+
+/**
+ * Unseal token value and return the token with the unsealed value
+ * @param  {TokenModel}         token   Token to unseal
+ * @param  {String}             pw      Password to unseal with[description]
+ * @return {Promise}
+ */
+function unsealToken(token, pw) {
+    return nodeify.withContext(iron, 'unseal', [token.value, pw, iron.defaults])
```
petey: This is also identical to `secretFactory`.

lib/token.js
```diff
@@ -0,0 +1,45 @@
...
+            .then((sealed) => {
+                this.value = sealed;
+                this.lastUsed = (new Date()).toISOString();
```
petey: Does updating token metadata really count as "used"?

lib/tokenFactory.js
ian-fox: ~~I had thought that was important from looking at other factories, but having gone back and looked at where the other factories getInstance methods are called I don't think it's actually necessary for tokens. It seems to be for when one object with a factory has another object with a factory as an attribute, and as far as I know there isn't going to be anything that uses a token like that.~~

It's called in screwdriver/bin/server.js when setting up the API

lib/tokenFactory.js
petey: We should never expose the token value to the user except after a create (or regenerate?).

test/lib/token.test.js
ian-fox: Specifically, [this](https://www.youtube.com/watch?v=90X5NJleYJQ) is the video in question

test/lib/token.test.js
```diff
@@ -0,0 +1,117 @@
...
+
+        ironMock = {
+            seal: sinon.stub(),
+            unseal: sinon.stub(),
+            defaults: {}
+        };
+        mockery.registerMock('iron', ironMock);
+
+        // eslint-disable-next-line no-global-assign
+        Date = function Date() {
```
petey: Never ever modify an Object you do not own. Use `sinon.match.string()` or some other matching technique.

If you REALLY need to validate operations on timestamps, you can use the `sinon.useFakeTimers` API, but note that this can have unexpected consequences and may break the code under test (and/or the tests themselves).

test/lib/token.test.js
petey: See above for comment about NEVER modifying objects you do not own.

test/lib/tokenFactory.test.js
```diff
@@ -0,0 +1,187 @@
...
+        // eslint-disable-next-line global-require
+        Token = require('../../lib/token');
+        // eslint-disable-next-line global-require
+        TokenFactory = require('../../lib/tokenFactory');
+
+        factory = new TokenFactory({ datastore, password });
+    });
+
+    afterEach(() => {
+        datastore = null;
```
petey, line 61: Why are you setting datastore to `null`?

petey (changes requested):

lib/token.js
ian-fox: Ooh that's a good point, I was thinking "this will make things easier, you can just call update when a token gets used" but didn't consider that we probably don't want to update that when changing name/description. I'll add another method like "touch" or something to specifically do that

lib/token.js
ian-fox: For changing: I was talking to Dao and Min yesterday, and they pointed out that github has an option to regenerate a token, which changes the value while leaving the name/description intact. If we wanted to add that, that would do it, but I might be getting ahead of myself there.

For keeping only the sealed version in value, I was following the pattern secrets use, which seems to be that when you retrieve it from the datastore you unseal it so that you can actually use the value, and then you seal it again when the model is put back into the datastore.

ian-fox: I may be missing something, is there a way to use the tokens without unsealing the value?

My plan was that when a request with a token came in, we could list all tokens for that user (with unsealed values) and compare to see if the token they provided matched any of them, and if so authenticate the request. If there's a way to do that without unsealing the values that certainly seems better though.

lib/tokenFactory.js
d2lam: Agree. We never expose unsealed values. The values stored in the database are sealed.

lib/tokenFactory.js
ian-fox: Sounds good. Would it be good to make a subclass of BaseFactory that deals with objects that have to be sealed, and then have SecretFactory and TokenFactory inherit from that?

test/lib/token.test.js
ian-fox: That makes a lot of sense, I hadn't considered the effect that could have on other code. 

In this case I just want to check that timestamps actually get updated when they're supposed to, so rather than asserting that the value that comes back is a specific date I'll try storing the initial value, and then comparing the one that comes back against that to make sure it's changed.

lib/tokenFactory.js
r3rastogi: Got it, thanks

lib/token.js
petey: We don't need to have a separate method to set a property of a Token object.

r3rastogi:
All my original questions have been answered, and everything I still had a question on has basically been asked by Petey! So as soon as all his comments are addressed, should be a :+1: from me 💃 [I will still review again though, don't worry about that hehe]

lib/tokenFactory.js
petey: It's a static method that turns TokenFactory into a singleton.

lib/token.js
petey: This is really tricky. I think that secrets may have an issue here, too, but we have never encountered it as we never call `update()` on a brand new Secret object. Secrets returned in `list` and `get` have values unsealed. Looking at how we are doing secrets, it isn't really the responsibility of the model to determine access control.

lib/token.js
ian-fox: Just pass the new date in the config when it needs to be updated then?

lib/tokenFactory.js
petey: Those seal and unseal functions aren't class methods. Make a separate library.

test/lib/token.test.js
petey: You wouldn't think it would be a big deal until you spend two days trying to figure out why `async.waterfall` starts causing tests to timeout because it is waiting on a `process.nextTick` that never comes.

lib/tokenFactory.js
ian-fox: Ahh right I was thinking of the Token and Secret models themselves, because their update methods are also very similar. The list methods are also identical, and the get methods (if we want a get method for tokens)

test/lib/token.test.js
ian-fox: Wow, that sounds painful. Definitely glad you pointed that out and I didn't have to learn that the hard way!

test/lib/tokenFactory.test.js
ian-fox: Checked with Dao, it resets the counter on calledOnce checks. Right now they pass without it because there's only one test for each one though

lib/token.js
ian-fox: ~~Would there ever be a case where we want to have a model, call update() on it and then continue performing operations with it?~~

~~If that's the case then update should either unseal the value again after saving the sealed value to the datastore, or maybe we should be using a separate field to do the saving so that we don't have to worry about if the value field is currently sealed or unsealed.~~

This does happen, I don't *think* it will happen with tokens but I'm not 100% on that.

test/lib/tokenFactory.test.js
FenrirUnbound: Even if you add a second test, calls counts shouldn't creep into other test cases. The stubs are [instantiated before every test case](https://github.com/screwdriver-cd/models/pull/160/files#diff-dda11b25e2c5b96ec3647835ce936ecaR39), so you get a new instance with no calls on every test run.

lib/token.js
FenrirUnbound: > it isn't really the responsibility of the model to determine access control

If the concern is exposing too much to the caller, an alternative would be to provide another method that compares a given value to the token. Something like:

```
const myPassedInValue = 'maybe_the_right_secret';

const match = token.compare(myPassedInValue);

console.log(match); // true if the value matches the token's value, false otherwise
```

The methods defined by the base model are the essential methods for synchronizing a model's data to the datastore. If there are other business rules that a particular model is responsible for, you can create additional methods to handle that case.

test/lib/tokenFactory.test.js
ian-fox: Just checked, right you are! I'll remove the line setting it to null.

lib/token.js
ian-fox: Consensus seems to be leave it for now, but remember that it exists. To this end I created an [issue](https://github.com/screwdriver-cd/screwdriver/issues/554) for it.

test/lib/tokenFactory.test.js
petey: FWIW, the setup and teardown of these tests is pretty expensive. There's not enough of them for it to really matter right now, but I have seen `mockery.deregisterAll();` with `mockery.resetCache();` cause big delays with large numbers of tests and mocked modules. You can actually get away with moving all your mock setup (except `factory = new TokenFactory`) into `before()`, and `beforeEach` is used to call:
```
datastore.save.reset();
...
factory = new TokenFactory({ datastore, password });
```

Then you'd move everything from `afterEach` into `after`.

lib/token.js
petey: I've been thinking that the way to solve this issue is to write custom getters and setters for the secret values. Then consumers of the model don't need to care about the seal/unseal operations (including things like get/update/save) because the actual value in the model is always the sealed one.

The drawback to this approach is that we may need to refactor places that already use these models to accept a promise as the return value instead of the raw value.

Something like:
```
get secret() {
  return nodeify.withContext(iron, 'unseal', [this.value, this[password], iron.defaults])
}
set secret() {
  // mark field as "dirty", then
  return nodeify.withContext(iron, 'seal', [this.value, this[password], iron.defaults])
}
```

FenrirUnbound: A method would set a better expectation of the result for the caller. The caller has to be aware of the return value, whereas a getter/setter makes it seem more like a flat property and would be more surprising if it was an asynchronous getter.

```
// method
model.getSecret().then(...);

// getter with promise
model.secret.then(..);
```

Why prefer the getter/setter in the case where you need to wait for a promise?

test/lib/tokenFactory.test.js
FenrirUnbound: > FWIW, the setup and teardown of these tests is pretty expensive

`useCleanCache` is a very expensive setup operation, but necessary.

As for teardown, it's a combination of `useCleanCache` and `.resetCache`. The `afterEach` is pretty redundant since `.deregisterAll` removes all the registered mocks from the cache, and then `.resetCache` just dumps it all and primes it again.

PR #161, opened 2017-05-17
feat: Add token models and tests

ian-fox:

Blocked by https://github.com/screwdriver-cd/data-schema/pull/132

Logic is much simpler now that we aren't storing the token value (and thus don't need to worry about encryption).

The JWT issued to the user will be able to be decoded to get a UUID. After we've used the public key to validate the token, we can then compare the received UUID to a table in the datastore to ensure the token hasn't been revoked. If it hasn't, the request is authenticated and the last-used time is updated.

### DIFF
--- a/lib/tokenFactory.js
+++ b/lib/tokenFactory.js
@@ -1,0 +1,59 @@
+'use strict';
+
+const BaseFactory = require('./baseFactory');
+const Token = require('./token');
+
+let instance;
+
+class TokenFactory extends BaseFactory {
+    /**
+     * Construct a TokenFactory object
+     * @method constructor
+     * @param  {Object}    config
+     * @param  {Object}    config.datastore     Object that will perform operations on the datastore
+     */
+    constructor(config) {
+        super('token', config);
+    }
+
+    /**
+     * Instantiate a Token class
+     * @method createClass
+     * @param  {Object}     config
+     * @return {Token}
+     */
+    createClass(config) {
+        return new Token(config);
+    }
+
+    /**
+     * Create a token model
+     * @method create
+     * @param  {Object}     config
+     * @param  {String}     config.userId            The ID of the associated user
+     * @param  {String}     config.value             The token value
+     * @param  {String}     config.name              The token name
+     * @param  {String}     config.description       The token description
+     * @return {Promise}
+     */
+    create(config) {
+        config.lastUsed = null;
+
+        return super.create(config);
+    }
+
+    /**
+     * Get an instance of the TokenFactory
+     * @method getInstance
+     * @param  {Object}     config
+     * @param  {Datastore}  config.datastore    A datastore instance
+     * @return {TokenFactory}
+     */
+    static getInstance(config) {
+        instance = BaseFactory.getInstance(TokenFactory, instance, config);
+
+        return instance;
+    }
+}
+
+module.exports = TokenFactory;
--- a/lib/user.js
+++ b/lib/user.js
@@ -3,6 +3,8 @@
 const BaseModel = require('./base');
 const iron = require('iron');
 const nodeify = require('./nodeify');
+const PAGINATE_COUNT = 50;
+const PAGINATE_PAGE = 1;
 // Get symbols for private fields
 const password = Symbol('password');
 
@@ -41,6 +43,60 @@ class UserModel extends BaseModel {
     unsealToken() {
         return nodeify.withContext(iron, 'unseal',
             [this.token, this[password], iron.defaults]);
+    }
+
+    /** Fetch a user's tokens
+    /* @property tokens
+    /* @return {Promise}
+    */
+    get tokens() {
+        const listConfig = {
+            params: {
+                userId: this.id
+            },
+            paginate: {
+                count: PAGINATE_COUNT,
+                page: PAGINATE_PAGE
+            }
+        };
+
+        // Lazy load factory dependency to prevent circular dependency issues
+        // https://nodejs.org/api/modules.html#modules_cycles
+        /* eslint-disable global-require */
+        const TokenFactory = require('./tokenFactory');
+        /* eslint-enable global-require */
+        const factory = TokenFactory.getInstance();
+        const tokens = factory.list(listConfig);
+
+        // ES6 has weird getters and setters in classes,
+        // so we redefine the pipeline property here to resolve to the
+        // resulting promise and not try to recreate the factory, etc.
+        Object.defineProperty(this, 'tokens', {
+            enumerable: true,
+            value: tokens
+        });
+
+        return tokens;
+    }
+
+    /**
+     * Test if a token is valid and belongs to the user
+     * @method validateToken
+     * @param  {String}  uuid       UUID of token to validate
+     * @return {Promise}
+     */
+    validateToken(uuid) {
+        return this.tokens.then((tokens) => {
+            const token = tokens.find(t => t.uuid === uuid);
+
+            if (!token) {
+                throw new Error('Token has been revoked.');
+            }
+
+            token.lastUsed = (new Date()).toISOString();
+
+            return token.update();
+        });
     }
 
     /**
--- a/test/lib/token.test.js
+++ b/test/lib/token.test.js
@@ -1,0 +1,75 @@
+'use strict';
+
+const assert = require('chai').assert;
+const sinon = require('sinon');
+const mockery = require('mockery');
+const schema = require('screwdriver-data-schema');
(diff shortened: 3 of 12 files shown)


lib/tokenFactory.js
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * Instantiate a Token class
+     * @method createClass
+     * @param  config
+     * @return {Token}
+     */
+    createClass(config) {
+        const c = config;
+
+        return new Token(c);
```
d2lam: can just pass in `config` here: `new Token(config)`

lib/user.js
```diff
@@ -43,6 +45,71 @@ class UserModel extends BaseModel {
...
     }
 
+    /** Fetch a user's tokens
+    /* @property tokens
+    /* @return {Promise}
+    */
+    get tokens() {
+        const listConfig = {
+            params: {
+                user: this.id
```
d2lam: `userId`

d2lam (changes requested):

lib/user.js
```diff
@@ -43,6 +45,71 @@ class UserModel extends BaseModel {
...
+
+            // Lazy load factory dependency to prevent circular dependency issues
+            // https://nodejs.org/api/modules.html#modules_cycles
+            /* eslint-disable global-require */
+            const TokenFactory = require('./tokenFactory');
+            /* eslint-enable global-require */
+
+            const factory = TokenFactory.getInstance();
+
+            return factory.get(match.id).then((token) => {
```
d2lam: I'm wondering why we need to call `factory.get(match.id)` again? Isn't `match` the token we need?
Can't we do:
```
match.lastUsed = ...
match.update()
```

test/lib/token.test.js
```diff
@@ -0,0 +1,82 @@
...
+        });
+        datastore = {
+            update: sinon.stub()
+        };
+    });
+
+    beforeEach(() => {
+        datastore.update.resolves({});
+
+        /* eslint-disable global-require */
```
FenrirUnbound: You can push these `require()` statements to the top of the file.

test/lib/token.test.js
```diff
@@ -0,0 +1,82 @@
...
+            lastUsed: '2017-05-10T01:49:59.327Z',
+            password
+        };
+        token = new TokenModel(createConfig);
+    });
+
+    after(() => {
+        mockery.deregisterAll();
+        mockery.resetCache();
+        mockery.disable();
```
FenrirUnbound: If you're doing a `.disable()`, you don't need `.deregisterAll()` or `.resetCache()`

test/lib/tokenFactory.test.js
```diff
@@ -0,0 +1,120 @@
...
+        });
+    });
+
+    beforeEach(() => {
+        datastore = {
+            save: sinon.stub()
+        };
+
+        /* eslint-disable global-require */
+        Token = require('../../lib/token');
```
FenrirUnbound, line 42: This `require()` can be pushed to the top of the file as well

lib/user.js
```diff
@@ -43,6 +45,71 @@ class UserModel extends BaseModel {
...
+     * @method validateToken
+     * @param  {String}  uuid       UUID of token to validate
+     * @return {Promise}
+     */
+    validateToken(uuid) {
+        return this.tokens.then((tokens) => {
+            const match = tokens.find(token => token.uuid === uuid);
+
+            if (!match) {
+                throw new Error('Token has been revoked.');
```
FenrirUnbound: In this case, what's the difference between a token that's been revoked and a token that was never made?

FenrirUnbound:
If this PR relies on screwdriver-cd/data-schema#132, we should bump the dependency version as well.

lib/tokenFactory.js
```diff
@@ -0,0 +1,61 @@
...
+     * Get an instance of the TokenFactory
+     * @method getInstance
+     * @param  {Object}     config
+     * @param  {Datastore}  config.datastore    A datastore instance
+     * @return {TokenFactory}
+     */
+    static getInstance(config) {
+        instance = BaseFactory.getInstance(TokenFactory, instance, config);
+
+        return instance;
```
r3rastogi, line 55: can't we just say `return BaseFactory.getInstance(TokenFactory, instance, config);` here?

lib/user.js
ian-fox: The match that comes from the list doesn't have the update method on it.

Just realized that factory.createClass is probably better though.

lib/user.js
```diff
@@ -43,6 +45,71 @@ class UserModel extends BaseModel {
...
+            }
+        };
+
+        // Lazy load factory dependency to prevent circular dependency issues
+        // https://nodejs.org/api/modules.html#modules_cycles
+        /* eslint-disable global-require */
+        const TokenFactory = require('./tokenFactory');
+        /* eslint-enable global-require */
+        const factory = TokenFactory.getInstance();
+
```
r3rastogi: remove extra newline

lib/user.js
ian-fox: In the API layer, to authenticate first we check that the token is valid with the public key. If it fails there, it doesn't have to bother checking to see if the token is still valid or not because it already knows it isn't. So it should only get to here if it would be a valid token except for the fact that it's been revoked.

lib/tokenFactory.js
ian-fox: We need to assign the value to instance, which was declared in a `let` above. Otherwise I think it will make a new factory each time.

lib/user.js
minzcmu: Hmm, a jwt token that is not generated by us shouldn't reach here. We will verify the signature of the token first, then decode to get the uuid and do the check here.

test/lib/tokenFactory.test.js
ian-fox: This one is breaking the test for some reason, the factory is creating an Object instead of a Token.

lib/user.js
ian-fox: Wait a second, list calls createclass. I must have been mocking it wrong.

ian-fox:
Data schema dependency updated.

test/lib/tokenFactory.test.js
ian-fox: I think it's because we need to mock the datastore before we import the TokenModel.

lib/tokenFactory.js
r3rastogi: Thanks, makes sense

r3rastogi (approved):
Looks good to me!

lib/tokenFactory.js
```diff
@@ -0,0 +1,59 @@
...
+     * @param  {Object}    config.datastore     Object that will perform operations on the datastore
+     */
+    constructor(config) {
+        super('token', config);
+    }
+
+    /**
+     * Instantiate a Token class
+     * @method createClass
+     * @param  config
```
d2lam: need to change the comment here to include the type of `config`

test/lib/tokenFactory.test.js
FenrirUnbound: I see it now. You're getting that because you'll also need to mock the `Token` class as well.

This is fine as-is.

test/lib/user.test.js
```diff
@@ -186,4 +191,76 @@ describe('User Model', () => {
...
+                paginate
+            };
+
+            tokenFactoryMock.list.resolves(null);
+            // when we fetch tokens it resolves to a promise
+            assert.isFunction(user.tokens.then);
+            // and a factory is called to create that promise
+            assert.calledWith(tokenFactoryMock.list, listConfig);
+
+            // When we call user.tokens again it is still a promise
```
FenrirUnbound, line 215: Why is it significant if it's still a promise the second time around?

FenrirUnbound (approved):
👍 

Looks good to me

losing to a rock paper scissors robot - imgur

test/lib/user.test.js
ian-fox: I'm checking with Petey, this was something I saw in the tests for the other places that we're using getters.

~~My guess is that we want to make sure the bit where we do `Object.defineProperty` doesn't override the getter, but I'm not sure why we'd need to test for that every time.~~

It's to check that it _is_ overwritten, that it doesn't go back to the database again once it gets it the first time.

lib/tokenFactory.js
ian-fox: Should I do that on the other factories too?

d2lam (approved):

PR #162, opened 2017-05-22
WIP: fix: Refactor token model and add sealField function

ian-fox:
First commit:
* Added a function to seal a field on a model, since that will be reused between tokens and secrets

Second commit:
* Tokens store their value, as using a key to verify them instead would mean we couldn't rotate the key
* Iron stuff added back to tokens so the value isn't stored unencrypted

### DIFF
--- a/test/lib/token.test.js
+++ b/test/lib/token.test.js
@@ -32,7 +32,7 @@ describe('Token Model', () => {
         createConfig = {
             datastore,
             userId: 12345,
-            uuid: '1a2b3c',
+            value: '1a2b3c',
             id: 6789,
             name: 'Mobile client auth token',
             description: 'For the mobile app',
--- a/test/lib/tokenFactory.test.js
+++ b/test/lib/tokenFactory.test.js
@@ -10,7 +10,7 @@ require('sinon-as-promised');
 describe('Token Factory', () => {
     const name = 'mobile_token';
     const description = 'a token for a mobile app';
-    const uuid = 'abc123';
+    const value = 'abc123';
     const userId = 6789;
     const tokenId = 12345;
     const tokenData = {
@@ -18,7 +18,7 @@ describe('Token Factory', () => {
         userId,
         description,
         name,
-        uuid,
+        value,
         lastUsed: null
     };
     let TokenFactory;
@@ -70,7 +70,7 @@ describe('Token Factory', () => {
                 userId,
                 name,
                 description,
-                uuid,
+                value,
                 lastUsed: null
             };
 
@@ -80,7 +80,7 @@ describe('Token Factory', () => {
                 userId,
                 name,
                 description,
-                uuid
+                value
             }).then((model) => {
                 assert.isTrue(datastore.save.calledOnce);
                 assert.calledWith(datastore.save, {
--- a/lib/token.js
+++ b/lib/token.js
@@ -9,7 +9,7 @@ class TokenModel extends BaseModel {
      * @param  {Object}    config
      * @param  {Object}    config.datastore         Object that will perform operations on the datastore
      * @param  {Number}    config.userId            The ID of the associated user
-     * @param  {String}    config.uuid              UUID for revoking the token
+     * @param  {String}    config.value             Hashed token value
      * @param  {String}    config.name              The token name
      * @param  {String}    config.description       The token description
      * @param  {String}    config.lastUsed          The last time the token was used (ISO String)
--- a/lib/tokenFactory.js
+++ b/lib/tokenFactory.js
@@ -31,7 +31,7 @@ class TokenFactory extends BaseFactory {
      * @method create
      * @param  {Object}     config
      * @param  {String}     config.userId            The ID of the associated user
-     * @param  {String}     config.value             The token value
+     * @param  {String}     config.value             The hashed token value
      * @param  {String}     config.name              The token name
      * @param  {String}     config.description       The token description
      * @return {Promise}
--- a/lib/user.js
+++ b/lib/user.js
@@ -80,26 +80,6 @@ class UserModel extends BaseModel {
     }
 
     /**
-     * Test if a token is valid and belongs to the user
-     * @method validateToken
-     * @param  {String}  uuid       UUID of token to validate
-     * @return {Promise}
-     */
-    validateToken(uuid) {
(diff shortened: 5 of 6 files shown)


lib/sealField.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+const iron = require('iron');
+const nodeify = require('./nodeify');
+
+/**
+ * Seal a field
+ * @method sealField
+ * @param  {Object}         object              The object to seal
```
FenrirUnbound: `object` isn't very descriptive. Call it `model`

lib/sealField.js
```diff
@@ -0,0 +1,27 @@
...
+ * @param  {Object}         config
+ * @param  {Mixed}          config.field        The field to store the sealed value in
+ * @param  {String}         [config.value]      The value to seal. Defaults to the current value of the field
```
FenrirUnbound: Why bother?

lib/sealField.js
ian-fox: It's not necessarily a model, though I don't know where else it'd be used. Changed.

lib/sealField.js
ian-fox: Was thinking about the case where you set a field to an unsealed value, called seal, iron errored, and then called update. Wouldn't want to save the unsealed value in the datastore.

Though now I realize that no code should just blindly be calling update, if iron fails it should go to a catch block of some sort.

ian-fox:
Oh dear, it's failing now because it's using the new data schema and this commit was just the refactor, so it doesn't have the token model fixes. I should have realized that'd happen.

lib/sealField.js
```diff
@@ -0,0 +1,24 @@
...
+
+const iron = require('iron');
+const nodeify = require('./nodeify');
+
+/**
+ * Seal a field
+ * @method sealField
+ * @param  {Object}         model               The model to seal a property of
+ * @param  {Object}         config
+ * @param  {Mixed}          config.field        The field to store the sealed value in
```
FenrirUnbound: When would you not use a `String` for this parameter?

ian-fox: When the field is a symbol, like how we do passwords as Symbol('password')

FenrirUnbound: Each instance where it's being used, it's being passed a `String`. 

* https://github.com/screwdriver-cd/models/pull/162/files#diff-326dff258ee0c2713dca9ab6a5d701d3R32
* https://github.com/screwdriver-cd/models/pull/162/files#diff-db62f6a8d7d20845993eb9a470a318dfR67
* https://github.com/screwdriver-cd/models/pull/162/files#diff-398ed3b014436a5204583323ea29320bR48
* https://github.com/screwdriver-cd/models/pull/162/files#diff-1fb87bb8462436f368b6060fd2cfb374R51

EDIT: formatting

test/lib/sealField.test.js
```diff
@@ -0,0 +1,79 @@
...
+const symbolField = Symbol('symbol field');
+
+sinon.assert.expose(assert, { prefix: '' });
+
+describe('sealField', () => {
+    const password = 'aPasswordThatHasToBeAtLeast32CharactersLong';
+    const sealed = 'sealed value';
+    let ironMock;
+    let sealField;
+    let object;
```
FenrirUnbound: There's a strong preference to not name a variable something generic like `object`. 

1. Best practices -- `object` is subtly one case from being `Object`, which is both a reserved word and is its own object. It's the same reason why we avoid `error` for Errors.


2. It's vague -- It's not very descriptive of what the variable represents. You have to look at how it's being used to figure out what it truly represents, which requires at least one level of indirection when debugging.

test/lib/token.test.js
```diff
@@ -24,24 +27,45 @@ describe('Token Model', () => {
...
             description: 'For the mobile app',
             lastUsed: '2017-05-10T01:49:59.327Z',
             password
         };
+
+        datastore.update.resolves({});
+        ironMock.seal.yieldsAsync(null, sealed);
+        ironMock.unseal.yieldsAsync(null, unsealed);
+
+        mockery.registerMock('iron', ironMock);
```
FenrirUnbound: You want to be mocking the `sealField` dependency instead of the `iron` dependency that is used in `sealField`.

ian-fox:
Marked as WIP because I have some questions to ask about login flows which will affect models. Specifically, would affect if the user model still has the validateToken method.

test/lib/tokenFactory.test.js
```diff
@@ -31,23 +36,28 @@ describe('Token Factory', () => {
...
         };
+        ironMock = {
+            seal: sinon.stub(),
+            unseal: sinon.stub(),
+            defaults: 'defaults'
+        };
+    });
+
+    beforeEach(() => {
+        mockery.registerMock('iron', ironMock);
```
FenrirUnbound: Same thing here. You want to be mocking `sealField` instead.

FenrirUnbound (changes requested):

test/lib/sealField.test.js
```diff
@@ -0,0 +1,79 @@
...
+                assert.strictEqual(object[symbolField], 'sealed value');
+            }));
+
+    it('allows errors through', () => {
+        const err = new Error('an error');
+
+        ironMock.seal.yieldsAsync(err);
+
+        return sealField(object, { field: 'field1', password })
+            .then(() => {
```
FenrirUnbound: Could also be written as 

```
    .then(() => assert.fail('reason'), (e) => {
        // assert error case
    })
```

The way it's currently written, it will throw an error when the `assert.fail()` is reached. Then the `.catch` will have the "Should not get here" error instead of the true error that `sealField()` yields.

ian-fox: Oh yeah, fail throws an error you're right.

test/lib/token.test.js
ian-fox: Roger 👍

lib/sealField.js
ian-fox: I was thinking towards if we ever used it for other things in the future, but I can see how the logic about levels of indirection while debugging applies here too. Probably better to have it be more descriptive, and modify it if at some point in the future modifications are needed.

PR #163, opened 2017-05-24
fix: Tokens store hash value

ian-fox:
* Use sha256 to hash a token value, and store the hash
* When generating (or regenerating) a token, the first time the model is returned it will have a "value" field.
* Can `get` a token by value, and the factory will hash it so the api doesn't have to worry about it

### DIFF
--- a/lib/tokenFactory.js
+++ b/lib/tokenFactory.js
@@ -2,6 +2,7 @@
 
 const BaseFactory = require('./baseFactory');
 const Token = require('./token');
+const generateToken = require('./generateToken');
 
 let instance;
 
@@ -27,19 +28,45 @@ class TokenFactory extends BaseFactory {
     }
 
     /**
-     * Create a token model
+     * Create a token model, returning the model including the unhashed value
      * @method create
      * @param  {Object}     config
      * @param  {String}     config.userId            The ID of the associated user
-     * @param  {String}     config.value             The token value
      * @param  {String}     config.name              The token name
      * @param  {String}     config.description       The token description
      * @return {Promise}
      */
     create(config) {
-        config.lastUsed = null;
+        let value;
 
-        return super.create(config);
+        return generateToken.generateValue()
+            .then((bytes) => {
+                value = bytes;
+                config.hash = generateToken.hashValue(bytes);
+                config.lastUsed = null;
+
+                return super.create(config);
+            }).then((model) => {
+                model.value = value;
+
+                return model;
+            });
+    }
+
+    /**
+     * Get a token
+     * @method get
+     * @param  {Mixed}     config
+     * @param  {String}    [config.value]    (Un-hashed) value of token to look for
+     * @return {Promise}
+     */
+    get(config) {
+        if (config && config.value) {
+            config.hash = generateToken.hashValue(config.value);
+            delete config.value;
+        }
+
+        return super.get(config);
     }
 
     /**
--- a/test/lib/generateToken.test.js
+++ b/test/lib/generateToken.test.js
@@ -1,0 +1,38 @@
+'use strict';
+
+const assert = require('chai').assert;
+const sinon = require('sinon');
+const generateToken = require('../../lib/generateToken');
+
+sinon.assert.expose(assert, { prefix: '' });
+require('sinon-as-promised');
+
+describe('generateToken', () => {
+    const RANDOM_BYTES = 'some random bytes';
+    // Result of passing 'some random bytes' through a sha256 hash, in base64
+    const HASH = 'mF0EvjvxWMrVz5ZGJcnbe0ZPooUlv/DAB9VrV6bmZmg=';
+    let firstValue;
+
+    it('generates a value', () =>
+        generateToken.generateValue()
+            .then((value) => {
+                firstValue = value;
+                // Check that it's a base64 value of the right length
+                assert.match(value, /[a-zA-Z0-9+/]{43}=/);
+            }));
+
+    it('generates a different value on a second call', () => {
+        generateToken.generateValue()
+            .then((value) => {
+                assert.notEqual(value, firstValue);
+            });
+    });
+
+    it('hashes a value', () => {
+        assert.strictEqual(generateToken.hashValue(RANDOM_BYTES), HASH);
+    });
+
+    it('hashes a different value to a different hash', () => {
+        assert.notEqual(generateToken.hashValue('some different bytes'), HASH);
+    });
+});
--- a/test/lib/tokenFactory.test.js
+++ b/test/lib/tokenFactory.test.js
@@ -10,18 +10,28 @@ require('sinon-as-promised');
 describe('Token Factory', () => {
     const name = 'mobile_token';
     const description = 'a token for a mobile app';
-    const uuid = 'abc123';
     const userId = 6789;
     const tokenId = 12345;
+    const randomBytes = 'some random bytes';
+    const hash = 'abc123';
     const tokenData = {
         id: tokenId,
         userId,
         description,
         name,
+        hash,
+        lastUsed: null
+    };
+    const expected = {
+        userId,
+        name,
+        description,
+        hash,
         lastUsed: null
     };
     let TokenFactory;
     let datastore;
+    let generateTokenMock;
     let factory;
     let Token;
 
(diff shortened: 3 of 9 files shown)


test/lib/generateToken.test.js
```diff
@@ -0,0 +1,56 @@
...
+sinon.assert.expose(assert, { prefix: '' });
+require('sinon-as-promised');
+
+describe('generateToken', () => {
+    let nodeifyMock;
+    let generateToken;
+    const randomBytes = 'some random bytes';
+    // Can't mock crypto because code coverage tool uses it too
+    // This is what a sha256 hash turns the random bytes into
+    const hash = 'mF0EvjvxWMrVz5ZGJcnbe0ZPooUlv/DAB9VrV6bmZmg=';
```
ian-fox: I tried a few things but wasn't able to get mockery to register the `crypto` mock for the generateToken file but not the code coverage tool. If somebody knows a way to do this please let me know.

test/lib/user.test.js
```diff
@@ -219,48 +219,4 @@ describe('User Model', () => {
             assert.calledOnce(tokenFactoryMock.list);
         });
     });
-
```
ian-fox, line 222: We don't have a validateToken on the tokens any more, in the PR after this (ready for once this one is finished review) I added a way to get a user by token value, so the login flow for that will be the same as getting a user with oauth and then giving them a 12h token.

test/lib/generateToken.test.js
petey: In cases like that you can use `sinon.spy()` to watch the method(s), and have the same functionality for testing that those methods were called as expected.

ian-fox: I was more concerned with spoofing the returned values, because if I understand correctly I should be testing the `generateToken` file here, not `crypto`. Though `crypto` is built into node, so it's probably a safe bet I would think.

petey: Let's consider it a run-time dependency (like `async`) since it can't be mocked out. You should still inspect that it was called with the expected parameters.

ian-fox: Sounds good :+1:

ian-fox: Found an even better solution: can stub things without registering the entire module as a mock.

This means we can stub the first call to createHash, and have it call the original for all subsequent ones so it doesn't affect the coverage.

test/lib/tokenFactory.test.js
```diff
@@ -31,11 +40,21 @@ describe('Token Factory', () => {
             useCleanCache: true,
             warnOnUnregistered: false
         });
+        generateTokenMock = {
+            generateValue: sinon.stub(),
+            hashValue: sinon.stub()
```
minzcmu, line 45: @FenrirUnbound Should we mock the `hashValue` function here? Making sure the `hashValue` function returns the same hash for the same string is important. And not mocking this should not fail the test. But the crypto is from an external module, I am not sure if we should mock it or not.

ian-fox: That bit is tested in `generateToken.test.js`

FenrirUnbound: The mock that's here is fine. `generateToken.test.js` will handle how that particular test works out.

tkyi:
Add README documentation for Tokens?

test/lib/generateToken.test.js
```diff
@@ -0,0 +1,56 @@
...
+    before(() => {
+        mockery.enable({
+            useCleanCache: true,
+            warnOnUnregistered: false
+        });
+        nodeifyMock = {
+            withContext: sinon.stub()
+        };
+
+        mockery.registerMock('./nodeify', nodeifyMock);
```
petey: nodeify is a runtime dependency that shouldn't be mocked. You can spy on it if you want to.

test/lib/generateToken.test.js
```diff
@@ -0,0 +1,56 @@
...
+
+    after(() => {
+        mockery.deregisterAll();
+        mockery.disable();
+    });
+
+    it('generates a value', () =>
+        generateToken.generateValue()
+            .then((value) => {
+                assert.strictEqual(value, randomBytes);
```
petey: You can verify that the value is the right number of characters, and probably that it is base64 encoded, but not a strict value comparison.

You should verify that if this function is run twice it doesn't return the same value.

test/lib/generateToken.test.js
```diff
@@ -0,0 +1,56 @@
...
+const sinon = require('sinon');
+const crypto = require('crypto');
+
+sinon.assert.expose(assert, { prefix: '' });
+require('sinon-as-promised');
+
+describe('generateToken', () => {
+    const randomBytes = 'some random bytes';
+    // Result of passing 'some random bytes' through a sha256 hash, in base64
+    const hash = 'mF0EvjvxWMrVz5ZGJcnbe0ZPooUlv/DAB9VrV6bmZmg=';
```
petey: These constants should be all uppercase.

test/lib/generateToken.test.js
```diff
@@ -0,0 +1,56 @@
...
+
+    it('generates a value', () =>
+        generateToken.generateValue()
+            .then((value) => {
+                assert.strictEqual(value, randomBytes);
+                assert.calledWith(nodeifyMock.withContext, crypto, 'randomBytes', [32]);
+            }));
+
+    it('hashes a value', () => {
+        assert.strictEqual(generateToken.hashValue(randomBytes), hash);
```
petey: Should probably verify that a different string does not produce the same hash. All this test says is that the function is returns the value of `hash`

ian-fox:
Ooh, good idea.

FenrirUnbound (approved):
👍 

Seems fine to me 

zxcv

petey (changes requested):

test/lib/generateToken.test.js
ian-fox: Not mocking nodeify makes the testing a lot more interesting :+1:

minzcmu (approved):
image

ian-fox:
Implemented the changes Petey requested, don't want to bother him on his time off to verify them (especially since they were pretty straightforward)

lib/tokenFactory.js
```diff
@@ -27,19 +28,45 @@ class TokenFactory extends BaseFactory {
...
+            .then((bytes) => {
+                value = bytes;
+                config.hash = generateToken.hashValue(bytes);
+                config.lastUsed = null;
+
+                return super.create(config);
+            }).then((model) => {
+                model.value = value;
+
+                return model;
```
d2lam: This would include the `hash`. Any reason why the user needs to know it?

lib/tokenFactory.js
```diff
@@ -27,19 +28,45 @@ class TokenFactory extends BaseFactory {
...
+            });
+    }
+
+    /**
+     * Get a token
+     * @method get
+     * @param  {Mixed}     config
+     * @param  {String}    [config.value]    (Un-hashed) value of token to look for
```
d2lam, line 60: Why would this be optional?

lib/tokenFactory.js
```diff
@@ -27,19 +28,45 @@ class TokenFactory extends BaseFactory {
...
+     * @return {Promise}
+     */
+    get(config) {
+        if (config && config.value) {
+            config.hash = generateToken.hashValue(config.value);
+            delete config.value;
+        }
+
+        return super.get(config);
```
d2lam, line 69: Just a note that would include the `hash`, and it's fine. But from the API, I think it's better we don't show it.

lib/tokenFactory.js
ian-fox: The other methods of getting stuff still work (integer id, or config.id). I should maybe put those in the params

lib/tokenFactory.js
ian-fox: You mean like delete the `hash` field before returning it?

We can't do that in the model layer, it doesn't let you delete the fields that go in the datastore (unless we want to mark the field as one that can be deleted when creating the model, but I'm not sure what effect that would have if you tried to call save)

For secrets it looks like the value is deleted in the API layer before they're returned.

PR #165, opened 2017-05-24
fix: move build createTime closer to datastore save

d2lam:
Seeing some weird data where createTime is after startTime. 
This does not address that problem but moving it closer to datastore save to be more accurate.


minzcmu (approved):
👍

PR #166, opened 2017-05-25
feat: Get a user by token value

ian-fox:
* Can call userFactory.get with an unhashed token value in `token`
* Getting a user this way update's the token's "last used" property

### DIFF
--- a/README.md
+++ b/README.md
@@ -417,6 +417,7 @@ factory.create(config).then(model => {
 
 #### Get
 Get a user based on id. Can pass the generatedId for the user, or the username, and the id will be determined automatically.
+Can also pass a Screwdriver access token, and will get the user associated with that token.
 ```js
 factory.get(id).then(model => {
     // do stuff with user model
@@ -425,12 +426,17 @@ factory.get(id).then(model => {
 factory.get({ username }).then(model => {
     // do stuff with user model
 });
+
+factory.get({ token }).then(model => {
+    // do stuff with user model
+});
 ```
 
 | Parameter        | Type  |  Description |
 | :-------------   | :---- | :-------------|
 | id | String | The unique ID for the build |
 | config.username | String | User name |
+| config.accessToken | String | A user access token value |
 
 ### User Model
 ```js
--- a/lib/userFactory.js
+++ b/lib/userFactory.js
@@ -65,6 +65,38 @@ class UserFactory extends BaseFactory {
     }
 
     /**
+     * Get a user model by name, id, or personal access token
+     * @method get
+     * @param   {Mixed}     config
+     * @param   {String}    [config.username]       Username of the user to look up
+     * @param   {String}    [config.accessToken]    Access token of the user to look up
+     * @return  {Promise}
+     */
+    get(config) {
+        if (!config.accessToken) {
+            return super.get(config);
+        }
+
+        // Lazy load factory dependency to prevent circular dependency issues
+        // https://nodejs.org/api/modules.html#modules_cycles
+        /* eslint-disable global-require */
+        const TokenFactory = require('./tokenFactory');
+        /* eslint-enable global-require */
+        const tokenFactory = TokenFactory.getInstance();
+
+        return tokenFactory.get({ value: config.accessToken })
+            .then((token) => {
+                if (!token) {
+                    return token;
+                }
+
+                token.lastUsed = (new Date()).toISOString();
+
+                return token.update().then(() => this.get(token.userId));
+            });
+    }
+
+    /**
      * Get an instance of the UserFactory
      * @method getInstance
      * @param  {Object}     config
--- a/test/lib/userFactory.test.js
+++ b/test/lib/userFactory.test.js
@@ -5,6 +5,7 @@ const mockery = require('mockery');
 const sinon = require('sinon');
 
 sinon.assert.expose(assert, { prefix: '' });
+require('sinon-as-promised');
 
 describe('User Factory', () => {
     const password = 'totallySecurePassword';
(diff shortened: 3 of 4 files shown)


lib/userFactory.js
```diff
@@ -65,6 +65,38 @@ class UserFactory extends BaseFactory {
...
 
     /**
+     * Get a user model by name, id, or token
+     * @method get
+     * @param   {Mixed}     config
+     * @param   {String}    [config.token]      Token of the user to look up
+     * @return  {Promise}
+     */
+    get(config) {
+        if (!config || !config.token) {
```
d2lam: Shouldn't `config` always be there? What would `get()` return? 
It has to be either an `id` or `object` right?

ian-fox: You're right, force of habit on my part.

lib/userFactory.js
```diff
@@ -65,6 +65,38 @@ class UserFactory extends BaseFactory {
...
+        }
+
+        // Lazy load factory dependency to prevent circular dependency issues
+        // https://nodejs.org/api/modules.html#modules_cycles
+        /* eslint-disable global-require */
+        const TokenFactory = require('./tokenFactory');
+        /* eslint-enable global-require */
+        const tokenFactory = TokenFactory.getInstance();
+
+        return tokenFactory.get({ value: config.token })
```
d2lam: Would this work? From `baseFactory`, it looks like you can only search by fields that are in `keys`: https://github.com/screwdriver-cd/models/blob/master/lib/baseFactory.js#L73-L77

And `value` is not in token's keys: https://github.com/screwdriver-cd/data-schema/blob/master/models/token.js#L102

ian-fox: All of the hashing is handled by tokens, so I made the factory able to accept `value` which it will then turn into a `hash` param before actually doing the search.

https://github.com/screwdriver-cd/models/blob/user-tokens/lib/tokenFactory.js#L63

d2lam: Okay cool! 👍  The other PR makes more sense now

lib/userFactory.js
```diff
@@ -65,6 +65,38 @@ class UserFactory extends BaseFactory {
     }
 
     /**
+     * Get a user model by name, id, or token
+     * @method get
+     * @param   {Mixed}     config
+     * @param   {String}    [config.token]      Token of the user to look up
```
d2lam: should also add `config.username` to this

test/lib/userFactory.test.js
```diff
@@ -5,13 +5,15 @@ const mockery = require('mockery');
 const sinon = require('sinon');
 
 sinon.assert.expose(assert, { prefix: '' });
+require('sinon-as-promised');
```
FenrirUnbound, line 8: `sinon` stubs support Promises

lib/userFactory.js
```diff
@@ -65,6 +65,38 @@ class UserFactory extends BaseFactory {
...
+        return tokenFactory.get({ value: config.token })
+            .then((token) => {
+                if (!token) {
+                    return token;
+                }
+
+                token.lastUsed = (new Date()).toISOString();
+
+                return token.update()
+                    .then(() => this.get(token.userId));
```
d2lam: This doesn't need to be in this scope. I think it'll be cleaner like this:
```
.then((token) => {
    .... 
    return token.update();
})
.then(updatedToken => this.get(updatedToken.userId));

File: test/lib/userFactory.test.js (hunk 4 above)
Comment (ian-fox) [reply test/lib/userFactory.test.js:8]: Sinon version 2 does, we're still using 1.17.4. There are a lot of places using `sinon-as-promised` that could be converted to just using `sinon`, if we want to switch to sinon 2 I feel like that should be a separate PR.

File: lib/userFactory.js (hunk 3 above)
Comment (minzcmu) [reply lib/userFactory.js]: I think it's okay not to specify config.username, you can also pass in `id` to get a user model. Technically you can pass in any valid key here to retrieve the model.

Comment (minzcmu) [inline lib/userFactory.js]: Can we make it clear that the `token` here refers to `screwdriver personal access token`. There is also `token` in the same file referring to the `github token` . Also the README please

**Review by minzcmu — CHANGES_REQUESTED** [review]

File: lib/userFactory.js (hunk 5 above)
Comment (ian-fox) [reply lib/userFactory.js]: If it's out there it'll also run after the case where we return `null` if the token isn't found, and then it will error when it tries to get `null.userId`.

Moved to the same line as `return token.update()` so that it looks slightly prettier though.

File: README.md [hunk 6]
```diff
@@ -417,6 +417,7 @@ factory.create(config).then(model => {
 
 #### Get
 Get a user based on id. Can pass the generatedId for the user, or the username, and the id will be determined automatically.
+Can also pass an access token, and will get the user associated with that token.
```
tkyi: pass a **Screwdriver** access token

README.md
```diff
@@ -425,12 +426,17 @@ factory.get(id).then(model => {
...
+factory.get({ token }).then(model => {
+    // do stuff with user model
+});
 ```
 
 | Parameter        | Type  |  Description |
 | :-------------   | :---- | :-------------|
 | id | String | The unique ID for the build |
 | config.username | String | User name |
+| config.token | String | A user access token value |
```
minzcmu: Need to update this to accessToken?

minzcmu (changes requested):

README.md
ian-fox: Oh right of course, I even have a sticky reminding me not to forget readme's :P

d2lam (approved):

minzcmu (approved):

PR #167, opened 2017-05-26
refactor(tests): Upgrade to Sinon V2

ian-fox:
* Sinon version 2 supports promises, so we no longer need `sinon-as-promised`

### DIFF
--- a/test/lib/job.test.js
+++ b/test/lib/job.test.js
@@ -7,7 +7,6 @@ const schema = require('screwdriver-data-schema');
 const hoek = require('hoek');
 
 sinon.assert.expose(assert, { prefix: '' });
-require('sinon-as-promised');
 
 describe('Job Model', () => {
     let pipelineFactoryMock;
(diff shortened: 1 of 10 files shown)


test/lib/job.test.js
```diff
@@ -305,7 +304,7 @@ describe('Job Model', () => {
         });
 
         it('fail if getBuilds returns error', () => {
-            buildFactoryMock.list.rejects('error');
+            buildFactoryMock.list.rejects(new Error('error'));
```
ian-fox, line 307: In `sinon-as-promised`, if you call `rejects` with a string it turns it into an error with the string as the message. `sinon` turns it into an error with the string as the type, so if we're checking against messages we need to construct the error object ourselves.

tkyi (approved):
Nice!

FenrirUnbound (approved):
👍 

Looks fine to me

wateryanguishedcrayfish

PR #169, opened 2017-06-02
fix(tokens): lastUsed defaults to empty string instead of null

ian-fox:
The datastore wasn't validating tokens that hadn't been used, because the `.allow(null)` was being ignored by the datastore-sequelize plugin. Now using empty strings instead of `null`.


tkyi (approved):
LGTM

minzcmu (approved):

PR #170, opened 2017-06-06
fix(tokens): Update tokens to be URL-safe

ian-fox:
* The endpoint for auth takes tokens through query string, or a bearer header
* Bearer headers are used for JWT's. We could try to pattern match and handle tokens differently based on if they're user tokens or JWT's, but that seems suspect
* Currently tokens are encoded in base64, which is not URL-safe
* This PR switches to using hex, which *is* URL-safe

More discussion at https://github.com/screwdriver-cd/data-schema/pull/139

### DIFF
--- a/lib/generateToken.js
+++ b/lib/generateToken.js
@@ -3,7 +3,7 @@
 const nodeify = require('./nodeify');
 const crypto = require('crypto');
 const ALGORITHM = 'sha256';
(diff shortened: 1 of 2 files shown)


ian-fox:
Closing as more investigation is needed into whether we can use the authorization header with both the hapi-bearer and hapi-jwt plugins.

PR #171, opened 2017-06-06
fix(tokens): Use URL-safe values, add toJson

ian-fox:
* add `toJson` to hide the hash value, since the only time the API ever even knows a hash value exists is to delete it before returning the token, and so that `value` doesn't have to be added manually by the API
* switch base64 encoding to be URL-safe.
    - Could use a library like `base64url`, but the same thing can be accomplished by a small function in one file, which doesn't add a new dependency.
    - If adding the dependency for increased code-cleanliness is preferred I can switch to doing that.
* test still had `password` in it, we no longer use that in tokens

### DIFF
--- a/lib/generateToken.js
+++ b/lib/generateToken.js
@@ -2,8 +2,8 @@
 
 const nodeify = require('./nodeify');
 const crypto = require('crypto');
+const base64url = require('base64url');
 const ALGORITHM = 'sha256';
-const ENCODING = 'base64';
 const TOKEN_LENGTH = 256; // Measured in bits
 
 /**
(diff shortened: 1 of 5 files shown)


lib/generateToken.js
```diff
@@ -7,13 +7,24 @@ const ENCODING = 'base64';
 const TOKEN_LENGTH = 256; // Measured in bits
 
 /**
+ * Replace default base64 characters with ones that are url-safe
+ * @param  {String} token
```
minzcmu: should have `@method`

minzcmu (changes requested):

ian-fox:
Just now saw @FenrirUnbound's comment [here](https://github.com/screwdriver-cd/data-schema/pull/140#discussion_r120495728), switching to use the library.

r3rastogi (approved):
Looks good

tkyi (approved):

minzcmu (approved):

PR #175, opened 2017-06-20
feat(472): configure a default docker registry for container images

FenrirUnbound:
## Context

The typical behavior expected when a Docker image lacks an explicit registry, the default registry it uses is Docker Hub. For example, specifying `node:6` translates to `registry.hub.docker.com/library/node:6`.

This change allows for the default registry to be something other than Docker Hub (e.g., a private registry) when an image name lacks an explicit registry.

## Objective

Use a different Docker registry when the image is missing an explicit registry.

## Reference

* Related to screwdriver-cd/screwdriver#472

### DIFF
--- a/lib/buildFactory.js
+++ b/lib/buildFactory.js
@@ -2,6 +2,7 @@
 
 const BaseFactory = require('./baseFactory');
 const Build = require('./build');
+const imageParser = require('docker-parse-image');
 let instance;
 
 /**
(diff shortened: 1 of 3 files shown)


lib/buildFactory.js
```diff
@@ -63,18 +64,46 @@ function getCommitInfo(config) {
...
+ * When the image name contains an explicit registry, it returns the passed-in
+ * container value. When a default registry is _not_ configured for the BuildFactory,
+ * it simply returns the passed-in container value.
+ * @method dockerImageName
+ * @param  {Object}        config
+ * @param  {String}        config.container         Docker image name to use.
+ * @param  {String}        [config.dockerRegistry]  Docker registry where the image is stored. Defaults to Docker Hub
+ * @return {String}                                 The full image name to use
+ */
+function dockerImageName({ container, dockerRegistry }) {
```
d2lam, line 80: Cool! Didn't know we could do this!

lib/buildFactory.js
```diff
@@ -152,7 +181,11 @@ class BuildFactory extends BaseFactory {
                         modelConfig.commit = data.decoratedCommit;
                     }
 
-                    modelConfig.container = permutation.image;
+                    modelConfig.container = dockerImageName({
+                        container: permutation.image,
+                        dockerRegistry: this.dockerRegistry
+                    });
+                    // modelConfig.container = permutation.image;
```
d2lam: Removed unused code

d2lam (changes requested):

lib/buildFactory.js
FenrirUnbound: You passed ✅  my test 😸

ian-fox (approved):
Looks good :+1:

d2lam (approved):
Cool! 👍

d2lam (approved):

PR #176, opened 2017-06-20
fix(562): Update pipelines with new repo and branch

tkyi:
## Context
Users currently cannot update the repo/branch for their pipeline. It would be nice if users could update that information without having to create a new pipeline (and lose their pipeline ID).

## Objective
This PR adds a method for updating a pipeline. It will check if the `scmUri` value is dirty. If so, it will get a user token, decorate the URL to create an `scmRepo` object, and update the pipeline with the new information.

`scmRepo` object: 
```
"scmRepo": {
"branch": "master",
"name": "screwdriver-cd/screwdriver",
"url": "https://github.com/screwdriver-cd/screwdriver/tree/master"
}
```

## References
Data-schema PR: https://github.com/screwdriver-cd/data-schema/pull/148
Original issue: https://github.com/screwdriver-cd/screwdriver/issues/562

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -497,6 +497,34 @@ class PipelineModel extends BaseModel {
     }
 
     /**
+     * Update the repository and branch
+     * @method update
+     * @return {Promise}
+     */
+    update() {
+        if (this.isDirty('scmUri')) {
+            // Lazy load factory dependency to prevent circular dependency issues
+            // https://nodejs.org/api/modules.html#modules_cycles
+            /* eslint-disable global-require */
+            const UserFactory = require('./userFactory');
+            /* eslint-enable global-require */
+
+            const userFactory = UserFactory.getInstance();
+
+            return userFactory.get({ username: Object.keys(this.admins)[0] })
+                .then(user => user.unsealToken())
+                .then(token => this.scm.decorateUrl({ scmUri: this.scmUri, token }))
+                .then((scmRepo) => {
+                    this.scmRepo = scmRepo;
+
+                    return super.update();
+                });
+        }
+
+        return super.update();
+    }
+
+    /**
      * Remove all jobs & builds associated with this pipeline and the pipeline itself
      * @return {Promise}        Resolves to null if remove successfully
      */
--- a/test/lib/pipeline.test.js
+++ b/test/lib/pipeline.test.js
@@ -775,6 +775,73 @@ describe('Pipeline Model', () => {
         });
     });
 
+    describe('update', () => {
+        const scmRepo = {
+            name: 'foo/bar',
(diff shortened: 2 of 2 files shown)


test/lib/pipeline.test.js
```diff
@@ -775,6 +775,101 @@ describe('Pipeline Model', () => {
...
+                    admins: { d2lam: true },
+                    id: 'd398fb192747c9a0124e9e5b4e6e8e841cf8c71c'
+                },
+                table: 'pipelines'
+            };
+            const scmRepo = {
+                branch: 'master',
+                name: 'screwdriver-cd/screwdriver',
+                url: 'https://github.com/screwdriver-cd/screwdriver/tree/master'
+            };
```
minzcmu: the `scmRepo` in this test case is the same as the previous one. Can we re-use it?

test/lib/pipeline.test.js
```diff
@@ -775,6 +775,101 @@ describe('Pipeline Model', () => {
...
+
+            scmMock.decorateUrl.resolves(scmRepo);
+            userFactoryMock.get.withArgs({ username: 'd2lam' }).resolves({
+                unsealToken: sinon.stub().resolves('foo')
+            });
+            datastore.update.resolves({
```
minzcmu: And this can resolve `null` since we are not testing the resolved value

minzcmu (changes requested):

lib/pipeline.js
```diff
@@ -497,6 +497,34 @@ class PipelineModel extends BaseModel {
...
+
+            const userFactory = UserFactory.getInstance();
+
+            return userFactory.get({ username: Object.keys(this.admins)[0] })
+                .then(user => user.unsealToken())
+                .then(token => this.scm.decorateUrl({ scmUri: this.scmUri, token }))
+                .then((scmRepo) => {
+                    this.scmRepo = scmRepo;
+
+                    return super.update();
```
FenrirUnbound, line 520: Does `this.update()` not work here? I'm curious why it has to be `super.update()`

tkyi: I think they both work. I think I had that earlier. I might have accidentally crtl+z'ed too far.

tkyi: Okay I'm getting some weird error when I use `this.update()` so I'm going to leave it at `super.update()`

d2lam: Wouldn't `this.update()` go back to THIS function?
`super.update()` would go to https://github.com/screwdriver-cd/models/blob/master/lib/base.js#L70, which is what we want.

d2lam (approved):

minzcmu (approved):

PR #178, opened 2017-07-10
feat(523): Add collection model and factory with tests

joelseq:
# Context

To be able to perform CRUD operations for collections, we need a CollectionModel and CollectionFactory to interface with the datastore.

# Objective

This PR adds a CollectionModel and CollectionFactory along with some tests

# References

issue: [screwdriver-cd/screwdriver#523](https://github.com/screwdriver-cd/screwdriver/issues/523)

This is the first pass for the CollectionModel and CollectionFactory so it will change when I start working on adding / updating / removing pipelines from Collections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -883,6 +883,92 @@ token.refresh()
     });
 ```
 
+### Collection Factory
+#### Search
+```js
+'use strict';
+const Model = require('screwdriver-models');
+const factory = Model.CollectionFactory.getInstance({
+    datastore
+});
+const config = {
+    params: {
+        userId: 12345
+    }
+};
+
+factory.list(config).then(collections => {
+    // Do stuff with list of collections
+});
+```
+
+| Parameter     | Type   | Description         |
+| :-------------| :------| :-------------------|
+| config        | Object | Config object       |
+| config.params | Object | Fields to search on |
+
+#### Create
+```js
+factory.create(config).then(model => {
+    // do stuff with collection model
+});
+```
+
+| Parameter          | Type   | Required | Description                                              |
+| :------------------| :------| :--------| :--------------------------------------------------------|
+| config             | Object | Yes      | Configuration Object                                     |
+| config.userId      | Number | Yes      | User that this collection belongs to                     |
+| config.name        | String | Yes      | Collection name                                          |
+| config.description | String | No       | Collection description                                   |
+| config.pipelineIds | Array  | No       | List of ids of pipelines associated with this collection |
+
+#### Get
+Get a collection based on unique id of collection. Can also pass in a combination of userId and collection name, and the id will be determined automatically.
+```js
+factory.get(id).then(model => {
+    // do stuff with collection model
+});
+
+factory.get({ userId, name }).then(model => {
+    // do stuff with collection model
+})
+```
+
+| Parameter | Type   | Description                      |
+| :---------| :------| :--------------------------------|
+| id        | Number | The unique id for the collection |
+
+### Collection Model
+```js
+'use strict';
+const Model = require('screwdriver-models');
+const factory = Model.CollectionFactory.getInstance({
+    datastore
+});
+const config = {
+    userId: 12345,
+    name: 'Screwdriver',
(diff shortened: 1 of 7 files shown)


index.js
```diff
@@ -17,5 +18,6 @@ module.exports = {
     SecretFactory,
     UserFactory,
     TemplateFactory,
-    TokenFactory
+    TokenFactory,
+    CollectionFactory
```
ian-fox: These should probably be in alphabetical order (I know they weren't before, that's my fault)

lib/collection.js
```diff
@@ -0,0 +1,20 @@
...
+class CollectionModel extends BaseModel {
+    /**
+     * Construct a CollectionModel object
+     * @method constructor
+     * @param {Object} config
+     * @param {Object} config.datastore     Object that will perform operations on the datastore
+     * @param {Number} config.userId        The ID of the associated user
+     * @param {String} config.name          The collection name
+     * @param {String} config.description   The collection description
+     */
```
ian-fox: no `pipelineIds`?

ian-fox (changes requested):

lib/collection.js
joelseq: whoops, my bad!

ian-fox (approved):
:+1:!

lib/collection.js
```diff
@@ -0,0 +1,21 @@
...
+class CollectionModel extends BaseModel {
+    /**
+     * Construct a CollectionModel object
+     * @method constructor
+     * @param {Object} config
+     * @param {Object} config.datastore     Object that will perform operations on the datastore
+     * @param {Number} config.userId        The ID of the associated user
+     * @param {String} config.name          The collection name
+     * @param {String} config.description   The collection description
+     * @param {Array}  config.pipelineIds   The ids of the pipelines associated with this collection
```
minzcmu: https://github.com/screwdriver-cd/data-schema/blob/master/models/collection.js
`description` and `pipelineIds` are optional for `create` based on your data-schema. But `pipelineIds` is required for `get`. So if user don't pass `pipelineIds` in when they create the collection, you should give it a default value as empty array either here or in the api. Otherwise there will be err when you try to get a collection without value for pipelineIds

Use `[config.description]` if it's optional.

minzcmu (changes requested):

minzcmu (changes requested):
update README?

minzcmu (commented):
https://github.com/screwdriver-cd/data-schema/blob/master/models/collection.js#L89

I am confused by this line in `data-schema`. `id` is created by auto-increment and should be unique by itself already. If the `key` is set as `id`, then you allow user to have collections of same names?

joelseq:
@minz1027 Oh my bad, my thought process was that there might be multiple people that create a collection called `Team` for example so in that case, it shouldn't cause issues in the collections table due to name having to be unique. Let me know what is the best thing to set for `key`.

minzcmu:
@joelseq I think `userId` and `name` together will be a good key. This PR LGTM 👍

minzcmu (approved):

README.md
```diff
@@ -883,6 +883,88 @@ token.refresh()
...
+| Parameter          | Type   | Required | Description                                              |
+| :------------------| :------| :--------| :--------------------------------------------------------|
+| config             | Object | Yes      | Configuration Object                                     |
+| config.userId      | Number | Yes      | User that this collection belongs to                     |
+| config.name        | String | Yes      | Collection name                                          |
+| config.description | String | No       | Collection description                                   |
+| config.pipelineIds | Array  | No       | List of ids of pipelines associated with this collection |
+
+#### Get
+Get a collection based on unique id of collection.
```
ian-fox: If userId+name also uniquely determine a collection, those can be passed in here too. See https://github.com/screwdriver-cd/models/blob/152eb81e9aa3cc40dac8f3b44854a2b629853241/README.md#get-1

ian-fox (approved):

PR #179, opened 2017-07-12
feat(616): Add template tag model

tkyi:
## Context

Users would like to tag their template versions with words such as "latest" or "stable" (much like Docker tags work).

## Objective

This PR adds a TemplateTag and TemplateTagFactory model.

## Related Links
Blocked by: https://github.com/screwdriver-cd/data-schema/pull/157
Original issue: https://github.com/screwdriver-cd/screwdriver/issues/616
Relevant API PRs: https://github.com/screwdriver-cd/screwdriver/pull/647, https://github.com/screwdriver-cd/screwdriver/pull/646

### DIFF
--- a/README.md
+++ b/README.md
@@ -789,6 +789,60 @@ factory.get(config).then(model => {
 | config.version | String | Yes | Version of the template |
 | config.label | String | No | Label of the template |
 
+### Template Tag Model
+```js
+'use strict';
+const Model = require('screwdriver-models');
+const factory = Model.TemplateTagFactory.getInstance({
+    datastore
+});
+const config = {
+    name: 'testTemplate',
+    tag: 'stable',
+    version: '1.3'
+}
+
+factory.create(config)
+    .then(model => {    // do something
+    });
+```
+
+#### Update
+Update a specific template tag
+```js
+// update template version value
+model.version = '2.4';
+
+model.update()
+```
+
+### Template Tag Factory
+#### Create
+```js
+factory.create(config).then(model => {
+    // do stuff with template tag model
+});
+```
+
+| Parameter        | Type  |  Required | Description |
+| :-------------   | :---- | :-------------|  :-------------|
+| config        | Object | Yes | Configuration Object |
+| config.name | String | Yes | The template name |
+| config.tag | String | Yes | The template tag (e.g. stable, latest, etc) |
+| config.version | String | Yes | Version of the template |
+
+#### Get
+Get a template tag based on id.
+```js
+factory.get(id).then(model => {
+    // do stuff with template model
+});
+```
+
+| Parameter        | Type  |  Description |
+| :-------------   | :---- | :-------------|
+| id | String | The unique ID for the Template Tag |
+
 ### Token Factory
 #### Search
 ```js
--- a/lib/templateTag.js
+++ b/lib/templateTag.js
@@ -1,0 +1,21 @@
+'use strict';
+
+const BaseModel = require('./base');
+
+class TemplateTagModel extends BaseModel {
+    /**
+     * Construct a TemplateTagModel object
+     * @method constructor
+     * @param  {Object}    config
+     * @param  {Object}    config.datastore     Object that will perform operations on the datastore
+     * @param  {String}    config.name          The template name
+     * @param  {String}    config.tag           The template tag (e.g.: 'stable' or 'latest')
+     * @param  {String}    config.version       Version of the template
+     */
+    constructor(config) {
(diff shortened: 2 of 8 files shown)


lib/templateTag.js
```diff
@@ -0,0 +1,18 @@
...
+const BaseModel = require('./base');
+
+class TemplateTagModel extends BaseModel {
+    /**
+     * Construct a TemplateTagModel object
+     * @method constructor
+     * @param  {Object}    config
+     * @param  {Object}    config.datastore         Object that will perform operations on the datastore
+     */
+    constructor(config) {
```
minzcmu: Should have `name`, `tag`, `version` here in the param

lib/templateTagFactory.js
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Instantiate a TemplateTag class
+     * @method createClass
+     * @param  {Object}     config               Template tag data
+     * @param  {Datastore}  config.datastore     Object that will perform operations on the datastore
+     * @param  {String}     config.name          The template name
+     * @param  {String}     config.tag           The template tag (e.g.: 'stable' or 'latest')
+     * @param  {String}     config.version       Version of the template
+     * @return {Job}
```
minzcmu: {TemplateTag}

minzcmu (changes requested):

lib/templateTagFactory.js
tkyi: Whoops. fixed!

README.md
```diff
@@ -789,6 +789,57 @@ factory.get(config).then(model => {
...
+
+factory.create(config)
+    .then(model => {    // do something
+    });
+```
+
+#### Update
+Update a specific template tag
+```js
+model.update()
```
FenrirUnbound: Showing that you're changing a model's value then calling update would be more valuable

FenrirUnbound:
I guess I'm not understanding the implementation in full.

So there is a separate `TemplateTag` model. This sounds like we're making a new table between templates and tag names, but I'm not sure that's actually the intention. Why did we choose to make an independent model for template tags?

FenrirUnbound:
The other thing I've noticed is that your example says we should instantiate the Factory independently of a template. I cannot think of a situation where we would want the `TagTemplate` independent of the `Template` it's associated with. What scenarios do we expect where we would want just the `TagTemplate` and not also want the `Template` model?

FenrirUnbound:
Lastly, the label for this PR says "minor", but the PR header and the commit both use "fix".

tkyi:
@FenrirUnbound Yes you should read the implementation details in the issue I tagged above: https://github.com/screwdriver-cd/screwdriver/issues/616#issuecomment-314568245
We are making another table that just maps between template names, versions, and tags.

We would use a `TagTemplate` without the `Template` model when deleting a tag. I'd take a look at the PRs open for the API to better understand the implementation details. (also linked above)

I can change the commit message to `feat`.

lib/templateTagFactory.js
```diff
@@ -0,0 +1,49 @@
...
+const TemplateTag = require('./templateTag');
+let instance;
+
+class TemplateTagFactory extends BaseFactory {
+    /**
+     * Construct a TemplateTagFactory object
+     * @method constructor
+     * @param  {Object}     config
+     * @param  {Datastore}  config.datastore     Object that will perform operations on the datastore
+     * @param  {String}     config.name          The template name
```
FenrirUnbound: Why does the factory need all of these variables?

FenrirUnbound: I would expect these more in the `.create()`

FenrirUnbound:
> We would use a TagTemplate without the Template model when deleting a tag. I'd take a look at the PRs open for the API to better understand the implementation details. (also linked above)

The DELETE route also requires a lookup for the `Template` 🤔: https://github.com/screwdriver-cd/screwdriver/pull/647/files#diff-9b71f6afabf90daef582f5ba13502794R43.

The point I was getting to was to prefer composition. If you're always going to have a `TemplateTag` associated with a `Template`, why not have the `TemplateTag` do the lookup in the same operation? Doing the lookups separately means we're paying two roundtrips to the datastore.

lib/templateTagFactory.js
tkyi: Put it in the wrong place earlier when addressing Min's comment 😆

FenrirUnbound (approved):
👍 

zxcv

minzcmu (approved):

PR #181, opened 2017-07-13
fix(tokens): Convert buffer directly to base64url

ian-fox:
## Context
I realized in my testing that I was getting wildly different lengths of strings as tokens. It turns out that calling `toString` on the buffer of random bytes was sometimes giving me strings of length between 29 and 32. Fortunately, `base64url` can be called directly from buffers (which I missed on the first pass), and it does not seem to have this problem.

## Objective
Have all tokens (and token hashes) be the same length so they can be reliably filtered from logs

## References
Related to https://github.com/screwdriver-cd/screwdriver/pull/642


joelseq (approved):
👍

FenrirUnbound (approved):
👍 

zxcv

PR #182, opened 2017-07-21
feat(616): Configure getTemplate to take in a full template name with version or tag [1]

tkyi:
## Context
Moving logic to `models`. Since the same logic needs to be used by `config-parser`. (https://github.com/screwdriver-cd/config-parser/pull/41)

## Objective
This PR determines if a full template name contains a version or tag. If there's a tag, it'll get the version by getting a template tag. The `getTemplate` will return a template with the corresponding version or the latest published template if no version is provided. If no template tag is found, the method will return undefined.

Also refactored version regex matching in the same file.

## Related Links
Issue: https://github.com/screwdriver-cd/screwdriver/issues/616#issuecomment-314568245
Related PRs: https://github.com/screwdriver-cd/data-schema/pull/166, https://github.com/screwdriver-cd/screwdriver/pull/657, https://github.com/screwdriver-cd/config-parser/pull/41

### DIFF
--- a/README.md
+++ b/README.md
@@ -775,19 +775,16 @@ factory.get(id).then(model => {
 | id | String | The unique ID for the Template |
 
 #### Get Template
-Get the latest template by name, version with/without label. The version can be in any valid version format. It can can be only major, or major.minor, or major.minor.patch. If label is specified, then the latest version with that label will be resolved. If no match found, the function will resolve undefined.
+Get the latest template by name or get a specific template using name and version or name and tag. The version can be in any valid version format: either major, major.minor, or major.minor.patch. If no version is specified, the function will resolve with the latest version published. If no match is found, the function will resolve null.
 ```js
-factory.get(config).then(model => {
+factory.getTemplate(fullTemplateName).then(model => {
     // do stuff with template model
 });
 ```
 
-| Parameter        | Type  |  Required | Description |
-| :-------------   | :---- | :-------------|  :-------------|
-| config        | Object | Yes | Configuration Object |
-| config.name | String | Yes | The template name |
-| config.version | String | Yes | Version of the template |
-| config.label | String | No | Label of the template |
+| Parameter        | Type   |  Required | Description |
+| :-------------   | :----- | :-------------|  :-------------|
+| fullTemplateName | String | Yes | Name of the template and the version or tag (e.g. chef/publish@1.2.3 or chef/publish@latest). Can also be just name of the template (e.g. chef/publish) |
 
 ### Template Tag Model
 ```js
--- a/lib/templateFactory.js
+++ b/lib/templateFactory.js
@@ -6,10 +6,9 @@ const compareVersions = require('compare-versions');
 const schema = require('screwdriver-data-schema');
 let instance;
 
+const TEMPLATE_NAME_REGEX = schema.config.regex.FULL_TEMPLATE_NAME;
+const EXACT_VERSION_REGEX = schema.config.regex.EXACT_VERSION;
 const VERSION_REGEX = schema.config.regex.VERSION;
-const MAJOR_MATCH = 1;
-const MINOR_MATCH = 2;
-const PATCH_MATCH = 3;
 
 class TemplateFactory extends BaseFactory {
     /**
@@ -55,22 +54,17 @@ class TemplateFactory extends BaseFactory {
      * @return {Promise}
      */
     create(config) {
-        const match = VERSION_REGEX.exec(config.version);
-        const major = match[MAJOR_MATCH];
-        const minor = match[MINOR_MATCH];
+        const [, major, minor] = VERSION_REGEX.exec(config.version);
         const searchVersion = minor ? `${major}${minor}` : major;
 
-        return this.getTemplate({
-            name: config.name,
-            version: searchVersion
-        }).then((latest) => {
+        return this.getTemplate(`${config.name}@${searchVersion}`)
+        .then((latest) => {
             if (!latest) {
                 config.version = minor ? `${major}${minor}.0` : `${major}.0.0`;
             } else {
-                const latestMatch = VERSION_REGEX.exec(latest.version);
-                const latestMajor = latestMatch[MAJOR_MATCH];
-                const latestMinor = latestMatch[MINOR_MATCH];
-                const patch = parseInt(latestMatch[PATCH_MATCH].slice(1), 10) + 1;
+                // eslint-disable-next-line max-len
+                const [, latestMajor, latestMinor, latestPatch] = VERSION_REGEX.exec(latest.version);
+                const patch = parseInt(latestPatch.slice(1), 10) + 1;
 
                 config.version = `${latestMajor}${latestMinor}.${patch}`;
             }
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -775,19 +775,16 @@ factory.get(id).then(model => {
 | id | String | The unique ID for the Template |
 
 #### Get Template
-Get the latest template by name, version with/without label. The version can be in any valid version format. It can can be only major, or major.minor, or major.minor.patch. If label is specified, then the latest version with that label will be resolved. If no match found, the function will resolve undefined.
+Get the latest template by name or get a specific template using name and version or name and tag. The version can be in any valid version format: either major, major.minor, or major.minor.patch. If no version is specified, the latest version published will be resolve. If no match is found, the function will resolve undefined.
```
ian-fox: Will be resolve?

lib/templateFactory.js
```diff
@@ -6,6 +6,9 @@ const compareVersions = require('compare-versions');
 const schema = require('screwdriver-data-schema');
 let instance;
 
+const TEMPLATE_NAME_REGEX = schema.config.regex.FULL_TEMPLATE_NAME;
+const TEMPLATE_NAME_MATCH = 1;
```
ian-fox: Should these values be here, or in data schema? They're tied to the regex right?

lib/templateFactory.js
```diff
@@ -80,33 +81,72 @@ class TemplateFactory extends BaseFactory {
...
-                }
+    getTemplate(fullTemplateName) {
+        const match = TEMPLATE_NAME_REGEX.exec(fullTemplateName);
+        const templateName = match[TEMPLATE_NAME_MATCH];
+        const versionOrTag = match[VERSION_OR_TAG_MATCH];
+        const isVersion = VERSION_REGEX.exec(versionOrTag);
+
+        return new Promise((resolve) => {
+            // If no version or tag provided, return version is null
+            if (!versionOrTag) {
```
d2lam: Don't think we need this. If we don't have this & versionOrTag is null, it will go to line 122 and `resolve(null)` anyway

lib/templateFactory.js
```diff
@@ -80,33 +81,72 @@ class TemplateFactory extends BaseFactory {
...
+                    return resolve(templateTag.version);
+                });
+            }
+
+            return resolve(versionOrTag);
+        })
+        .then((version) => {
+            // If no template tag exists, return undefined
+            if (version === -1) {
+                return undefined;
```
d2lam: Should be `null` for consistency with other factories.

d2lam (changes requested):

lib/templateFactory.js
tkyi: We still need it because if we don't have a `versionOrTag`, `!isVersion` will evaluate to true and the code will think the `versionOrTag` is a tag.

lib/templateFactory.js
tkyi: I was looking at the original README and it said `undefined`, so I was trying to be consistent. I can change it to `null` if that's better.

lib/templateFactory.js
tkyi: I think it makes sense to have them here. I can move it to the data-schema if that makes more sense. We have version regex matching values in this file as well.

ian-fox: I feel like it might make more sense to have them in the schema since that's where the regexes themselves are defined. In this file they seem like magic numbers.

To avoid having to import five separate constants for all the different groups they could possibly also be wrapped in an object called `FULL_TEMPLATE_NAME_GROUPS` or something like that.

lib/templateFactory.js
d2lam: Oh I see. You can keep it this way or you can do `if (versionOrTag && !isVersion)` on line 102 then you don't need these extra lines to check. Up to you.

lib/templateFactory.js
d2lam: Hmmm can you double check what other factories return? I remember they return `null`. If that's the case then we should keep it consistent and update the doc.

lib/templateFactory.js
FenrirUnbound: It can also be written:

```
[ fullText, templateName, versionOrTag ] = VERSION_REGEX.exec(config.version);
```

It removes the need for a magic number and it still fulfills the need to pull out the necessary RegEx matches.

tkyi: @FenrirUnbound Writing it that way would mean I'd have a variable defined that's never used (`fullText`) and a couple extra lines to declare the `templateName` and `versionOrTag` variables.

@ian-fox I think it's a good idea to move the magic numbers to the schema. I'll refactor this code after the template tags PRs are merged.

lib/templateFactory.js
tkyi: Switched to return `null`. The datastore returns `null` when nothing is found. The documentation was inconsistent.

tkyi:
**Update:** Moved the logic around based on comments from @minz1027. Hopefully it's easier to read now.

lib/templateFactory.js
FenrirUnbound: > Writing it that way would mean I'd have a variable defined that's never used (`fullText`) and a couple extra lines to declare the `templateName` and `versionOrTag` variables.

If you don't need one of the arguments, you can ignore it.

```
const [, templateName, versionOrTag ] = VERSION_REGEX.exec(config.version);
```

FenrirUnbound: It's called [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment)

lib/templateFactory.js
```diff
@@ -80,32 +81,60 @@ class TemplateFactory extends BaseFactory {
...
+        const match = TEMPLATE_NAME_REGEX.exec(fullTemplateName);
+        const templateName = match[TEMPLATE_NAME_MATCH];
+        const versionOrTag = match[VERSION_OR_TAG_MATCH];
+        const isVersion = VERSION_REGEX.exec(versionOrTag);
+
+        // If tag is passed in, get the version from the tag
+        if (versionOrTag && !isVersion) {
+            // Lazy load factory dependency to prevent circular dependency issues
+            // eslint-disable-next-line global-require
+            const TemplateTagFactory = require('./templateTagFactory');
```
FenrirUnbound, line 99: Why can't we do this in the constructor?

FenrirUnbound: Doing this in the constructor would allow us to do this operation exactly once.

lib/templateFactory.js
```diff
@@ -80,32 +81,60 @@ class TemplateFactory extends BaseFactory {
...
-                // get templates that have version prefix as config.version
+                // If no version provided, return the most recently published template
+                if (!versionOrTag) {
+                    return templates[0];
+                }
+
+                // Get templates that have versions beginning with the version given
                 const filtered = templates.filter(template =>
-                    template.version.startsWith(config.version));
+                    template.version.startsWith(versionOrTag));
```
FenrirUnbound, line 131: So if I ask to use version `1.0.1`, it will give me `1.0.100`?

minzcmu: Should have a check before the list. If it's a exact version then just use a `get`

lib/templateFactory.js
tkyi: Cool! Updated

lib/templateFactory.js
tkyi: Hm but how do I know if it's an exact version or not?
`1.0.1` and `1.0.100` are both exact versions

tkyi: Is there ever a case where a user might want `1.0.1xx` but don't know the exact version and this might make sense?

tkyi: ^ Probably not

Added a check

lib/templateFactory.js
ian-fox: I don't think putting lazy-loads in the constructor actually solves the circular dependency problem. If we have modules `A` and `B` that depend on each other we'd end up with constructor `A` calling require for `B`, which calls constructor `B`, which calls require for `A`, and so on.

lib/templateFactory.js
```diff
@@ -80,32 +74,67 @@ class TemplateFactory extends BaseFactory {
...
 
-                // If no config.label, return latest one
-                if (filtered.length > 0 && !config.label) {
-                    return filtered[0];
-                }
-
-                // Loop through filtered to find the first match for config.label
-                return filtered.find(template => template.labels.includes(config.label));
+                // Return latest one
+                return filtered[0];
```
ian-fox: Are we sure there's at least one element in this array?

tkyi: Addressed

lib/templateFactory.js
```diff
@@ -80,32 +74,67 @@ class TemplateFactory extends BaseFactory {
...
 
-                // If no config.label, return latest one
-                if (filtered.length > 0 && !config.label) {
-                    return filtered[0];
-                }
-
-                // Loop through filtered to find the first match for config.label
-                return filtered.find(template => template.labels.includes(config.label));
+                // Return first filtered template or null if none
+                return filtered[0] ? filtered[0] : null;
```
ian-fox: I think `filtered[0] || null` is a bit cleaner. Other than that everything looks good!

tkyi: Fixed

ian-fox (approved):
:+1:

minzcmu (approved):
👍 LGTM

PR #184, opened 2017-07-27
test(collections): Improve code coverage of collectionFactory

joelseq:
## Context

Currently a branch is not being tested (when a user does not specify pipelineIds in their request).

## Objective

This PR adds a test for that branch and makes sure that the collectionFactory adds in the pipelineIds field with a value of empty array.

## References

Issue: [screwdriver-cd/screwdriver#523](https://github.com/screwdriver-cd/screwdriver/issues/523)

### DIFF
--- a/test/lib/collectionFactory.test.js
+++ b/test/lib/collectionFactory.test.js
@@ -90,6 +90,32 @@ describe('Collection Factory', () => {
                 });
             });
         });
+
+        it('should create a Collection without pipelineIds', () => {
+            const dataWithoutPipelineIds = Object.assign({}, collectionData);
+
+            dataWithoutPipelineIds.pipelineIds = [];
+            datastore.save.resolves(dataWithoutPipelineIds);
+
+            return factory.create({
+                userId,
+                name,
+                description
+            }).then((model) => {
+                assert.isTrue(datastore.save.calledOnce);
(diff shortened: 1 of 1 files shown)


test/lib/collectionFactory.test.js
```diff
@@ -90,6 +90,34 @@ describe('Collection Factory', () => {
                 });
             });
         });
+
+        it('should create a Collection without pipelineIds', () => {
+            const dataWithoutPipelineIds = Object.assign({}, collectionData);
+
+            delete dataWithoutPipelineIds.pipelineIds;
+            datastore.save.resolves(dataWithoutPipelineIds);
```
minzcmu: This should resolve pipelineIds as empty array right? That's what you put as save param

minzcmu (changes requested):

test/lib/collectionFactory.test.js
joelseq: Whoops! You're right, I messed up on that. But the main part of the test is actually [this line](https://github.com/screwdriver-cd/models/pull/184/files#diff-1804da3236cbfdf1b4a8055e43ebb888R111). I'll change the test right now

minzcmu (approved):

tkyi (approved):

PR #185, opened 2017-08-14
fix(tokens): use pbkdf2 to securely hash tokens

ian-fox:
## Context

After taking a security workshop, I learned that sha256 is not a secure way to store passwords. PBKDF2 is much better, as it takes longer (making it harder to crack).

This new algorithm will invalidate all previously generated tokens, of which there are about 5 on cd.screwdriver.cd (all of which belong to us).

100,000 iterations of PBKDF2 are used as the performance is still pretty reasonable (84ms to run the unit test), but it should be secure (Apple, for example, uses 10,000 for itunes passwords according to OWASP).

## Goal

Switch tokens to use PBKDF2 instead of sha256.

## References

[OWASP Password Storage Cheat Sheet](https://www.owasp.org/index.php/Password_Storage_Cheat_Sheet)

### DIFF
--- a/lib/generateToken.js
+++ b/lib/generateToken.js
@@ -1,10 +1,18 @@
 'use strict';
 
-const nodeify = require('./nodeify');
-const crypto = require('crypto');
 const base64url = require('base64url');
+const crypto = require('crypto');
+const nodeify = require('./nodeify');
+
+// Config for pbkdf2
+// 100,000 iterations is still fast enough to be performant (84ms to run the unit test) while also
+// creating enough of a delay to stop brute force attacks
+// For more info, see https://www.owasp.org/index.php/Password_Storage_Cheat_Sheet
 const ALGORITHM = 'sha256';
-const TOKEN_LENGTH = 256; // Measured in bits
+const ITERATIONS = 100000;
+
+// Token length, measured in bytes
+const TOKEN_LENGTH = 32;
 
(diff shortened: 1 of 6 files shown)


joelseq (approved):
👍

d2lam (approved):

lib/generateToken.js
```diff
@@ -1,28 +1,35 @@
 'use strict';
 
-const nodeify = require('./nodeify');
-const crypto = require('crypto');
 const base64url = require('base64url');
+const crypto = require('crypto');
+const nodeify = require('./nodeify');
```
FenrirUnbound, line 5: 👍

lib/generateToken.js
```diff
@@ -1,28 +1,35 @@
...
+
+// Config for pbkdf2
 const ALGORITHM = 'sha256';
-const TOKEN_LENGTH = 256; // Measured in bits
+const ITERATIONS = 100000;
```
FenrirUnbound: Not a blocker. It may be note-worthy to state why this constant was chosen. I understand it's recorded in the PR, so that may be sufficient enough.

FenrirUnbound (approved):
👍 

Looks fine to me. Did not merge it due to an outstanding review

giphy 81

r3rastogi (approved):

PR #191, opened 2017-08-24
Secure tokens

ian-fox:
* undoes https://github.com/screwdriver-cd/models/pull/189, adding the token changes back to models
* fix all uses of `generateToken.hashValue` to be async


ian-fox:
Should this be a breaking change? Previous tokens won't work with it.

tkyi (approved):
Should probably bump major

joelseq (approved):
👍

PR #192, opened 2017-08-24
fix(tokens): token hashes are 512 bits

ian-fox:
It was recommended to me that token hashes should be 512 bits.


tkyi (approved):
Seems reasonable

joelseq (approved):
👍

r3rastogi (approved):

PR #194, opened 2017-08-26
feat(collections): Add type to config for pipeline getJobs

joelseq:
## Context

For the user dashboards we need to get info about the number of PRs open and failing for each pipeline in a collection. It would be nice to be able to get only the PR jobs for a pipeline with a single call to the model.

## Objective

This PR adds a param to the config of getJobs for the pipeline model to specify an optional type for the jobs wanted. The user can specify either 'pipeline' or 'pr' which would return only the workflow jobs or PR jobs respectively. Not specifying a type will return all the jobs.

## References

Issue: [screwdriver-cd/screwdriver#523](https://github.com/screwdriver-cd/screwdriver/issues/523)

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -423,6 +423,7 @@ class PipelineModel extends BaseModel {
      * @param  {Object}   [config]                  Configuration object
      * @param  {Object}   [config.params]           Filter params
      * @param  {Boolean}  [config.params.archived]  Get archived/non-archived jobs
+     * @param  {String}   [config.type]             Type of jobs (pr or pipeline)
      * @return {Promise}  Resolves to an array of jobs
      */
     getJobs(config) {
@@ -438,6 +439,10 @@ class PipelineModel extends BaseModel {
         };
 
         const listConfig = (config) ? hoek.applyToDefaults(defaultConfig, config) : defaultConfig;
+
+        if (listConfig.type) {
+            delete listConfig.type;
+        }
 
         // Lazy load factory dependency to prevent circular dependency issues
         // https://nodejs.org/api/modules.html#modules_cycles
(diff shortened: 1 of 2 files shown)


lib/pipeline.js
```diff
@@ -469,6 +474,13 @@ class PipelineModel extends BaseModel {
                 // sort them by pr number
                 prJobs = prJobs.sort((job1, job2) => job1.prNum - job2.prNum);
 
+                if (config && config.type === 'pr') {
```
r3rastogi, line 477: I'd prefer if we use "pullrequest" instead of "pr"

joelseq: I was trying to keep it consistent with https://github.com/screwdriver-cd/models/blob/master/lib/eventFactory.js#L33 but I can change it if this is still better, what do you think?

tkyi: I'm okay with `pr`

tkyi (approved):
👍

lib/pipeline.js
r3rastogi: Ahh ok. I guess if it's already defined like that elsewhere, it's fine to keep it consistent here.

test/lib/pipeline.test.js
```diff
@@ -639,6 +639,52 @@ describe('Pipeline Model', () => {
             });
         });
 
+        it('Only gets PR jobs', () => {
```
r3rastogi, line 642: Should we also add a test where if we don't specify a config type, it returns all jobs?

joelseq: I believe that test case is already handled previously in the [older tests](https://github.com/screwdriver-cd/models/blob/master/test/lib/pipeline.test.js#L601-L640) above this one where no config is given.

r3rastogi (approved):

test/lib/pipeline.test.js
r3rastogi: mkay 👍

PR #195, opened 2017-10-06
feat(723): eventFactory also creates builds

d2lam:
eventFactory now takes in a new field `startFrom`, it will create corresponding builds as well. 

For example:
startFrom = `main` then the `main` job will be started
startFrom = `~pr` then all the jobs that have `requires: ['~pr']` will be started

With this change, we will also get partial events. 

Related: https://github.com/screwdriver-cd/screwdriver/issues/723
Blocked by: https://github.com/screwdriver-cd/config-parser/pull/48

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ factory.get({ scmUri }).then(model => {
 
 | Parameter        | Type  |  Description |
 | :-------------   | :---- | :-------------|
-| id | String | The unique ID for the pipeline |
+| id | Number | The unique ID for the pipeline |
 | config.scmUri | String | Source Code URI for the application |
 
 
@@ -159,7 +159,7 @@ const factory = Model.JobFactory.getInstance({
 });
 const config = {
     params: {
-        pipelineId: 'aabbccdd1234'
+        pipelineId: 1
     },
     paginate {
         page: 2,
@@ -189,7 +189,7 @@ factory.create(config).then(model => {
 | Parameter        | Type  |  Description |
 | :-------------   | :---- | :-------------|
 | config        | Object | Configuration Object |
-| config.pipelineId | String | The pipelineId that the job belongs to |
+| config.pipelineId | Number | The pipelineId that the job belongs to |
 | config.name | String | The name of the job |
 
 #### Get
@@ -206,8 +206,8 @@ factory.get({ pipelineId, name }).then(model => {
 
 | Parameter        | Type  |  Description |
 | :-------------   | :---- | :-------------|
-| id | String | The unique ID for the job |
-| config.pipelineId | String | Id of the pipeline the job is associated with |
+| id | Number | The unique ID for the job |
+| config.pipelineId | Number | Id of the pipeline the job is associated with |
 | config.name | String Name of the job |
 
 ### Job Model
@@ -260,7 +260,7 @@ const factory = Model.BuildFactory.getInstance({
 });
 const config = {
     params: {
-        jobId: 'aaabbccdd1234'
+        jobId: 4
     },
     paginate {
         page: 2,
@@ -297,7 +297,7 @@ factory.create(config).then(model => {
 | config.container | String | No | Container for the build to run in |
 | config.sha | String | No | SHA used to kick off the build |
 | config.prRef | String | No | PR branch or reference; required for PR jobs |
-| config.eventId | String | No | Id of the event this build belongs to |
+| config.eventId | Number | No | Id of the event this build belongs to |
 
 #### Get
 Get a build based on id. Can pass the generatedId for the build, or the unique keys for the model, and the id will be determined automatically.
@@ -313,8 +313,8 @@ factory.get({ jobId, number }).then(model => {
 
 | Parameter        | Type  |  Description |
 | :-------------   | :---- | :-------------|
-| id | String | The unique ID for the build |
-| config.jobId | String | The unique ID for a job |
+| id | Number | The unique ID for the build |
+| config.jobId | Number | The unique ID for a job |
 | config.number | Number | build number |
 
 ### Build Model
@@ -438,7 +438,7 @@ factory.get({ token }).then(model => {
 
 | Parameter        | Type  |  Description |
 | :-------------   | :---- | :-------------|
-| id | String | The unique ID for the build |
+| id | Number | The unique ID for the build |
 | config.username | String | User name |
 | config.scmContext | String | Scm context to which user belongs |
 | config.accessToken | String | A user access token value |
@@ -520,7 +520,7 @@ const factory = Model.SecretFactory.getInstance({
 });
 const config = {
     params: {
-        pipelineId: '2d991790bab1ac8576097ca87f170df73410b55c'
+        pipelineId: 1
     },
     paginate {
         page: 2,
@@ -550,7 +550,7 @@ factory.create(config).then(model => {
 | Parameter        | Type  |  Required | Description |
 | :-------------   | :---- | :-------------|  :-------------|
 | config        | Object | Yes | Configuration Object |
-| config.pipelineId | String | Yes | Pipeline that this secret belongs to |
+| config.pipelineId | Number | Yes | Pipeline that this secret belongs to |
 | config.name | String | Yes | Secret name |
 | config.value | String | Yes | Secret value |
 | config.allowInPR | String | Yes | Flag to denote if this secret can be shown in PR builds |
@@ -569,8 +569,8 @@ factory.get({ pipelineId, name }).then(model => {
 
 | Parameter        | Type  |  Description |
 | :-------------   | :---- | :-------------|
-| id | String | The unique ID for the build |
-| config.pipelineId | String | Pipeline that the secret belongs to |
+| id | Number | The unique ID for the build |
+| config.pipelineId | Number | Pipeline that the secret belongs to |
 | config.name | String | Secret name |
 
 
@@ -583,7 +583,7 @@ const factory = Model.SecretFactory.getInstance({
     password            // Password for encryption operations
 });
 const config = {
-    pipelineId: '2d991790bab1ac8576097ca87f170df73410b55c',
+    pipelineId: 1,
     name: 'NPM_TOKEN',
     value: banana,
     allowInPR: false
@@ -611,7 +611,7 @@ const factory = Model.EventFactory.getInstance({
 });
 const config = {
     params: {
-        pipelineId: '2d991790bab1ac8576097ca87f170df73410b55c'
+        pipelineId: 1
     }
 }
 
@@ -636,9 +636,10 @@ factory.create(config).then(model => {
 | :-------------   | :---- | :-------------|  :-------------|
 | config        | Object | Yes | Configuration Object |
 | config.type | String | No | Event type: pipeline or pr |
-| config.pipelineId | String | Yes | Unique identifier of pipeline |
+| config.pipelineId | Number | Yes | Unique identifier of pipeline |
 | config.sha | String | Yes | Commit sha that the event was based on |
-| config.workflow | Array | Yes | Workflow of the pipeline |
+| config.workflow | Array | No | Workflow of the pipeline |
+| config.workflowGraph | Object | No | Workflow graph of the pipeline, with edges and nodes |
 | config.username | String | Yes | Username of the user that creates this event |
 | config.causeMessage | String | No | Message that describes why the event was created |
 
@@ -656,8 +657,8 @@ factory.get({ pipelineId, sha }).then(model => {
 
 | Parameter        | Type  |  Description |
 | :-------------   | :---- | :-------------|
-| id | String | The unique ID for the build |
-| config.pipelineId | String | Unique identifier of pipeline |
+| id | Number | The unique ID for the build |
+| config.pipelineId | Number | Unique identifier of pipeline |
 | config.sha | String | Commit sha that the event was based on |
 
 ### Event Model
@@ -669,9 +670,21 @@ const factory = Model.EventFactory.getInstance({
     scm
 });
 const config = {
-    pipelineId: '12345f642bbfd1886623964b4cff12db59869e5d',
+    pipelineId: 1,
     sha: 'ccc49349d3cffbd12ea9e3d41521480b4aa5de5f',
-    workflow: ['main', 'publish'],    
+    workflowGraph: {
+        nodes: [
+            { name: '~pr' },
+            { name: '~commit' },
+            { name: 'main' },
+            { name: 'publish' }
+        ],
+        edges: [
+            { src: '~pr', dest: 'main' },
+            { src: '~commit', dest: 'main' },
+            { src: 'main', dest: 'publish' }
+        ]
+    },    
     username: 'stjohn',
     causeMessage: 'Merge pull request #26 from screwdriver-cd/models'
 }
@@ -685,7 +698,7 @@ Example event model that got created:
 ```json
 {
     "type": "pipeline",
-    "pipelineId": "12345f642bbfd1886623964b4cff12db59869e5d",
+    "pipelineId": "1",
     "sha": "ccc49349d3cffbd12ea9e3d41521480b4aa5de5f",
     "createTime": "2038-01-19T03:14:08.131Z",
     "commit": {
@@ -698,7 +711,19 @@ Example event model that got created:
             "username": "imbatman"
         }
     },
-    "workflow": ["main", "publish"],
+    "workflowGraph": {
+        "nodes": [
+            { "name": "~pr" },
+            { "name": "~commit" },
+            { "name": "main" },
+            { "name": "publish" }
+        ],
+        "edges": [
+            { "src": "~pr", "dest": "main" },
+            { "src": "~commit", "dest": "main" },
+            { "src": "main", "dest": "publish" }
+        ]
+    },
     "causeMessage": "Merge pull request #26 from screwdriver-cd/models",
     "creator": {
         "avatar": "https://avatars.githubusercontent.com/u/2042?v=3",
@@ -765,7 +790,7 @@ factory.create(config).then(model => {
 | config.description | String | Yes | Description of the template |
 | config.maintainer | Array | Yes | Maintainer's email |
 | config.config | Object | Yes | Config of the screwdriver-template.yaml |
-| config.pipelineId | String | Yes | pipelineId of the template |
+| config.pipelineId | Number | Yes | pipelineId of the template |
 | config.labels | Array | No | Labels attached to the template |
 
 #### Get
@@ -778,7 +803,7 @@ factory.get(id).then(model => {
 
 | Parameter        | Type  |  Description |
 | :-------------   | :---- | :-------------|
-| id | String | The unique ID for the Template |
+| id | Number | The unique ID for the Template |
 
 #### Get Template
 Get the latest template by name or get a specific template using name and version or name and tag. The version can be in any valid version format: either major, major.minor, or major.minor.patch. If no version is specified, the function will resolve with the latest version published. If no match is found, the function will resolve null.
@@ -844,7 +869,7 @@ factory.get(id).then(model => {
 
 | Parameter        | Type  |  Description |
 | :-------------   | :---- | :-------------|
-| id | String | The unique ID for the Template Tag |
+| id | Number | The unique ID for the Template Tag |
 
 ### Token Factory
 #### Search
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -14,11 +14,11 @@ let instance;
  * @param  {String}     jobName     Job name
  * @return {Job}                    Job object
  */
-function createJob(jobFactory, job, jobName) {
+function createJob(jobFactory, { permutations, jobName, pipelineId }) {
     return jobFactory.create({
-        pipelineId: job.pipelineId,
+        pipelineId,
         name: jobName,
-        permutations: job.permutations
+        permutations
     });
 }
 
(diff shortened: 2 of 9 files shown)


test/lib/eventFactory.test.js
```diff
@@ -167,6 +205,43 @@ describe('Event Factory', () => {
             })
         );
 
+        it('should create builds if using new workflow', () => {
+            pipelineMock.jobs = Promise.resolve(jobsMock);
+            config.startFrom = '~pr';
```
petey: For completeness, should also have a test for `~commit`.

test/lib/eventFactory.test.js
```diff
@@ -167,6 +205,43 @@ describe('Event Factory', () => {
...
+                    eventId: model.id,
+                    jobId: 3,
+                    prRef: 'branch'
+                }));
+            });
+        });
+
+        it('should create build if startFrom is a jobName', () => {
+            pipelineMock.jobs = Promise.resolve(jobsMock);
+            config.startFrom = 'component';
```
petey: Let's try this test with`integration` job instead of `component`. Incidentally, what will happen if `startFrom` is a disabled job? It doesn't look like you'd get an error (but you should), and we should have a test for this case.

petey (changes requested):
Need more tests.

test/lib/eventFactory.test.js
petey: Also a test for `startFrom` is a job that doesn't exist.

d2lam: @petey  The expected behavior is reject in both cases?

d2lam:
@petey  I added more test cases!

test/lib/eventFactory.test.js
```diff
@@ -167,6 +205,84 @@ describe('Event Factory', () => {
...
+                    jobId: 3,
+                    prRef: 'branch'
+                }));
+            });
+        });
+
+        it('should create commit builds if using new workflow', () => {
+            pipelineMock.jobs = Promise.resolve(jobsMock);
+            config.startFrom = '~commit';
+            config.prRef = 'branch';
```
petey: copy/paste: prRef not needed when it isn't a PR

lib/eventFactory.js
```diff
@@ -2,9 +2,61 @@
...
+    // If no startFrom, then do nothing. Remove once we switch to new workflow design
+    if (!startFrom) {
+        return null;
+    }
+
+    // eslint-disable-next-line global-require
+    const BuildFactory = require('./buildFactory');
+    const buildFactory = BuildFactory.getInstance();
+
+    return pipeline.jobs.then((jobs) => {
```
petey: I'm kind of concerned that `createBuilds` is looking up the jobs from the pipeline instead of getting the information from `config.workflow`. I feel like this results in a race condition where the pipeline's workflow could be changed between the event creation and event's initial jobs starting.

petey (changes requested):

lib/eventFactory.js
d2lam: Sounds like a legit concern. We can change that.

lib/eventFactory.js
```diff
@@ -2,9 +2,94 @@
...
+ * @param  {String}   [config.startFrom]         Where the event starts from (jobname or ~commit, ~pr, etc)
+ *                                               Making optional for backward compatibility
+ * @param  {Number}   eventId                    Event's id
+ */
+function createBuilds(pipeline, config, eventId) {
+    const startFrom = config.startFrom;
+
+    // If no startFrom, then do nothing. Remove once we switch to new workflow design
+    if (!startFrom) {
+        return null;
```
jithine: Shouldn't it start build for `main` job ?

lib/eventFactory.js
```diff
@@ -2,9 +2,94 @@
...
+    }
+
+    /* eslint-disable global-require */
+    const BuildFactory = require('./buildFactory');
+    const buildFactory = BuildFactory.getInstance();
+    const JobFactory = require('./jobFactory');
+    const jobFactory = JobFactory.getInstance();
+    /* eslint-enable global-require */
+
+    const createJob = (job, jobName) => jobFactory.create({
```
jithine: don't create function inside function.

lib/eventFactory.js
```diff
@@ -2,9 +2,94 @@
...
+function createBuilds(pipeline, config, eventId) {
+    const startFrom = config.startFrom;
+
+    // If no startFrom, then do nothing. Remove once we switch to new workflow design
+    if (!startFrom) {
+        return null;
+    }
+
+    /* eslint-disable global-require */
+    const BuildFactory = require('./buildFactory');
```
jithine, line 157: any reason why these are not global ?

lib/eventFactory.js
```diff
@@ -2,9 +2,94 @@
...
+
+    return pipeline.jobs.then((jobs) => {
+        const jobNamesArray = jobs.map(j => j.name);
+        const jobsToCreate = [];
+        let jobsToStart = [];
+
+        jobs.forEach((job) => {
+            const enabled = job.state === 'ENABLED';
+            const requires = job.permutations.requires;
+            const isJobName = joi.validate(startFrom, jobConfigSchema.jobname);
```
jithine: I would rather have `const isJobName = !!joi.validate(startFrom, jobConfigSchema.jobname).error`

and then `isJobName && startFrom === job.name` This way if conditions in line 59 & 62 are more readable. I expected `isJobName` & `isTrigger` to be booleans based on their naming.

lib/eventFactory.js
```diff
@@ -2,9 +2,94 @@
...
+            const isTrigger = joi.validate(startFrom, jobConfigSchema.trigger);
+
+            if (enabled) {
+                // if startFrom is a jobname, then start that job
+                if (!isJobName.error && startFrom === job.name) {
+                    jobsToStart.push(job);
+                // if startFrom is a trigger (example: ~commit, ~pr) then start jobs that include that key
+                } else if (!isTrigger.error && requires && requires.includes(startFrom)) {
+                    const prJobName = `PR-${config.prNum}-${job.name}`;
+                    const prJobIndex = jobNamesArray.indexOf(prJobName);
```
jithine: what results in this case ? Rerunning PR ?

jithine (changes requested):

lib/eventFactory.js
d2lam: If it's using the old workflow, we don't want to create the build. Since that means the build will be created twice. For example, this is using the old workflow: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/builds/create.js#L75-L94

lib/eventFactory.js
d2lam: This is due to a circular dependency issue
https://github.com/screwdriver-cd/models/blob/master/lib/eventFactory.js#L43-L45

lib/eventFactory.js
d2lam: Yes. Two cases can happen: 
- The first time PR is created, no jobs are created yet, so we need to create the job. 
- Rerunning PR or a new commit to push to the PR, then the PR job already exists

d2lam: Found a bug in the "rerunning case" actually. 
Example. Two cases:
- `PR-1-component` doesn't exist yet, `prJobIndex<0`: it creates the job `PR-1-component` first, then concat the newly created job to `jobToStart`, then start the job (create a build of that job). **This case works fine.** 
- `PR-1-component` already exists: let's say the pipeline has 2 jobs [component, PR-1-component]. When job is `component`, prJobName = `PR-1-component`, since prJobIndex >= 0, it starts the current job, which is `component`, instead of `PR-1-component` (line 68). When job is `PR-1-component`, prJobName = `PR-1-component-PR-1-component`, and it doesn't find that job, so it will create the job. **This is bad!**

lib/eventFactory.js
jithine: Won't this get called for all cases ? Where is the old workflow handling happening ? I thought if it's old workflow then just run `main` else follow `startFrom`

d2lam: We haven't switch everything to the new workflow yet. This is here to ensure backward compatibility. 

If it's using the old design, it will not behave differently from before (aka, just create the event and then done). 

Take a look at the link I included above. It creates the new event with no `startFrom`, so it's using the old design. Hence, it will just create the event, and then it creates the build on its own (https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/builds/create.js#L94)

lib/eventFactory.js
```diff
@@ -2,9 +2,98 @@
...
+                    jobsToStart.push(job);
+                // If startFrom is a trigger (example: ~commit, ~pr), then start jobs that include that key
+                } else if (isTrigger && requires && requires.includes(startFrom)) {
+                    const isPRJob = (jobName).startsWith('PR-');
+                    const prJobName = `PR-${config.prNum}-${jobName}`;
+                    const prJobIndex = jobNamesArray.indexOf(prJobName);
+
+                    // If startFrom is not ~pr, or if ~pr and PR job already exists,
+                    // or if ~pr and isPRJob already, start job
+                    if (startFrom !== '~pr' || prJobIndex >= 0 || isPRJob) {
```
jithine: Referring to  previous explanation (See last scenario https://github.com/screwdriver-cd/models/pull/195#discussion_r143594530)  and then this implementation, I think it will create one extra job if it's re-run.

Let's say we have a job with name `JOB_REQUIRES_PR`. If my understanding is correct then, during first run `createJob` creates a new job with name  `PR-1-JOB_REQUIRES_PR`. Now for what ever reason if this event is processed again, when job with name `JOB_REQUIRES_PR` is processed, it will result in pushing  `PR-1-JOB_REQUIRES_PR` to `jobsToStart` and so does job with name `PR-1-JOB_REQUIRES_PR`, which results in 2 entries of `PR-1-JOB_REQUIRES_PR` in `jobsToStart` Array.

So I propose modifying logic like this.

```
          // If true then it's some case of retry scenario, we should ignore it since whichever job in jobs created it, will process it again.
          const isPrJob = jobName.startsWith('PR-'); 

          if (isEnabled && !prJob) {
                // If startFrom is a jobname, then start that job
                if ((isJobName && startFrom === jobName)) {
                     jobsToStart.push(job);
                // If startFrom is a trigger (example: ~commit, ~pr), then start jobs that include that key
                } else if (isTrigger && requires && requires.includes(startFrom)) {
                    const prJobName = `PR-${config.prNum}-${jobName}`;

                    // If PR job doesn't exist yet then create the job
                    if (startFrom === '~pr' && !jobNamesArray.includes(prJobIndex)) {
                        jobsToCreate.push(createJob(job, prJobName));

                    } else {
                        jobsToStart.push(job);
                    }
}
```

test/lib/eventFactory.test.js
```diff
@@ -142,6 +157,184 @@ describe('Event Factory', () => {
...
+                };
+
+                jobFactoryMock.create.onCall(0).resolves(prComponent);
+                config.startFrom = '~pr';
+                config.prRef = 'branch';
+                config.prNum = '1';
+
+                return factory.create(config).then((model) => {
+                    assert.instanceOf(model, Event);
+                    assert.calledOnce(buildFactoryMock.create);
```
jithine: This test should have failed, if comment above is addressed. Also see potential fix I've given above https://github.com/screwdriver-cd/models/pull/195#discussion_r143636987

test/lib/eventFactory.test.js
```diff
@@ -142,6 +157,184 @@ describe('Event Factory', () => {
...
+                    },
+                    state: 'ENABLED'
+                }];
+
+                pipelineMock.jobs = Promise.resolve(jobsMock);
+                buildFactoryMock.create.resolves(null);
+            });
+
+            it('should start pr builds if using new workflow', () => {
+                jobsMock = [{
```
jithine: to reflect real world it should also have another job config of this format 

```
{
+                    id: 'X' ?,
+                    pipelineId: 8765,
+                    name:  'component',
+                    permutations: {
+                        requires: ['~pr']
+                    },
}
```

jithine (changes requested):

jithine:
Last PR build is stuck https://cd.screwdriver.cd/pipelines/9/builds/11086

lib/eventFactory.js
tkyi: This logic will lead to starting the PR job with the wrong `j.id`. After talking to Jithin, we moved the skipping logic to inside the if statement for `isTrigger`. When the PR job already exists, we'll skip. Otherwise, if the job does not have a `startFrom` that is equal to `~pr` or if it is a PR job, start the job.

tkyi:
Note: We will need to edit the `syncPR` function in future PRs, as it's currently hardcoded to handle PR jobs in the format `PR-${prNum}`, while we're switching to the format `PR-${prNum}-${jobName}`. 
https://github.com/screwdriver-cd/models/blob/8bcb9c8b3d86f03f310ccaca22b76bb7d28826c2/lib/pipeline.js#L227

lib/eventFactory.js
```diff
@@ -2,9 +2,111 @@
...
+    }
+
+    /* eslint-disable global-require */
+    const BuildFactory = require('./buildFactory');
+    const buildFactory = BuildFactory.getInstance();
+    const JobFactory = require('./jobFactory');
+    const jobFactory = JobFactory.getInstance();
+    /* eslint-enable global-require */
+
+    return pipeline.jobs.then((jobs) => {
```
jithine, line 163: How does this behave when there are 2+ PR's open at the same time ? Does `pipeline.jobs` return all ? If so we need to filter out jobs with Pr Number not matching config.prNum.

I think we need logic like ``` if (config.startFrom === '~pr' && job.isPR() && job.prNum !== config.prNum) return ```

jithine (changes requested):

lib/eventFactory.js
tkyi: Made changes to handle this case

lib/eventFactory.js
```diff
@@ -2,9 +2,116 @@
...
+
+                    // If PR job already exists, skip
+                    if (jobNamesArray.includes(prJobName)) {
+                        return;
+                    }
+
+                    // If startFrom is not ~pr, or if ~pr and isPRJob already, start job
+                    if (startFrom !== '~pr' || job.isPR()) {
+                        // Skip if PR number does not match
+                        if (startFrom === '~pr' && job.isPR() && job.prNum() !== config.prNum) {
```
jithine: Can you move this out to top? There is too much nesting right now. I would have something like 
```
jobs
  .filter(filterUnwantedPRJobs ?)
  .forEach()
  ....
)
```

tkyi: I tried. 😢 🆘 
Can you take a look?

jithine (approved):

tkyi:
Rewrote much of the logic in the PR to pull in the new `screwdriver-workflow-parser` module and `workflowGraph` object.

README.md
```diff
@@ -611,7 +611,7 @@ const factory = Model.EventFactory.getInstance({
 });
 const config = {
     params: {
-        pipelineId: '2d991790bab1ac8576097ca87f170df73410b55c'
+        pipelineId: '1'
```
minzcmu: Should be number not string

README.md
```diff
@@ -638,7 +638,8 @@ factory.create(config).then(model => {
 | config.type | String | No | Event type: pipeline or pr |
 | config.pipelineId | String | Yes | Unique identifier of pipeline |
```
minzcmu: Number

lib/eventFactory.js
```diff
@@ -2,9 +2,108 @@
...
+ * Create builds associated with this event
+ * For example, if startFrom ~commit, then create builds with jobs that have requires: ~commit
+ * @method createBuilds
+ * @param  {Pipeline} pipeline                   Pipeline to create builds
+ * @param  {Object}   config
+ * @param  {String}   config.sha                 SHA this project was built on
+ * @param  {String}   config.username            Username of the user that creates this event
+ * @param  {String}   config.scmContext          The scm context to which user belongs
+ * @param  {String}   [config.prRef]             Ref if it's a PR event
+ * @param  {String}   [config.prNum]             PR number if it's a PR event
```
minzcmu: Number

lib/eventFactory.js
```diff
@@ -33,19 +132,22 @@ class EventFactory extends BaseFactory {
      * @param  {String}  [config.type = 'pipeline'] Type of event (pipeline or pr)
      * @param  {String}  config.pipelineId          Unique id of the pipeline
      * @param  {Array}   config.workflow            Job names that will be executed for this event
+     * @param  {Array}   [config.workflowGraph]     Object with nodes and edges that represent the order of jobs
```
minzcmu: Object

minzcmu (changes requested):

lib/eventFactory.js
```diff
@@ -2,9 +2,108 @@
...
+        // Get all the missing PR- job names
+        const existingPRJobNames = PRJobArray.map(p => p.name);
+        const missingPRJobNames = nextJobs.filter(j => !existingPRJobNames.includes(j));
+        // Get the job name part, e.g. main from PR-1:main
+        const missingJobNames = missingPRJobNames.map(name => name.split(':')[1]);
+        const missingJobs = jobs.filter(j => missingJobNames.includes(j.name));
+
+        // Create missing PR jobs
+        return Promise.all(missingJobs.map(j =>
+            createJob(jobFactory, j, `PR-${config.prNum}:${j.name}`)
```
minzcmu: Shouldn't just steal the job config from pipeline cuz inside the PR ppl may change their job config. Should do the same logic currently in the webhook to get he config from prRef

https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/webhooks/index.js#L63

minzcmu (changes requested):

lib/eventFactory.js
```diff
@@ -2,9 +2,131 @@
...
+                    createJob(jobFactory, {
+                        permutations: parsedConfig.jobs[j.name],
+                        jobName: `PR-${config.prNum}:${j.name}`,
+                        pipelineId: pipeline.id
+                    })
+                ));
+            });
+        });
+    })
+    .then((newJobs) => {
```
jithine: can you modify logic to take `jobsToStart` here ? Then you don't need the global variable `jobsToStart`, and no need to do the `concat` business inside this function, which is purely a `PR` job thingy

lib/eventFactory.js
```diff
@@ -2,9 +2,131 @@
...
+        let nextJobs;
+
+        // When startFrom is ~commit
+        if (startFrom === '~commit') {
+            return pipeline.sync()
+            .then((p) => {
+                nextJobs = WorkflowParser.getNextJobs(p.workflowGraph, { trigger: startFrom });
+                jobsToStart = jobs.filter(j => nextJobs.includes(j.name) && j.state === 'ENABLED');
+
+                return Promise.resolve([]);
```
jithine: should make `jobsToStart` local and return here.

lib/eventFactory.js
```diff
@@ -2,9 +2,131 @@
...
+            });
+        }
+
+        // When startFrom is jobName
+        if (startFrom !== '~pr') {
+            jobsToStart = jobs.filter(j => j.name === startFrom && j.state === 'ENABLED');
+
+            return Promise.resolve([]);
```
jithine: should make `jobsToStart` local and return here.

lib/eventFactory.js
```diff
@@ -2,9 +2,131 @@
...
+            // Get the job name part, e.g. main from PR-1:main
+            const missingJobNames = missingPRJobNames.map(name => name.split(':')[1]);
+            const missingJobs = jobs.filter(j => missingJobNames.includes(j.name));
+
+            return pipeline.getConfiguration(config.prRef)
+            .then((parsedConfig) => {
+                console.log('parsedConfig: ', parsedConfig.jobs);
+
+                // Create missing PR jobs
+                return Promise.all(missingJobs.map(j =>
```
jithine: add a then handler here and return combined jobsToStart
```
 return Promise.all(missingJobs.map(...)).then(newJobs => jobsToStart.concat(newJobs))
```
This way `newJobs` which is only for PR jobs does not leak outside, meaning outside success handler will only see `jobsToStart`

jithine (commented):
Looks good, just one feedback

lib/eventFactory.js
petey: You are already in a promise chain. Is returning `Promise.resolve` necessary?

lib/eventFactory.js
petey: You are already in a promise chain. Is returning `Promise.resolve` necessary?

lib/eventFactory.js
```diff
@@ -2,9 +2,131 @@
...
+            });
+        });
+    })
+    .then((newJobs) => {
+        // Add newly created PR jobs to jobsToStart array
+        jobsToStart = jobsToStart.concat(newJobs);
+
+        // No jobs to start (eg: when jobs are disabled or startFrom is not valid)
+        if (jobsToStart.length <= 0) {
+            return Promise.reject(new Error('No jobs to start'));
```
petey: You are already in a promise chain. Why not just `throw new Error(...)`?

lib/eventFactory.js
```diff
@@ -2,9 +2,131 @@
...
+ * @param  {String}   config.sha                 SHA this project was built on
+ * @param  {String}   config.username            Username of the user that creates this event
+ * @param  {String}   config.scmContext          The scm context to which user belongs
+ * @param  {String}   [config.prRef]             Ref if it's a PR event
+ * @param  {Number}   [config.prNum]             PR number if it's a PR event
+ * @param  {String}   [config.startFrom]         Where the event starts from (jobname or ~commit, ~pr, etc)
+ *                                               (Optional for backwards compatibility)
+ * @param  {Number}   eventId                    Event id
+ */
+function createBuilds(pipeline, config, eventId) {
```
petey: This method is getting really complex. Can this be refactored into separate methods?

lib/eventFactory.js
```diff
@@ -2,9 +2,200 @@
...
+    const { jobFactory, jobs, pipeline, prNum, startFrom } = config;
+
+    if (startFrom !== '~pr') {
+        return [];
+    }
+
+    return pipeline.syncPR(prNum)
+    .then(() => {
+        // Get next jobs for when startFrom is ~pr
+        const nextJobs = workflowParser.getNextJobs(pipeline.workflowGraph, {
```
minzcmu: This is not right. It should use the `config.workflow` from `pipeline.getConfiguration(config.prRef)` instead of the one from pipeline.

minzcmu (changes requested):

minzcmu (approved):

test/lib/eventFactory.test.js
```diff
@@ -156,27 +409,18 @@ describe('Event Factory', () => {
...
         );
-
-        it('should only update lastEventId if type is pipeline', () => {
-            config.type = 'pr';
-
-            return factory.create(config).then((model) => {
-                assert.instanceOf(model, Event);
-                // lastEventId should not have been updated
-                assert.calledOnce(pipelineMock.update);
-                assert.strictEqual(pipelineMock.lastEventId, null);
```
jithine, line 177: can you add this assertion to one of pr tests ?

jithine (approved):

test/lib/eventFactory.test.js
tkyi: added!

PR #198, opened 2017-10-17
fix: pipeline.getJobs don't sort by workflow if no workflow

minzcmu:
`pipeline.getJobs` don't sort by workflow if no workflow or length equals to 0. This will be the case for new workflow.

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -497,8 +497,8 @@ class PipelineModel extends BaseModel {
 
         return factory.list(listConfig)
             .then((jobs) => {
-                // If archived is true, don't sort and just return jobs
-                if (listConfig.params.archived) {
+                // If using the new config or archived is true, don't sort and just return jobs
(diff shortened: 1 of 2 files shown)


tkyi (approved):

lib/pipeline.js
```diff
@@ -497,8 +497,8 @@ class PipelineModel extends BaseModel {
 
         return factory.list(listConfig)
             .then((jobs) => {
-                // If archived is true, don't sort and just return jobs
-                if (listConfig.params.archived) {
+                // If using the new config or archived is true, don't sort and just return jobs
+                if (!workflow || workflow.length === 0 || listConfig.params.archived) {
```
jithine: it could be `!Array.isArray(worklflow) || workflow.length`

minzcmu: Nice!

jithine (approved):

PR #202, opened 2017-10-24
fix: force this.jobs to reload after sync

minzcmu:
In eventFactory, we called `pipeline.sync()` first and then call `pipeline.jobs`.
https://github.com/screwdriver-cd/models/blob/master/lib/eventFactory.js#L275
https://github.com/screwdriver-cd/models/blob/master/lib/eventFactory.js#L163

This leads to a weird problem that `pipeline.jobs` only contains an old list of jobs. 
That's because we call `this.jobs` inside pipeline `sync()` first, and update/create jobs afterwards. `this.jobs` is set and cached. So eventFactory only gets the old jobs. 

Delete this.jobs will force a DB query when next time `pipeline.jobs` is called.

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -340,6 +340,10 @@ class PipelineModel extends BaseModel {
                     }
                 });
 
+                // jobs updated or new jobs created during sync
+                // delete it here so next time this.jobs is called a DB query will be forced and new jobs will return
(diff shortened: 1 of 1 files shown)


tkyi (approved):

d2lam (approved):
Interesting. Let's try 👍

PR #203, opened 2017-10-27
feat(770): support external trigger (rebuild_on)

d2lam:
This PR creates Trigger Factory, Trigger Model, and Sync external triggers. 
There are a couple scenarios:
- For an existing job: if job is not in the yaml anymore, requireList = []; if job is still in yaml, requireList = requires value of that job config
- It is a new job: requireList = requires value of that job config

For syncExternalTrigger: list triggers that has `dest` as the current job. If the record is not in the new requireList, then remove it. If the requireList has new values that are not currently in the database, then create it. 

Blocked by: https://github.com/screwdriver-cd/data-schema/pull/190
Related: https://github.com/screwdriver-cd/screwdriver/issues/770

### DIFF
--- a/README.md
+++ b/README.md
@@ -871,6 +871,61 @@ factory.get(id).then(model => {
 | :-------------   | :---- | :-------------|
 | id | Number | The unique ID for the Template Tag |
 
+### Trigger Model
+```js
+'use strict';
+const Model = require('screwdriver-models');
+const factory = Model.TriggerFactory.getInstance({
+    datastore
+});
+const config = {
+    dest: '~sd@123:component',
+    src: '~sd@456:main'
+}
+
+factory.create(config)
+    .then(model => {    // do something
+});
+```
+
+### Trigger Factory
+#### Create
+```js
+factory.create(config).then(model => {
+    // do stuff with trigger model
+});
+```
+
+| Parameter        | Type  |  Required | Description |
+| :-------------   | :---- | :-------------|  :-------------|
+| config        | Object | Yes | Configuration Object |
+| config.src | String | Yes | The job that initiates the trigger (ex: ~sd@123:component) |
+| config.dest | String | Yes | The job that is triggered (ex: ~sd@456:main) |
+
+#### Get
+Get trigger based on id.
+```js
+factory.get(id).then(model => {
+    // do stuff with trigger model
+});
+```
+
+| Parameter        | Type  |  Description |
+| :-------------   | :---- | :-------------|
+| id | Number | The unique ID for the trigger |
+
+#### List
+List triggers that have dest as `~sd@456:main`
+```js
+// update template version value
+factory.list({
+    params: {
+        dest: '~sd@456:main'
+    }
+}).then(recs => 
+    // do things with the records 
+);
+```
 ### Token Factory
 #### Search
 ```js
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -335,11 +335,11 @@ class PipelineModel extends BaseModel {
                         const triggersToSync = [];
                         const parsedConfigJobNames = Object.keys(parsedConfig.jobs);
                         const pipelineId = this.id;
-                        let requiresList = [];
 
                         // Loop through all existing jobs
                         existingJobs.forEach((job) => {
                             const jobName = job.name;
+                            let requiresList = [];
 
                             // if it's in the yaml, update it
                             if (parsedConfigJobNames.includes(jobName)) {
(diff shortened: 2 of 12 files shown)


lib/pipeline.js
```diff
@@ -6,14 +6,55 @@ const BaseModel = require('./base');
...
+    const dest = `~sd@${config.pipelineId}:${config.jobName}`;
+    const processed = [];
+
+    // list records that would trigger this job
+    return triggerFactory.list({ params: { dest } })
+    .then((records) => {
+        // if the src is not in the new src list, then remove it
+        const toRemove = records.filter(rec => !newSrcList.includes(rec.src));
+
+        // if the new src is not in the old src list, then create in datastore
```
tkyi: I think you can omit "new" here so it's consistent with the comment above. Otherwise the two src sound like different src to me.

lib/pipeline.js
```diff
@@ -291,43 +332,69 @@ class PipelineModel extends BaseModel {
                     .then((existingJobs) => {
                         const jobsToSync = [];
                         const jobsProcessed = [];
+                        const triggersToSync = [];
                         const parsedConfigJobNames = Object.keys(parsedConfig.jobs);
+                        const pipelineId = this.id;
+                        let requiresList;
```
tkyi: Can you set `requiresList = []` here? Then you don't need to have line 354 and the default value on line 363

d2lam: Sure. I can remove 354 with this but still need `[]` as default value for the case where it sets 
`requireList = permutations[0].requires` and it's set to `undefined` if requires is not there.

d2lam: Added the default case inside the other if instead :)

lib/pipeline.js
d2lam: Fixed!

tkyi (approved):

lib/pipeline.js
```diff
@@ -291,43 +332,68 @@ class PipelineModel extends BaseModel {
...
+                                job.permutations = permutations;
                                 job.archived = false;
                                 jobsToSync.push(job.update());
-                            // if it's not in the yaml and archive it
+                            // if it's not in the yaml then archive it
                             } else if (!job.isPR()) {
                                 job.archived = true;
                                 jobsToSync.push(job.update());
                             }
+
```
minz1027: Should also clean up the triggers table for jobs that get archived. E.g. I previously have a job `A` requries `[~sd@123:main]`. After I removed job A from my yaml, there shouldn't be a trigger record for `{ src: ~sd@123:main, dest: ~sd@456:A }`

lib/pipeline.js
```diff
@@ -291,43 +332,68 @@ class PipelineModel extends BaseModel {
                     .then((existingJobs) => {
                         const jobsToSync = [];
                         const jobsProcessed = [];
+                        const triggersToSync = [];
                         const parsedConfigJobNames = Object.keys(parsedConfig.jobs);
+                        const pipelineId = this.id;
+                        let requiresList = [];
```
minzcmu: I think it's better to move `let requiresList = [];` inside existingJobs.forEach otherwise this can cause problem. E.g. for jobA requiresList = permutations[0].requires, jobB is not in the yaml. Then for jobA it got the requiresList from A since it's not reset.

minzcmu (changes requested):

lib/triggerFactory.js
```diff
@@ -0,0 +1,45 @@
...
+    constructor(config) {
+        super('trigger', config);
+    }
+
+    /**
+     * Instantiate a Trigger class
+     * @method createClass
+     * @param  {Object}     config               Trigger data
+     * @param  {Datastore}  config.datastore     Object that will perform operations on the datastore
+     * @param  {Number}     config.src           Job that initiates the trigger
```
minzcmu: String

minzcmu (changes requested):

minzcmu:
README?

README.md
```diff
@@ -871,6 +871,61 @@ factory.get(id).then(model => {
...
+### Trigger Model
+```js
+'use strict';
+const Model = require('screwdriver-models');
+const factory = Model.TriggerFactory.getInstance({
+    datastore
+});
+const config = {
+    dest: '~sd@123:component',
+    tag: '~sd@456:main'
```
minzcmu: tag? Other than this, it LGTM 👍

minzcmu (changes requested):

minzcmu (approved):

PR #204, opened 2017-11-02
fix(770): External trigger should start one event

tkyi:
Currently, an external trigger will start different events for each job that requires an external trigger.

This PR passes along the entire external trigger as a `startFrom` so one event can be created.

Related to https://github.com/screwdriver-cd/screwdriver/pull/779

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -2,6 +2,7 @@
 
 const BaseFactory = require('./baseFactory');
 const Event = require('./event');
+const { EXTERNAL_TRIGGER } = require('screwdriver-data-schema').config.regex;
 const workflowParser = require('screwdriver-workflow-parser');
 
 let instance;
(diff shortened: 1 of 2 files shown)


lib/eventFactory.js
```diff
@@ -38,20 +39,21 @@ function splitPRJobs(config) {
...
  * @param  {Array}    config.jobs                           Array of job objects
  * @param  {Object}   config.pipelineConfig
  * @param  {Array}    [config.pipelineConfig.workflow]      Order of jobs to be run
  * @param  {Object}   config.pipelineConfig.workflowGraph   Object with nodes and edges that represent the order of jobs
- * @param  {String}   config.startFrom                      Startfrom (e.g. ~commit, ~pr, etc)
+ * @param  {String}   config.startFrom                      Startfrom (e.g. ~commit, ~pr, ~sd@123:main, etc)
  * @return {Array}                                          Array of commit jobs to start
  */
-function getJobsFromCommit(config) {
+function getJobsFromTrigger(config) {
```
d2lam: We still need the `getJobsFromCommit` right?

lib/eventFactory.js
```diff
@@ -38,20 +39,21 @@ function splitPRJobs(config) {
...
     const { jobs, pipelineConfig, startFrom } = config;
+    const isTrigger = EXTERNAL_TRIGGER.test(startFrom);
 
-    if (startFrom !== '~commit') {
+    if (startFrom !== '~commit' && !isTrigger) {
```
d2lam, line 56: This means `getJobsFromCommits` won't work anymore?

d2lam (changes requested):

lib/eventFactory.js
d2lam: Oh I see. This should handle both? 
The function name is a little confusing to me.

d2lam (approved):

minzcmu (approved):
👍 LGTM

PR #205, opened 2017-11-06
fix: create PR jobs when new jobs added to the config in PR

minzcmu:
Removing  `const missingJobs = jobs.filter(j => missingJobNames.includes(j.name));`

We are only using `missingJobs.name` in the following calls. So there is no need to get the job objects. And this is giving side effect that when user adds a new job in PR, e.g. `new_main` and requires `~pr`, that PR job won't be created because it's filtered out by the above line.

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -16,7 +16,7 @@ let instance;
  * @param  {Array}     config.jobs      PR jobs
  * @param  {Array}     config.nextJobs  PR job names to trigger next
  * @param  {Number}    config.prNum     PR number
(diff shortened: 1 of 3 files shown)


d2lam:
Any test change?

minz1027:
@d2lam I can add a test to test the case I mentioned

minz1027:
@d2lam addressed

tkyi (approved):

minzcmu:
@d2lam test modified

d2lam (approved):

PR #206, opened 2017-11-28
fix(771): allow meta to be passed externally

d2lam:
Launcher will look up the meta from `parentBuildId`: https://github.com/screwdriver-cd/launcher/blob/master/launch.go#L202-L222

However, this value is not set when being triggered by an external pipeline. This and https://github.com/screwdriver-cd/screwdriver/pull/807 will fix the issue. There is actually no code change, but it will work because we will now pass in the `parentBuildId` from API. 

Related: 
https://github.com/screwdriver-cd/screwdriver/issues/771

### DIFF
--- a/lib/buildFactory.js
+++ b/lib/buildFactory.js
@@ -145,13 +145,14 @@ class BuildFactory extends BaseFactory {
     /**
      * Create a new Build
      * @method create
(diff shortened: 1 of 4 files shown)


tkyi (approved):

minzcmu (approved):
👍

PR #211, opened 2018-01-29
feat: add pipelineId to command

kumada626:
## Context
Like template, to limit a pipeline which updates the command only a pipeline which created the command, I add a pipelineId to the command in [data-schema](https://github.com/screwdriver-cd/data-schema/pull/199). Related to data-schema, I fixed models.

## Objective
Add pipelineId to the command.

## Note
Unit tests will fail for the moment, because these models require `command` data-schema.
https://github.com/screwdriver-cd/data-schema/pull/199

## To do before merged
- [x] update the screwdriver-data-schema dependency.

## Related
https://github.com/screwdriver-cd/data-schema/pull/199

### DIFF
--- a/lib/commandFactory.js
+++ b/lib/commandFactory.js
@@ -35,6 +35,7 @@ class CommandFactory extends BaseFactory {
      * @param  {Object}     config.habitat       Habitat config of the command
      * @param  {Object}     config.docker        Docker config of the command
      * @param  {Object}     config.binary        Binary config of the command
+     * @param  {String}     config.pipelineId    pipelineId of the command
      * @return {Command}
      */
     createClass(config) {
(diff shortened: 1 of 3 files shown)


lib/commandFactory.js
```diff
@@ -55,6 +56,7 @@ class CommandFactory extends BaseFactory {
      * @param  {Object}     config.habitat       Habitat config of the command
      * @param  {Object}     config.docker        Docker config of the command
      * @param  {Object}     config.binary        Binary config of the command
+     * @param  {Object}     config.pipelineId    pipelineId of the command
```
s-yoshika: Object?

kumada626: Oh, it's not an Object, it is a String. Thanks.

tkyi (approved):

d2lam (approved):

PR #215, opened 2018-02-16
fix: Update iron and fix seal and unseal token funcs

tkyi:
## Context
Iron has changed its method signatures for seal and unseal.
They no longer have callbacks in their method signature(https://github.com/hueniverse/iron/blob/master/lib/index.js#L202) and will throw an error rather than return it as the first arg.

## Objective
This PR pulls in the new Iron version and updates how iron methods are called according to their README (https://github.com/hueniverse/iron/commit/e95a5c23149efb5361967abf01152e4d40836fdb#diff-04c6e90faac2675aa89e2176d2eec7d8).

## Related links
Related to https://github.com/screwdriver-cd/models/pull/214

### DIFF
--- a/lib/secret.js
+++ b/lib/secret.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const BaseModel = require('./base');
-const nodeify = require('./nodeify');
 const iron = require('iron');
 // Get symbols for private fields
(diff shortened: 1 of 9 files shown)


minzcmu (approved):
👍 LGTM

PR #216, opened 2018-02-16
Revert "fix: Update iron and fix seal and unseal token funcs"

tkyi:
Reverts screwdriver-cd/models#215

The new iron package requires node8 functionality. We are still using node6.
Rolling back so we can unblock the screwdriver-cd/screwdriver pipeline.


minzcmu (approved):
👍

PR #223, opened 2018-02-26
chore: Upgrading versions of mocha, chai, and sinon

stjohnjohnson:
I had to update two tests because the tests were not actually working as expected and Chai was just ignoring the error.  Latest version of Chai does not ignore that error anymore.


minzcmu (approved):
👍

PR #224, opened 2018-02-27
feat: get parentEventId config when recreate

d2lam:
API now passes in `workflowGraph` so that when we create a new event, it can take the workflowGraph of the parentEvent, instead of the latest config. 

## References
Related to https://github.com/screwdriver-cd/screwdriver/issues/806

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -263,7 +263,7 @@ class EventFactory extends BaseFactory {
      *                                              Optional for backwards compatibility
      * @param  {String}  [config.causeMessage]      Message that describes why the event was created
      * @param  {String}  [config.parentBuildId]     Id of the build that starts this event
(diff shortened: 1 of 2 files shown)


lib/eventFactory.js
```diff
@@ -314,34 +314,43 @@ class EventFactory extends BaseFactory {
                         modelConfig.commit = commit;
                         modelConfig.createTime = (new Date()).toISOString();
 
+                        if (config.workflowGraph) {
```
d2lam, line 317: Note: this can be combined with the ` if (config.parentEventId)` later but having it here to be backward compatible. Since currently, having `parentEventId` doesn't mean it will have `workflowGraph` for sure.

minzcmu (approved):
👍 LGTM

philipjscott (approved):

FenrirUnbound (approved):
👍

PR #225, opened 2018-03-01
fix: make sync take in optional ref

minzcmu:
## Context
When user restarts a job from a past event. We want to fetch the exact same screwdriver.yaml in that event/sha and create/update jobs accordingly.

## Objective
make `pipeline.sync()` take in optional ref so that it can fetch the exact job config from parentEvent.

## Related to https://github.com/screwdriver-cd/screwdriver/issues/806

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -200,6 +200,7 @@ class EventFactory extends BaseFactory {
      *                                              Optional for backwards compatibility
      * @param  {String}  [config.causeMessage]      Message that describes why the event was created
      * @param  {String}  [config.parentBuildId]     Id of the build that starts this event
+     * @param  {String}  [config.parentEventId]     Id of the parent event
      * @param  {String}  [config.workflowGraph]     workflowGraph of parentEvent if there is a parentEvent
      * @return {Promise}
      */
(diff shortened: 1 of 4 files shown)


lib/eventFactory.js
```diff
@@ -211,65 +212,69 @@ class EventFactory extends BaseFactory {
...
+                    // Get latest config
+                    return getLatestWorkflowGraph({
+                        pipeline,
+                        eventConfig: config
+                    }).then((latestConfig) => {
+                        if (modelConfig.type === 'pr') {
+                            return latestConfig;
                         }
+                        modelConfig.workflowGraph = latestConfig.workflowGraph;
+                        modelConfig.workflow = latestConfig.workflow; // TODO: deprecate this
```
d2lam: We can remove this

lib/eventFactory.js
```diff
@@ -211,65 +212,69 @@ class EventFactory extends BaseFactory {
...
+                    modelConfig.commit = commit;
+                    modelConfig.createTime = (new Date()).toISOString();
+
+                    // Get latest config
+                    return getLatestWorkflowGraph({
+                        pipeline,
+                        eventConfig: config
+                    }).then((latestConfig) => {
+                        if (modelConfig.type === 'pr') {
+                            return latestConfig;
```
d2lam: we don't need this anymore

lib/eventFactory.js
```diff
@@ -211,65 +212,69 @@ class EventFactory extends BaseFactory {
...
                         }
+                        modelConfig.workflowGraph = latestConfig.workflowGraph;
+                        modelConfig.workflow = latestConfig.workflow; // TODO: deprecate this
 
-                        return super.create(modelConfig);
-                    })
+                        return modelConfig;
```
d2lam: should call `super.create(modelConfig)` here

d2lam (changes requested):

d2lam (approved):

PR #226, opened 2018-03-01
fix: syncPR creates missing PR jobs

d2lam:
## Context
When create the actual event, we need `modelConfig` https://github.com/screwdriver-cd/models/blob/2bf6214146427ff8b71a2f3e8516e157e55a18ab/lib/eventFactory.js#L327

After creating the event, for PR job, `getLatestConfig` returns the latest pipeline config for that prRef and create missing PR jobs: https://github.com/screwdriver-cd/models/blob/2bf6214146427ff8b71a2f3e8516e157e55a18ab/lib/eventFactory.js#L339

This behavior got messed up from this PR: https://github.com/screwdriver-cd/models/pull/224/files 
https://github.com/screwdriver-cd/models/blob/d3cb7e40ab2389e3404500f38a7a41174330fca7/lib/eventFactory.js#L328-L334
because we need return BOTH `modelConfig` and the `latestConfig` (for PRs). For non-PR, it only needs `modelConfig`. 

This behavior requires us to call `getConfiguration` twice for PR jobs, and it's confusing. We only need the `pipelineConfig` because we want to create the missing PR jobs. Ideally, this should happen when we syncPR, since the expected behavior of syncing is to correct the state. So this PR moves the logic there and remove the needs of `pipelineConfig` inside event create.

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -8,36 +8,6 @@ const workflowParser = require('screwdriver-workflow-parser');
 let instance;
 
 /**
- * Returns 2 arrays:
- * 1. Jobs to start
- * 2. Job to create
- * @method splitPRJobs
- * @param  {Object}    config
- * @param  {Array}     config.jobs      PR jobs
- * @param  {Array}     config.nextJobs  PR job names to trigger next
- * @param  {Number}    config.prNum     PR number
- * @return {Object}                     Object with jobsToStart array of job object and jobsToCreate array of jobName
- */
-function splitPRJobs(config) {
-    const { jobs, nextJobs, prNum } = config;
-    // Get jobsToStart
-    // To start all existing and ENABLED PR-prNum jobs
-    const PRJobArray = jobs.filter(j => j.name.startsWith(`PR-${prNum}:`));
-    const jobsToStart = PRJobArray.filter(j => j.state === 'ENABLED');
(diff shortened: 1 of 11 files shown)


lib/eventFactory.js
```diff
@@ -314,43 +252,37 @@ class EventFactory extends BaseFactory {
...
-                        }).then((latestConfig) => {
-                            if (modelConfig.type === 'pr') {
-                                return latestConfig;
-                            }
+                        });
+                    })
+                    .then((latestConfig) => {
+                        if (config.parentEventId) {
+                            modelConfig.parentEventId = config.parentEventId;
+                            modelConfig.workflowGraph = config.workflowGraph;
```
minzcmu, line 264: Don't need this line any more. We will get the `workflowGraph` form latestConfig.

minzcmu (changes requested):

lib/eventFactory.js
d2lam: We still need this for the restart of event, so it will get the workflowGraph from the parent event.

minzcmu: Cool cool, messed this with my PR. Once we have `pipeline.sync(ref)`, we can remove this. 👍

minzcmu (approved):

PR #227, opened 2018-03-01
fix: moves syncPR up and returns the new pipeline

d2lam:
In https://github.com/screwdriver-cd/models/pull/226, it calls `getLatestConfig` and creates new jobs inside `syncPR`, but this new pipeline needs to return. Otherwise the `pipeline.jobs` is from the previous pipeline before the sync. 

This PR moves the logic outside for `getLatestConfig` and make sure it gets the new pipeline after syncing.

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -144,29 +144,24 @@ function createBuilds(config) {
 }
 
 /**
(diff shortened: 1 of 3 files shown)


minzcmu (approved):
👍

PR #228, opened 2018-03-07
fix(881): Ignore scm updateCommitStatus errors

philipjscott:
# Context

From https://developer.github.com/v3/repos/statuses/

>Note: there is a limit of 1000 statuses per sha and context within a repository. Attempts to create more than 1000 statuses will result in a validation error.

There are build monitoring pipelines that run periodically, and reach the limit noted above. When the error is thrown, the pipeline stops entirely.

# Objective

Allow pipelines to continue if Github commit status fails to update.

# References

* Issue: https://github.com/screwdriver-cd/screwdriver/issues/881


lib/build.js
```diff
@@ -53,7 +53,7 @@ class BuildModel extends BaseModel {
                 pipelineId: pipeline.id
             };
 
-            return this.scm.updateCommitStatus(config);
+            return this.scm.updateCommitStatus(config).catch(() => null);
```
jithine: can you log the error so that we can mine it later ?

test/lib/build.test.js
```diff
@@ -162,13 +162,15 @@ describe('Build Model', () => {
                 })
         );
 
-        it('reject on error', () => {
+        it('it discards rejections and resolves null', () => {
```
tkyi: don't need to start statement with "it"

jithine (approved):

lib/build.js
```diff
@@ -53,7 +53,12 @@ class BuildModel extends BaseModel {
                 pipelineId: pipeline.id
             };
 
-            return this.scm.updateCommitStatus(config);
+            return this.scm.updateCommitStatus(config).catch((err) => {
```
jithine, line 57: should we check the error and ignore only the rate limit one ?

lib/build.js
```diff
@@ -53,7 +54,24 @@ class BuildModel extends BaseModel {
...
+                const shaLimitErr = {
```
jithine, line 58: is this how `scm-github` works ? is there a simple API status code or something Github API returns ?

philipjscott: `scm-github` just relays the errors thrown from the github package on NPM. See https://developer.github.com/v3/repos/statuses/#create-a-status and https://github.com/screwdriver-cd/screwdriver/issues/881

To be more general, we could just check `err.message === 'Validation Failed'`. The disadvantage with this approach is that validation errors unrelated to the sha limit issue will be ignored.

d2lam: The disadvantage of this is if the github API changes the message slightly, this won't work anymore. 🤷‍♀️ 
So they don't have a separate status code for this? Can't tell from the docs.

philipjscott: @d2lam it doesn't seem like it. If they did have a separate status code for this, wouldn't we see it in the JSON? Looking at the SHA limit error, its error code is `"custom"`, which isn't very helpful.

philipjscott: After testing with `github@8.1.1`, it seems the thrown error is different than I expected:

```
{ [Error: {"message":"Validation Failed","errors":[{"resource":"Status","code":"custom","message":"This SHA and context has reached the maximum number of statuses."}],"documentation_url":"https://developer.github.com/enterprise/2.10/v3/repos/statuses/#create-a-status"}]
  message: '{"message":"Validation Failed","errors":[{"resource":"Status","code":"custom","message":"This SHA and context has reached the maximum number of statuses."}],"documentation_url":"https://developer.github.com/enterprise/2.10/v3/repos/statuses/#create-a-status"}',
  code: 422,
  status: 'Unprocessable Entity',
  headers: 
   { server: 'GitHub.com',
     date: 'Fri, 09 Mar 2018 20:45:20 GMT',
     'content-type': 'application/json; charset=utf-8',
     'content-length': '259',
     connection: 'close',
     status: '422 Unprocessable Entity',
     'x-oauth-scopes': 'repo',
     'x-accepted-oauth-scopes': '',
     'x-github-media-type': 'github.v3; format=json',
     'access-control-expose-headers': 'ETag, Link, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval',
     'access-control-allow-origin': '*',
     'x-github-request-id': '98dc3a60-5cf4-4bce-aa17-3be4e551b72c',
     'content-security-policy': 'default-src \'none\'',
     'strict-transport-security': 'max-age=31536000; includeSubdomains',
     'x-content-type-options': 'nosniff',
     'x-frame-options': 'deny',
     'x-xss-protection': '1; mode=block' } }
```

`err.toString()` evaluates to:

```json
{
   "message":"Validation Failed",
   "errors":[
      {
         "resource":"Status",
         "code":"custom",
         "message":"This SHA and context has reached the maximum number of statuses."
      }
   ],
   "documentation_url":"https://developer.github.com/enterprise/2.10/v3/repos/statuses/#create-a-status"
}
```

Perhaps we could just check for `err.code === 422`?

jithine: yep, that's the right approach checking for code status`422`, if it's available. Perhaps this require change in `scm-github` to surface this status back to model ?

philipjscott: @jithin1987 Yep. Since this is Github specific, it makes sense to check for the error there. I'll make a PR in that repository and close this PR once that's done.

philipjscott:
New PR addressing issue in `scm-github` layer: https://github.com/screwdriver-cd/scm-github/pull/80

PR #229, opened 2018-03-09
fix: allow create build without starting

d2lam:
We never needed this use case before. 
Now, we want to create a build with just parentBuildId, and do not start the build. 
Related: https://github.com/screwdriver-cd/screwdriver/issues/904
Blocked by: https://github.com/screwdriver-cd/data-schema/pull/210


minzcmu (approved):
👍

PR #231, opened 2018-03-26
fix: set meta for event as {} by default

minzcmu:
## Context
There is a bug right now that you cannot set or get meta when you try to run a detached pipeline when the parent event doesn't have metadata.

## Objective
This PR will set meta for event as {} by default.

### DIFF
--- a/test/lib/eventFactory.test.js
+++ b/test/lib/eventFactory.test.js
@@ -157,7 +157,8 @@ describe('Event Factory', () => {
                 causeMessage: 'Started by github:stjohn',
                 createTime: nowTime,
                 creator,
-                commit
+                commit,
+                meta: {}
             };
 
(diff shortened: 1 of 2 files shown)


test/lib/eventFactory.test.js
```diff
@@ -392,7 +393,8 @@ describe('Event Factory', () => {
                 });
                 assert.strictEqual(syncedPipelineMock.lastEventId, model.id);
                 Object.keys(expected).forEach((key) => {
-                    if (key === 'workflowGraph') {
+                    if (key === 'workflowGraph' || key === 'meta') {
+                        assert.deepEqual(model[key], expected[key]);
```
tkyi: can remove this duplicate line

tkyi (approved):

PR #232, opened 2018-03-27
feat(911): Support subdirectory [6]

d2lam:
Instead of starting builds right after we get the `jobsToStart`, it will need to check if `sourcePaths` exists. If it exists, then check if the `changedFiles` is in `sourcePaths`. Only then, it will start the job.

Blocked by: 
- https://github.com/screwdriver-cd/scm-github/pull/83
- https://github.com/screwdriver-cd/config-parser/pull/56
- https://github.com/screwdriver-cd/screwdriver/pull/940

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -173,7 +173,15 @@ function createBuilds(config) {
                     changedFiles,
                     sourcePaths: j.sourcePaths
                 })
-            ));
+            )).then((buildsCreated) => {
+                const nobuilds = buildsCreated.every(b => b === null);
+
+                if (nobuilds) {
+                    throw new Error('No jobs to start');
+                }
+
+                return buildsCreated;
+            });
         });
 }
 
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -100,7 +100,7 @@ function startBuild(config) {
                 }
 
                 // source path is directory
-                return file.includes(source);
+                return file.startsWith(source);
             }));
     }
 
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -74,10 +74,13 @@ function getJobsFromPR(config) {
 }
 
 /**
- * [createBuild description]
+ * Starts the build if the changed file is part of the sourcePaths, or if there is no sourcePaths
  * @method startBuild
- * @param  {[type]}    config [description]
- * @return {[type]}           [description]
+ * @param  {Object}   config                configuration object
+ * @param  {Object}   config.buildConfig    Build Config to create the build with
+ * @param  {Array}    config.changedFiles   List of files that were changed
+ * @param  {Array}    config.sourcePaths    List of soure paths
+ * @return {Promise}
  */
 function startBuild(config) {
     /* eslint-disable global-require */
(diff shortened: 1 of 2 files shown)


lib/eventFactory.js
```diff
@@ -73,6 +73,32 @@ function getJobsFromPR(config) {
...
+    const BuildFactory = require('./buildFactory');
+    const buildFactory = BuildFactory.getInstance();
+    /* eslint-enable global-require */
+    const { buildConfig, changedFiles, sourcePaths } = config;
+    let hasChangeInSourcePaths = true;
+
+    // Only check if sourcePaths exists
+    if (sourcePaths) {
+        hasChangeInSourcePaths = changedFiles.some(file =>
+            sourcePaths.some(source => file.includes(source)));
```
tkyi: What if `sourcePaths = ['src/folder']` and `changedFiles = ['src/folder/file.txt']`? Will this logic still work?

lib/eventFactory.js
```diff
@@ -73,6 +73,32 @@ function getJobsFromPR(config) {
...
+    /* eslint-disable global-require */
+    const BuildFactory = require('./buildFactory');
+    const buildFactory = BuildFactory.getInstance();
+    /* eslint-enable global-require */
+    const { buildConfig, changedFiles, sourcePaths } = config;
+    let hasChangeInSourcePaths = true;
+
+    // Only check if sourcePaths exists
+    if (sourcePaths) {
+        hasChangeInSourcePaths = changedFiles.some(file =>
```
tkyi, line 95: Should we short circuit if `changedFiles = []`?

lib/eventFactory.js
d2lam: Yes. that is the test case. https://github.com/screwdriver-cd/models/pull/232/files#diff-88f33fafc6c2f1254b793e6e56a1917aR458

lib/eventFactory.js
d2lam: If it's `[]`, I think it will just be similar to `[].forEach`, which will do nothing. So I don't think we need to short circuit.

lib/eventFactory.js
d2lam: It goes through the changed files list and check if **ANY** of the changed file includes **ANY** of the sourcePaths

`'src/folder/file.txt'.includes('src/folder') === true`

**[EDIT]**
Actually, I see some problems with this, because
`'src/folder/file.txt'.includes('src/fold')` is also true

Probably need to do some extra parsing...

lib/eventFactory.js
tkyi: Okay

lib/eventFactory.js
```diff
@@ -73,6 +73,42 @@ function getJobsFromPR(config) {
...
+            sourcePaths.some((source) => {
+                let path = source;
+
+                // make sure it's in this format foo/bar/
+                // because with changedFile as 'foo/barTest.md', it should return false
```
tkyi: Wondering if it's possible for a `sourcePath` to be an exact path to a file, because this logic would break in that case

test/lib/eventFactory.test.js
```diff
@@ -7,12 +7,12 @@ const PARSED_YAML = require('../data/parserWithWorkflowGraph');
 
 sinon.assert.expose(assert, { prefix: '' });
 
-describe('Event Factory', () => {
+describe.only('Event Factory', () => {
```
tkyi: Remove "only"

lib/eventFactory.js
```diff
@@ -73,6 +73,40 @@ function getJobsFromPR(config) {
...
+    if (sourcePaths) {
+        hasChangeInSourcePaths = changedFiles.some(file =>
+            sourcePaths.some((source) => {
+                // source path is a file
+                if (source.slice(-1) !== '/') {
+                    return file === source;
+                }
+
+                // source path is directory
+                return file.includes(source);
```
tkyi: Do we need to check if the path matches from the beginning?

For example,
if `file = 'foo/bar/baz/meow.txt'` and `source = 'bar/baz/'`, `file.includes(source)` will still return `true`.

tkyi (approved):
LGTM 👍

lib/eventFactory.js
```diff
@@ -138,7 +168,11 @@ function createBuilds(config) {
 
             // Start builds
             return Promise.all(jobsToStart.map(j =>
-                buildFactory.create(Object.assign({ jobId: j.id, eventId }, eventConfig))
+                startBuild({
```
minzcmu: Can we throw err when no build can be started? it's easier for the user and us to figure out what happened when no build is starting

minzcmu (changes requested):

minzcmu (approved):

PR #233, opened 2018-03-29
fix: sourcePaths is part of permutations

d2lam:
The job object will look like this
```
    {
        "id": 3220,
        "name": "job1",
        "permutations": [
            {
                "annotations": {},
                "commands": [
                    {
                        "name": "echo",
                        "command": "echo hi"
                    }
                ],
                "environment": {},
                "image": "node:6",
                "secrets": [],
                "settings": {},
                "requires": [
                    "~commit",
                    "~pr"
                ],
                "sourcePaths": [
                    "README.md"
                ]
            }
        ],
        "pipelineId": 526,
        "state": "ENABLED",
        "archived": false
    }
```

So we need to get it from `permutations` instead
Also, fix the mocks to be more accurate, since `permutations` is an array. 

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -171,7 +171,7 @@ function createBuilds(config) {
                 startBuild({
                     buildConfig: Object.assign({ jobId: j.id, eventId }, eventConfig),
                     changedFiles,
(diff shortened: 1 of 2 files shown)


tkyi (approved):
👍

PR #234, opened 2018-03-30
fix: return error if the SCM does not support sourcepaths

d2lam:
Because it's harder for some SCMs to support this feature: https://github.com/screwdriver-cd/screwdriver/issues/911, this PR makes it optional. 

For SCMs that do not support the feature, `getChangedFiles` will return null. In such case, if `sourcePaths` is configured, then it will throw an error. 

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -92,6 +92,9 @@ function startBuild(config) {
 
     // Only check if sourcePaths exists
     if (sourcePaths) {
+        if (!changedFiles) {
+            throw new Error('Your SCM does not support Source Paths');
+        }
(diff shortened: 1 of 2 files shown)


tkyi:
Was wondering if it should throw an error or should it just ignore the `sourcePaths`?

d2lam:
@tkyi  Throwing error would be more obvious to users that it's not supported. Otherwise, they might think it is supported and our system is doing something wrong.

tkyi (approved):
👍

philipjscott (approved):
LGTM 👍 

doggo

minzcmu (approved):
👍

PR #235, opened 2018-04-03
fix(911): Only check for source paths if webhook

tkyi:
## Context

Restarting builds is broken right now after adding source paths feature. If a user has defined `sourcePaths` and they are trying to rerun a build from a past event, we get this error: `Cannot read property 'some' of undefined`, which is probably because `changedFiles` is undefined in that case (https://github.com/screwdriver-cd/models/blob/master/lib/eventFactory.js#L98).

## Objective

This PR will use a `webhooks` flag to determine if `sourcePaths` needs to be checked when starting a build or not.

## Related links
Related to:  https://github.com/screwdriver-cd/screwdriver/pull/975
Original issue: https://github.com/screwdriver-cd/screwdriver/issues/911


minzcmu (approved):
👍

d2lam (approved):

PR #237, opened 2018-04-11
fix(723): deprecate old workflow

d2lam:
No longer support the old workflow

There is some **logic change** in this PR

Related: https://github.com/screwdriver-cd/screwdriver/issues/723


minzcmu (approved):
👍

Filbird (approved):

PR #238, opened 2018-04-26
fix: add SD_SOURCE_PATH env variable for passing matched source path

jithine:
## Context

`sourcePaths` allows users to selectively run jobs if commit files happens to be on the paths/files mentioned in `sourcePaths`. But users have no means to programmatically find the actual source which was matched when `sourcePaths` is used.

This ENV variable allows users to write scripts which makes use of the actual source which was matched.

## References
Sub-directory support https://github.com/screwdriver-cd/screwdriver/issues/911

### DIFF
--- a/lib/buildFactory.js
+++ b/lib/buildFactory.js
@@ -149,6 +149,7 @@ class BuildFactory extends BaseFactory {
      * @param  {String}    [config.prRef]           The PR branch or reference
      * @param  {String}    [config.parentBuildId]   Id of the build that triggers this build
      * @param  {Boolean}   [config.start]           Whether to start the build after creating
+     * @param  {Object}    [config.environment]     Dynamically injected environment variables
      * @return {Promise}
      */
     create(config) {
--- a/lib/buildFactory.js
+++ b/lib/buildFactory.js
@@ -192,7 +192,12 @@ class BuildFactory extends BaseFactory {
                         container: permutation.image,
                         dockerRegistry: this.dockerRegistry
                     });
-                    modelConfig.environment = permutation.environment;
+
+                    modelConfig.environment = Object.assign({},
+                        modelConfig.environment,
+                        permutation.environment
+                    );
+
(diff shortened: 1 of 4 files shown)


test/lib/buildFactory.test.js
```diff
@@ -420,6 +420,32 @@ describe('Build Factory', () => {
...
+            return factory.create({
+                username,
+                jobId,
+                eventId,
+                parentBuildId: 12345,
+                start: false,
+                environment: { EXTRA: true }
+            }).then(() => {
+                assert.notCalled(startStub);
+                saveConfig.params.environment.EXTRA = true;
```
tkyi, line 444: you can move this line up before the `then`

tkyi (approved):

test/lib/buildFactory.test.js
jithine: no, this is just modifying the expected value.

lib/buildFactory.js
```diff
@@ -192,7 +192,12 @@ class BuildFactory extends BaseFactory {
                         container: permutation.image,
                         dockerRegistry: this.dockerRegistry
                     });
-                    modelConfig.environment = permutation.environment;
+
+                    modelConfig.environment = Object.assign({},
+                        modelConfig.environment,
```
minzcmu, line 198: We should add `[config.environment]` in the js doc for the create function
https://github.com/screwdriver-cd/models/blob/f58bd1a726e78ac46a331fb754fd49642598cf1b/lib/buildFactory.js#L150

minzcmu (approved):
except for the doc, LGTM 😀

lib/buildFactory.js
jithine: Added js doc.

test/lib/buildFactory.test.js
tkyi: so it shouldn't matter where it is

PR #239, opened 2018-04-26
[3] feat: Support periodic builds

d2lam:
Support periodic builds

Related: https://github.com/screwdriver-cd/screwdriver/issues/688

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -177,14 +177,14 @@ class Job extends BaseModel {
         const pipelineFactory = PipelineFactory.getInstance();
 
         return super.update()
-            // Update for periodic job
             .then(newJob => pipelineFactory.get(newJob.pipelineId)
                 .then(pipeline => this[executor].startPeriodic({
                     pipeline,
                     job: newJob,
                     tokenGen: this[tokenGen],
                     isUpdate: true
-                })));
+                }))
+                .then(() => this));
     }
 
     /**
@@ -206,8 +206,16 @@ class Job extends BaseModel {
 
         // Remove builds
         return removeBuilds()
-            // Remove periodic job
-            .then(() => this[executor].stopPeriodic({ jobId: this.id }))
+            .then(() => {
+                // Remove periodic job
+                const annotations = this.permutations[0].annotations;
+
+                if (annotations && annotations['beta.screwdriver.cd/buildPeriodically']) {
+                    return this[executor].stopPeriodic({ jobId: this.id });
+                }
+
+                return null;
+            })
             .then(() => super.remove());
     }
 }
--- a/lib/job.js
+++ b/lib/job.js
@@ -2,6 +2,7 @@
 
 const BaseModel = require('./base');
 const hoek = require('hoek');
+const getAnnotations = require('./helper').getAnnotations;
 const PAGINATE_PAGE = 1;
 const PAGINATE_COUNT = 50;
 const START_INDEX = 3;
@@ -208,9 +209,7 @@ class Job extends BaseModel {
         return removeBuilds()
             .then(() => {
                 // Remove periodic job
-                const annotations = this.permutations[0].annotations;
-
-                if (annotations && annotations['beta.screwdriver.cd/buildPeriodically']) {
+                if (getAnnotations(this.permutations[0], 'screwdriver.cd/buildPeriodically')) {
                     return this[executor].stopPeriodic({ jobId: this.id });
                 }
 
--- a/lib/job.js
+++ b/lib/job.js
@@ -5,6 +5,8 @@ const hoek = require('hoek');
 const PAGINATE_PAGE = 1;
 const PAGINATE_COUNT = 50;
 const START_INDEX = 3;
+const executor = Symbol('executor');
+const tokenGen = Symbol('tokenGen');
 
 class Job extends BaseModel {
     /**
@@ -12,11 +14,15 @@ class Job extends BaseModel {
      * @method constructor
      * @param  {Object}     config
      * @param  {Datastore}  config.datastore    Datastore instance
+     * @param  {Object}     config.executor     Object that will perform executor operations
+     * @param  {String}     config.tokenGen     Generator for tokens
      * See model schema
      * @constructor
      */
     constructor(config) {
         super('job', config);
+        this[executor] = config.executor;
+        this[tokenGen] = config.tokenGen;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


Filbird (approved):

lib/jobFactory.js
```diff
@@ -47,7 +57,21 @@ class JobFactory extends BaseFactory {
         c.state = 'ENABLED';
         c.archived = false;
 
-        return super.create(c);
+        return super.create(c).then((job) => {
+            if (!c.permutations ||
+                !c.permutations[0].annotations ||
+                !c.permutations[0].annotations['beta.screwdriver.cd/buildPeriodically']) {
```
pranavrc: Can probably use something like `if (!hoek.reach(c, 'permutations[0].annotations['beta.screwdriver.cd/buildPeriodically']'))` here. Not sure of the exact syntax but you get the idea

d2lam: That doesn't work on array, since `permutations` is an array. 
I can try to simplify it by assuming perm[0] is there.

d2lam: Ah, I remember why I didn't do this in the first place. Even if assuming `permutations[0]` is there, the `hoek.reach` will try to do a reach for `cd` inside `screwdriver` inside `beta` instead of interpreting the whole thing `beta.screwdriver.cd` as a `string`.

lib/jobFactory.js
```diff
@@ -46,8 +56,22 @@ class JobFactory extends BaseFactory {
 
         c.state = 'ENABLED';
         c.archived = false;
+        const perm = c.permutations[0];
+
+        return super.create(c).then((job) => {
+            if (!perm.annotations ||
+                !perm.annotations['beta.screwdriver.cd/buildPeriodically']) {
```
jithine: can we remove `beta` prefix for annotation ?

pranavrc (approved):
image

lib/job.js
```diff
@@ -157,6 +163,30 @@ class Job extends BaseModel {
...
+        // Lazy load factory dependency to prevent circular dependency issues
+        // https://nodejs.org/api/modules.html#modules_cycles
+        /* eslint-disable global-require */
+        const PipelineFactory = require('./pipelineFactory');
+        /* eslint-enable global-require */
+        const pipelineFactory = PipelineFactory.getInstance();
+
+        return super.update()
+            // Update for periodic job
+            .then(newJob => pipelineFactory.get(newJob.pipelineId)
```
jithine: shouldn't it check whether it's a periodic job prior to invoking `startPeriodic`

lib/job.js
```diff
@@ -174,8 +204,11 @@ class Job extends BaseModel {
                     .then(() => removeBuilds());
             }));
 
-        // Remove the job
-        return removeBuilds().then(() => super.remove());
+        // Remove builds
+        return removeBuilds()
+            // Remove periodic job
+            .then(() => this[executor].stopPeriodic({ jobId: this.id }))
```
jithine: shouldn't it check whether it's a periodic job prior to invoking `stopPeriodic`

jithine (changes requested):

lib/jobFactory.js
d2lam: We should keep it consistent with other features: https://docs.screwdriver.cd/user-guide/configuration/annotations

If we want to remove it, I would suggest to make an announcement that we are moving away from beta for ALL, not just this.

lib/job.js
d2lam: I was thinking the check should be done in `executor-queue`, since it can simply check if there is a record in redis, but I can also do an extra check here.

lib/job.js
d2lam: I was thinking the check should be done in `executor-queue`, since it can simply check if there is a record in redis, but I can also do an extra check here.

lib/job.js
minzcmu: We shouldn't do the check here, it's possible that user remove the periodic annotation in the new job. I think the check should be done in the `executor-queue` side since it has a record of the old job if it is a periodic build.

lib/jobFactory.js
jithine: Not necessarily, resources we are revisiting. If we had a common function to parse annotation name, then it would have been easier, since we are hardcoding annotation all over the place, change it now rather than later. 

I see no logic in calling this feature `beta` just because it's configuration is in `annotation`

jithine: This brings my second ask, hide this behind a utility function, which should work if `beta` is present or not. This way when we want to drop `beta` we we don't have to deal with hard coded annotation names.

lib/job.js
```diff
@@ -157,6 +163,30 @@ class Job extends BaseModel {
...
+
+        return super.update()
+            // Update for periodic job
+            .then(newJob => pipelineFactory.get(newJob.pipelineId)
+                .then(pipeline => this[executor].startPeriodic({
+                    pipeline,
+                    job: newJob,
+                    tokenGen: this[tokenGen],
+                    isUpdate: true
+                })));
```
minzcmu: We should still return the job object in the end for the update function, that's the expected behavior

minzcmu (changes requested):

lib/job.js
d2lam: Yup, I brought up this scenario in Pranav's PR. 
The only accurate way is to check if the old job has periodic annotations OR new job has periodic annotations. But I don't think it's possible since the attribute is already changed once it gets here.

lib/jobFactory.js
d2lam: The whole reason why we have `annotations` is because it's meant to be "experimental".
If we revisit, then we should revisit everything.

lib/job.js
```diff
@@ -174,8 +204,19 @@ class Job extends BaseModel {
...
 
-        // Remove the job
-        return removeBuilds().then(() => super.remove());
+        // Remove builds
+        return removeBuilds()
+            .then(() => {
+                // Remove periodic job
+                const annotations = this.permutations[0].annotations;
+
+                if (annotations && annotations['beta.screwdriver.cd/buildPeriodically']) {
```
jithine: can we have a helper function to parse annotation. `getAnnotation(job, name)`. Also drop `beta`

lib/jobFactory.js
```diff
@@ -46,8 +56,21 @@ class JobFactory extends BaseFactory {
 
         c.state = 'ENABLED';
         c.archived = false;
+        const annotations = c.permutations[0].annotations;
+
+        return super.create(c).then((job) => {
+            if (!annotations || !annotations['beta.screwdriver.cd/buildPeriodically']) {
```
jithine: use a helper function, this pattern is being repeated.

jithine (changes requested):

minzcmu (approved):

lib/jobFactory.js
d2lam: I can add it. The reason I didn't add it is because they are spread in 2 different files. This helper function will need a separate file, just for 1 line

jithine (approved):

PR #245, opened 2018-05-14
fix(1030): [3] Call tokenGen() method with temporal JWT expires time.

sakka2:
## Context
JWT is hardcoded to expire in 2 hours.
So, there is a case that JWT expires before the build timeout.
See this [issue #998](https://github.com/screwdriver-cd/screwdriver/issues/998).
See this [issue #1075](https://github.com/screwdriver-cd/screwdriver/issues/1075).

## Objective 
- When starting build, in build model, make temporal JWT and pass it to each executors. 
**temporal JWT will expire after 30 minutes.**
- Inside each executors, JWT for build will be generated using this temporal JWT via [JWT generate API](https://github.com/screwdriver-cd/screwdriver/pull/1087).

## References
- Issues: 
　https://github.com/screwdriver-cd/screwdriver/issues/998
　https://github.com/screwdriver-cd/screwdriver/issues/1030
　https://github.com/screwdriver-cd/screwdriver/issues/1075
- Other PRs:
　https://github.com/screwdriver-cd/screwdriver/pull/1087
　https://github.com/screwdriver-cd/models/pull/245
　https://github.com/screwdriver-cd/executor-base/pull/42
　https://github.com/screwdriver-cd/executor-k8s/pull/84
　https://github.com/screwdriver-cd/executor-k8s-vm/pull/33
　https://github.com/screwdriver-cd/executor-docker/pull/25
　https://github.com/screwdriver-cd/executor-jenkins/pull/20


wdstar (approved):
LGTM

lib/build.js
```diff
@@ -152,24 +154,35 @@ class BuildModel extends BaseModel {
...
         // Make sure that a pipeline and job is associated with the build
         return this.job
-            .then(job =>
-                job.pipeline.then(pipeline =>
+            .then((job) => {
+                let buildTimeout = DEFAULT_BUILD_TIMEOUT;
+
+                if (this[executor].kubernetes && this[executor].kubernetes.buildTimeout) {
+                    buildTimeout = this[executor].kubernetes.buildTimeout * 60;
+                }
```
sakka2: If buildTimeout is set by cluster config, it will be used.
buildTimeout setting is only set in executor-k8s or executor-k8s-vm

s-yoshika (approved):
LGTM

lib/build.js
s-yoshika: It is better to put comment for that, if these lines for executor specific.

lib/build.js
```diff
@@ -8,6 +8,8 @@ const apiUri = Symbol('apiUri');
 const tokenGen = Symbol('tokenGen');
 const uiUri = Symbol('uiUri');
 const ABORT_CODE = 130; // 128 + SIGINT 2 (^C)
+const DEFAULT_BUILD_TIMEOUT = 90 * 60; // seconds
```
tk3fftk: Other default build timeout set as minutes [like this](https://github.com/screwdriver-cd/executor-k8s-vm/blob/b4a556659cc0ddc9d754bd060e0bfbf5b80ab5bd/index.js#L14) so I think "minutes" is better in this case.

lib/build.js
sakka2: I put a comment.

lib/build.js
sakka2: Thanks, I fixed!
Converts to second [here](https://github.com/screwdriver-cd/models/pull/245/files#diff-26b681bd4dac10f25d5293c7595c435bR168) instead.

tk3fftk (approved):

lib/build.js
```diff
@@ -152,24 +154,36 @@ class BuildModel extends BaseModel {
...
+                let buildTimeout = DEFAULT_BUILD_TIMEOUT;
+
+                // buildTimeout is only set in executor-k8s or executor-k8s-vm
+                if (this[executor].kubernetes && this[executor].kubernetes.buildTimeout) {
+                    buildTimeout = this[executor].kubernetes.buildTimeout;
+                }
+
+                const annotations = job.permutations[0].annotations;
+                const jwtExpiresSec = annotations[ANNOTATION_BUILD_TIMEOUT]
+                    ? annotations[ANNOTATION_BUILD_TIMEOUT] * 60
```
tkyi: Should there be a check here to make sure the `annotations[ANNOTATION_BUILD_TIMEOUT]` is less than the `MAX_BUILD_TIMEOUT` (https://github.com/screwdriver-cd/screwdriver/blob/master/config/custom-environment-variables.yaml#L172)? 

Similar to https://github.com/screwdriver-cd/executor-k8s-vm/blob/master/index.js#L204-L206

lib/build.js
```diff
@@ -152,24 +154,36 @@ class BuildModel extends BaseModel {
     start() {
         // Make sure that a pipeline and job is associated with the build
         return this.job
-            .then(job =>
-                job.pipeline.then(pipeline =>
+            .then((job) => {
+                let buildTimeout = DEFAULT_BUILD_TIMEOUT;
```
jithine: can you move this logic to a separate function `const builtTimeout = getBuildTimeout(this[executor])` thereby hiding the special treatment given to kubernetes behind it.

lib/build.js
```diff
@@ -152,24 +154,36 @@ class BuildModel extends BaseModel {
...
+                        annotations,
                         apiUri: this[apiUri],
                         buildId: this.id,
                         container: this.container,
                         token: this[tokenGen](this.id, {
                             isPR: job.isPR(),
                             jobId: job.id,
                             pipelineId: pipeline.id
-                        }, pipeline.scmContext)
+                        }, pipeline.scmContext, jwtExpiresSec)
```
jithine: need a unit test case update for checking this new value being passed.

jithine (changes requested):

lib/build.js
sakka2: This `jwtExpiresSec` value is only used in `tokenGen(generateToken)` method, and I added the check that it is less than `max build timeout` in `generateToken` method.

https://github.com/screwdriver-cd/screwdriver/pull/1062/files#diff-cc45542b37e76ced3e366e4aa20f512cR99

lib/build.js
sakka2: I added `getBuildTimeout()` method.

lib/build.js
sakka2: I added the unit test.

lib/build.js
tkyi: Okay I see, makes sense 👍

jithine (commented):
just to be clear on the flow, when we use `executor-queue` with `executory-k8s(-vm)` what will `this[executor]`  be ? Will it be `executor-queue` or `executory-k8s-vm` ? 
Otherwise changes LGTM

sakka2:
Oh, I see.  `this[executor]` will be `executor-queue`.
So, we have to generate new JWT  in `queue-worker` before `executor.start` ?
https://github.com/screwdriver-cd/queue-worker/blob/master/lib/jobs.js#L49

jithine:
@sakka2 
> Oh, I see. this[executor] will be executor-queue.
So, we have to generate new JWT in queue-worker before executor.start ?
https://github.com/screwdriver-cd/queue-worker/blob/master/lib/jobs.js#L49

This change will work for non queue executors. However you are correct, we need to have the logic of token generation at `queue-worker` as well, but the change is not trivial. `queue-worker` does not have access to `tokenGenerator` and it shouldn't (else we will have to expose jwt signing key to worker which we should not). 

One design suggestion we have around this is for builds in queue to have a temporary `jwt ` (scope limited only for token exchange) which queue-worker will exchange with API for the actual build `jwt`
That requires an API end point whose purpose is to exchange JWT tokens. 

cc @Zhongtang

sakka2:
@jithin1987

Thank you for the suggestion!
I will change like below. Could you review for this please?

### 1.  Add the new API endpoint to generate build JWT.
**request_URL:**  `/v4/builds/{id}/token`
**request_body:** 
{
    `"expiresSec": "3700"`
}
**rquestHeader:** Authorization: Bearer `{temporal JWT}`

**response_body:**
{
    `"jwt": xxxxxxxxxxxxxxxxx (build JWT)`
}

### 2. In every build, firstly generate temporal JWT.
In `buildFactory.tokenGen()` method, make changes to always generate temporal JWT.
JWT scope will be changed from `['build']` to `['temporal']`.
https://github.com/screwdriver-cd/screwdriver/pull/1062/files#diff-c945a46d13b34fcaff544d966cffcabaR128

### 3.  In each executor, generate build JWT with calling new API.
For each `executor.start()` method,  add the process to call the new API endpoint for getting build JWT using temporal JWT passed in, except for `executor-queue`.
In case `executor-queue`, set the temporal JWT to Redis table, 
and in `queue-worker`, `executor.start()` method such as executor-k8s and executor-k8s-vm will call the new API with this temporal JWT as I wrote above.
https://github.com/screwdriver-cd/queue-worker/blob/master/lib/jobs.js#L49

d2lam:
@sakka2  Your proposal makes sense to me 👍

jithine:
@sakka2 Let's track temporal JWT work at https://github.com/screwdriver-cd/screwdriver/issues/1075

I have few open questions

sakka2:
@jithin1987 
Thanks for creating issue!

tk3fftk (approved):
LGTM

lib/build.js
```diff
@@ -226,7 +227,7 @@ class BuildModel extends BaseModel {
                         isPR: job.isPR(),
                         jobId: job.id,
                         pipelineId: pipeline.id
-                    }, pipeline.scmContext) }))
+                    }, pipeline.scmContext, TEMPORAL_JWT_TIMEOUT) }))
```
jithine, line 233: this sets an upper limit on how long a build can be in queue. I worry 30 mins won't be high enough for following use cases.

1. When we have [freeze windows](https://github.com/screwdriver-cd/screwdriver/issues/675).
2. Blocked_by jobs. Say a deploy job is waiting on a test job.

`TEMPORAL_JWT_TIMEOUT` value should be high enough to address this.

jithine (changes requested):

lib/build.js
sakka2: @jithin1987 
I see. It should be long enough. The scope of this JWT is just limited, so It can be more long.
How long is appropriate?

jithine: @sakka2 I would go for the `MAX_EXPIRES_IN` https://github.com/screwdriver-cd/screwdriver/pull/1087/files#diff-cc45542b37e76ced3e366e4aa20f512cR19 We can revisit when we get to `freeze windows`

That brings another question, rather than redefining the constant here, can we *move* the timeout constants from api to here in models such that everything is defined in a central place.

lib/build.js
```diff
@@ -10,6 +10,7 @@ const apiUri = Symbol('apiUri');
 const tokenGen = Symbol('tokenGen');
 const uiUri = Symbol('uiUri');
 const ABORT_CODE = 130; // 128 + SIGINT 2 (^C)
+const TEMPORAL_JWT_TIMEOUT = 120 * 60; // 120 minutes in seconds
```
jithine: I thought this was going to be 12hrs so that it's inline with `MAX_EXPIRES_IN`

sakka2: Yes, you are right. I fixed to 12hours.

lib/build.js
sakka2: @jithin1987 
I deleted the strategy of `maxAge` for JWT defined in [API](https://github.com/screwdriver-cd/screwdriver/pull/1087/files#diff-cc45542b37e76ced3e366e4aa20f512cL129).
This is the optional verification that if expires time of JWT is over this value, it will be error.
So, if we can always set appropriate expires time to JWT, this verification will not be needed.

Now we can declare the timeout constants only in this model.
I set `TEMPORAL_JWT_TIMEOUT` to `12h` (or should be more long?) as a default so as to be enough for `Blocked by` jobs.

And then, when we get `freeze windows`, may be freezing time will be obtained from the annotations. So even if it is 2 or 3days, we can set these terms to the expires time of JWT instead of default expires time above.

If expires time of JWT is too long, it will not good in the security even if it is temporal JWT.
So, I added the check whether the status of build is `QUEUED` or not. Only the build in `QUEUED` status, temporal JWT can be exchanged to the build JWT. Thus even if temporal JWT is long, once build has started, this JWT will not be used anymore.

https://github.com/screwdriver-cd/screwdriver/pull/1087/files#diff-2451ab46f6bc04f9551f9f33c8a65d7cR36

jithine: 👍

jithine (approved):

jithine (approved):

catto (approved):

PR #247, opened 2018-05-16
[1] fix(854): Add function to get permitted admin

yoshwata:
## Context

This pr fixes the issue [854](https://github.com/screwdriver-cd/screwdriver/issues/854)

## Objective

I add the function to delete the admin not permitted, and return the admin permitted. `pipeline.admin` enables to use admin who are not delete or suspended from repository.

## References

https://github.com/screwdriver-cd/screwdriver/issues/854

## Related PRs
https://github.com/screwdriver-cd/executor-queue/pull/25
https://github.com/screwdriver-cd/models/pull/247
https://github.com/screwdriver-cd/screwdriver/pull/1061


lib/pipeline.js
```diff
@@ -499,6 +487,50 @@ class PipelineModel extends BaseModel {
         return admin;
     }
 
+    async getFirstAdmin() {
```
tkyi: missing jsdocs

lib/pipeline.js
```diff
@@ -499,6 +487,50 @@ class PipelineModel extends BaseModel {
...
+        /* eslint-disable no-restricted-syntax */
+        for (const username of Object.keys(this.admins)) {
+            try {
+                /* eslint-disable no-await-in-loop */
```
tkyi: can replace with `// eslint-disable-next-line no-await-in-loop` and delete line 496

test/lib/pipeline.test.js
```diff
@@ -761,25 +767,50 @@ describe('Pipeline Model', () => {
...
+    describe('getFirstAdmin', () => {
+        beforeEach(() => {
+            getUserPermissionMocks({ username: 'batman', push: false });
+            getUserPermissionMocks({ username: 'robin', push: true });
+        });
+
+        it('has an admin robin', () => {
+            const admin = pipeline.getFirstAdmin();
+
+            admin.then((realAdmin) => {
```
tk3fftk: need `return` for correct test.

lib/pipeline.js
```diff
@@ -499,6 +487,50 @@ class PipelineModel extends BaseModel {
...
+                });
+            } catch (e) {
+                /* eslint-disable no-console */
+                console.log(e);
+                /* eslint-disable no-console */
+            }
+        }
+        /* eslint-disable no-restricted-syntax */
+
+        return false;
```
tk3fftk: Is this line needed or Is `return false` enough?

lib/pipeline.js
yoshwata: I add jsdocs. Thank you!

lib/pipeline.js
yoshwata: I fixed. Thank you!

lib/pipeline.js
```diff
@@ -499,6 +487,54 @@ class PipelineModel extends BaseModel {
...
+                                    return reject();
+                                }
+
+                                return resolve(realUser);
+                            });
+                    });
+                });
+            } catch (e) {
+                /* eslint-disable no-console */
+                console.log(e);
```
tk3fftk: `undefined` may be printed if `return reject()` is called.
I think need more information for logging or delete this `console.log`.

lib/pipeline.js
yoshwata: This is for lint. 
```
error  Expected to return a value at the end of async method 'getFirstAdmin'  consistent-return
```
Do you have any ideas other than `false`?

lib/pipeline.js
yoshwata: Thank you. I tried to delete, but `No catch clause` and `Empty block statement` is prohibited. Thus I add log message.

test/lib/pipeline.test.js
yoshwata: Thank you. I fixed.

lib/pipeline.js
yoshwata: I add `delete` non admin user in `catch`, and I could delete the `console.log`. I thought this is better.

lib/pipeline.js
```diff
@@ -499,6 +487,50 @@ class PipelineModel extends BaseModel {
...
+     * This function deletes admins who does not have proper
+     * GitHub permission, and returns an proper admin.
+     * @method getFirstAdmin
+     * @return {Promise}
+     */
+    async getFirstAdmin() {
+        /* eslint-disable no-restricted-syntax */
+        for (const username of Object.keys(this.admins)) {
+            try {
+                /* eslint-disable-next-line no-await-in-loop */
```
tkyi: should be a `//` format comment for next line only
like https://github.com/screwdriver-cd/models/blob/master/lib/templateFactory.js#L65

lib/pipeline.js
```diff
@@ -499,6 +487,50 @@ class PipelineModel extends BaseModel {
...
+
+                                return resolve(realUser);
+                            });
+                    });
+                });
+            } catch (usernameNotAdmin) {
+                delete this.admins[usernameNotAdmin];
+            }
+        }
+        /* eslint-disable no-restricted-syntax */
```
tkyi: should be eslint-enable*

lib/pipeline.js
```diff
@@ -499,6 +487,50 @@ class PipelineModel extends BaseModel {
...
+                /* eslint-disable-next-line no-await-in-loop */
+                return await new Promise((resolve, reject) => {
+                    // Lazy load factory dependency to prevent circular dependency issues
+                    // https://nodejs.org/api/modules.html#modules_cycles
+                    /* eslint-disable global-require */
+                    const UserFactory = require('./userFactory');
+                    /* eslint-enable global-require */
+                    const factory = UserFactory.getInstance();
+
+                    const user = factory.get({
```
d2lam: would this need `await`?

lib/pipeline.js
d2lam: I think `null` would make more sense in this scenario.

yoshwata: I add `throw new Error('Pipeline has no admin')`. I think this make it easy to understand what happened.

lib/pipeline.js
yoshwata: `UserFactory.get` is [here](https://github.com/screwdriver-cd/models/blob/master/lib/userFactory.js#L76).  It uses `Promise` only `return` context, so I think `return` will be called after `Promise` is resolved. Thus `factory.get` synchronized  without `await`.

lib/pipeline.js
yoshwata: I fixed. Thank you!

lib/pipeline.js
yoshwata: I fixed. Thank you!

tkyi (approved):
👍 LGTM

lib/pipeline.js
```diff
@@ -471,22 +471,10 @@ class PipelineModel extends BaseModel {
...
-        /* eslint-disable global-require */
-        const UserFactory = require('./userFactory');
-        /* eslint-enable global-require */
-        const factory = UserFactory.getInstance();
-
-        const admin = factory.get({
-            username: Object.keys(this.admins)[0],
-            scmContext: this.scmContext
-        });
+        const admin = this.getFirstAdmin();
```
minzcmu, line 477: Why not just `return this.getFirstAdmin()`?

lib/pipeline.js
```diff
@@ -499,6 +487,50 @@ class PipelineModel extends BaseModel {
...
+     * GitHub permission, and returns an proper admin.
+     * @method getFirstAdmin
+     * @return {Promise}
+     */
+    async getFirstAdmin() {
+        /* eslint-disable no-restricted-syntax */
+        for (const username of Object.keys(this.admins)) {
+            try {
+                // eslint-disable-next-line no-await-in-loop
+                return await new Promise((resolve, reject) => {
```
minzcmu: I am very confused by the code here since it's a mix of promise and async/await. If this is an async function, can we do something like this by using only async/await? 
```
try { 
    const user = await factory.get({});
    const permission =  await user.getPermissions();
    
     if (hasPermission....) {
          return ...
     } 
     
      throw err ....
} catch (err) { 
}
```

minzcmu (changes requested):

lib/pipeline.js
yoshwata: It is because `admin` is used [here](https://github.com/sonic-screwdriver-cd/models/blob/1e4da9ca7e82ec42f747499ecb488430c778a551/lib/pipeline.js#L484). I think it's for using  `admin` as property.

lib/pipeline.js
yoshwata: Thank you. It seems to work properly, and the code is nice. I fixed.

lib/pipeline.js
yoshwata: I'm sorry, I was wrong. It should be add `await` here. Thank you for comment.

lib/pipeline.js
```diff
@@ -499,6 +487,46 @@ class PipelineModel extends BaseModel {
...
+
+                // eslint-disable-next-line no-await-in-loop
+                const permission = await user.getPermissions(this.scmUri);
+
+                if (permission.push) {
+                    return user;
+                }
+
+                throw username;
+            } catch (usernameNotAdmin) {
```
jithine: why do you need `try - catch` here ? Instead of throwing in above line just do `delete this.admins[usernameNotAdmin]` there.

jithine (commented):
Does this sync back with git ? If new admins are added to git repo, is that handled ?

yoshwata:
@jithin1987 
https://github.com/screwdriver-cd/models/pull/247#pullrequestreview-123038350
> Does this sync back with git ? If new admins are added to git repo, is that handled ?

I think it is handled in another PR. New admins is added at [here](https://github.com/screwdriver-cd/screwdriver/pull/1061/files#diff-c235b867dd7760ebecef690395deec8bR57) before sync.

lib/pipeline.js
yoshwata: `try - catch` is used to catch application or system error, so it might be not needed for this situation. I removed `try - catch`. Thank you!

minzcmu (approved):

jithine (approved):
👍

lib/pipeline.js
```diff
@@ -499,6 +487,42 @@ class PipelineModel extends BaseModel {
...
 
+    /**
+     * This function deletes admins who does not have proper
+     * GitHub permission, and returns an proper admin.
+     * @method getFirstAdmin
+     * @return {Promise}
+     */
+    async getFirstAdmin() {
+        /* eslint-disable no-restricted-syntax */
+        for (const username of Object.keys(this.admins)) {
```
jithine, line 498: can this be made an array instead of object so that we can maintain order of admins. With features like ([External SD Config](https://github.com/screwdriver-cd/screwdriver/issues/937), person who created first might have a wider permissions in other repositories which new admins might not have. So trying to see if we can have order.

lib/pipeline.js
minzcmu: since `this.getFirstAdmin()` is an async function, do we need `await` here?

minzcmu (changes requested):

lib/pipeline.js
yoshwata: The idea that the order of admins related to width of permission is not reasonable.
That is defined too implicitly, so anyone can't notice the admin hierarchy.

Pipeline's admin will get fuzzy in the context of external config.

It is just an idea, but solusions of this problem are
1. Add new model for scmList, and each scm repository, not pipline, has admins.
2. Admin should have push permission of all repositories of `scmList`, so admin will be removed if any of the permission of `scmList` goes off. (If we keep using pipeline's admin)

jithine: AFAIK `admins` is an object of form `{adminName: true}`, and we are not making use of the object to store anything other than `true` so why not just make it an array ? However not a priority.

I agree this can't bring a solution to `scmList` auth concerns. And it seems that we might get insertion order for free http://2ality.com/2015/10/property-traversal-order-es6.html, assuming `admins` object is constructed similarly each time it's instantiated.

lib/pipeline.js
yoshwata: @minz1027 
> since this.getFirstAdmin() is an async function, do we need await here?

Only `async` function can have `await`, and `getter` seems not to be able to have `async`.

It might not need to add await in `this.getFirstAdmin()`.

If `pipepline.admin` is called from `async` function, I await like [here](https://github.com/screwdriver-cd/executor-queue/pull/25/files#diff-168726dbe96b3ce427e7fedce31bb0bcR166). If `pipeline.admin` is in NOT `async` function, `admin` is referred in `then` context like [here](https://github.com/screwdriver-cd/screwdriver/pull/1061/files#diff-7a514e440aa937beee722c7ac84dfb93R224).

yoshwata:
@minz1027 
> since this.getFirstAdmin() is an async function, do we need await here?

Only `async` function can have `await`, and `getter` seems not to be able to have `async`.

It might not need to add await in `this.getFirstAdmin()`.

If `pipepline.admin` is called from `async` function, I await like [here](https://github.com/screwdriver-cd/executor-queue/pull/25/files#diff-168726dbe96b3ce427e7fedce31bb0bcR166). If `pipeline.admin` is in NOT `async` function, `admin` is referred in `then` context like [here](https://github.com/screwdriver-cd/screwdriver/pull/1061/files#diff-7a514e440aa937beee722c7ac84dfb93R224).

minzcmu (approved):

minzcmu:
@yoshwata Thanks for the explanation! good to merge! 👍

PR #249, opened 2018-05-19
feat(926): Handle template namespace parsing [2]

tkyi:
## Context
We would like to introduce the concept of explicit namespaces into templates. People have historically been naming their templates with implicit namespaces by providing a template name with a slash in it. Since we want to be backwards compatible, we need to be able to parse the case where `namespace/name@version` is passed in along with the old format `name@version`, where `name` maybe contain `/`.

## Objective
This PR adds a `_getNameAndNamespace` method to first see if a template in that namespace exists. If it exists, it will try to return a template with that namespace. If no template exists with that namespace, it will revert to the old method of getting templates. This method will be called in all gets, creates, and lists for templates and templateTags.

Note: not sure if there's a better name for the methods: `_getNameAndNamespace` and `_getFullNameAndVersion `

## Related links
Blocked by https://github.com/screwdriver-cd/data-schema/pull/237
Related to https://github.com/screwdriver-cd/screwdriver/issues/926


minzcmu (approved):

Filbird (approved):

d2lam:
Hmmm I've stopped committing `package-lock.json`. Last time Tian checked, it does not do us any good. I think we can remove it, and add it to `.gitignore`

d2lam (approved):
small comment

lib/templateFactory.js
```diff
@@ -73,20 +136,74 @@ class TemplateFactory extends BaseFactory {
...
+     * @param  {Object}   config.params           Parameters to filter on
+     * @param  {Object}   config.paginate         Pagination parameters
+     * @param  {Number}   config.paginate.count   Number of items per page
+     * @param  {Number}   config.paginate.page    Specific page of the set to return
+     * @return {Promise}
+     */
+    list(config) {
+        if (config.params && config.params.name && !config.params.namespace) {
+            // eslint-disable-next-line no-underscore-dangle
+            return this._getNameAndNamespace(config.params.name).then((parsedTemplateName) => {
```
d2lam, line 152: Kinda weird we are doing the same `list` inside `_getNameAndNamespace` then again do another `list` in this function

tkyi: Yeah I'm not sure if there's a better way around this

PR #252, opened 2018-05-30
fix(653): Handle ~jobName for blockedBy

tkyi:
## Context
Since blockedBy is always using OR logic, we should at least make it consistent with previous workflow syntax by requiring a tilde before all the job names.

## Objective
This PR handles `blockedBy` when jobNames start with a tilde (`~`).

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/653
Blocked by https://github.com/screwdriver-cd/data-schema/pull/242

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -33,8 +33,8 @@ function findIdsOfMatchedJobs(jobs, jobNames) {
  * @return {Promise}             Array of blockedby JobIds
  */
 function getBlockedByIds(pipeline, job) {
-    const blockedByNames = job.permutations[0].blockedBy;
-    let blockedByIds = [job.id]; // always blocked by itself
+    let blockedByNames = job.permutations[0].blockedBy;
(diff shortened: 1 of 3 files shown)


lib/build.js
```diff
@@ -40,14 +40,18 @@ function getBlockedByIds(pipeline, job) {
         return Promise.resolve(blockedByIds);
     }
 
+    const externalBlockedByNames = blockedByNames.filter(
+        name => name.startsWith('~sd@'));
+
+    for (let i = 0; i < blockedByNames.length; i += 1) {
```
jithine: `blockedByNames` still contains `~sd@` entries, however `findIdsOfMatchedJobs` filters it out.

So this block can be reduced to `blockedByNames = blockedByNames.map(name => name.replace('~', ''));`

minzcmu (approved):

jithine (approved):

PR #254, opened 2018-05-31
fix(926): set namespace to null when not exist for get

minzcmu:
This is a problem introduced by template namespace feature:

`getTemplate` is broken for old templates with implicit namespace. When we call `super.get({ name: 'foo/bar', version: '1.0.0' })` , it will add `namespace: undefined` to the params inside `baseFactory.get` since namespace is now part of unique keys. We're assuming a `get` should always contain all unique keys. 
https://github.com/screwdriver-cd/models/blob/master/lib/baseFactory.js#L75

To be able to get old templates with name like `foo/bar`, we need to set namespace explicitly to `null` when fetching it.

### DIFF
--- a/lib/templateFactory.js
+++ b/lib/templateFactory.js
@@ -196,10 +196,9 @@ class TemplateFactory extends BaseFactory {
             return this._getNameAndNamespace(config.name).then((parsedTemplateName) => {
                 const { namespace, name } = parsedTemplateName;
 
-                if (namespace) {
-                    config.namespace = namespace;
-                    config.name = name;
-                }
+                // Always need to pass in namespace since it's a unique key
+                config.namespace = namespace || null;
+                config.name = name;
 
                 return super.get(config);
             });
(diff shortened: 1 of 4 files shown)


tkyi (changes requested):
Need to update the logic in the templateTagFactory to match as well: https://github.com/screwdriver-cd/models/blob/master/lib/templateTagFactory.js#L105-L108

lib/templateFactory.js
```diff
@@ -196,10 +196,9 @@ class TemplateFactory extends BaseFactory {
             return this._getNameAndNamespace(config.name).then((parsedTemplateName) => {
                 const { namespace, name } = parsedTemplateName;
 
-                if (namespace) {
-                    config.namespace = namespace;
-                    config.name = name;
-                }
+                // Always need to pass in namespace since it's a unique key
```
tkyi, line 199: 👍

tkyi (approved):
👍 Thanks, Min!

jithine (approved):

PR #258, opened 2018-06-01
[DNM] fix(926): Get all template namespaces

tkyi:
## Context
It would be nice to have an API endpoint `GET /templates/namespaces` to return all namespaces.

## Objective
Adding a templateFactory method to get all namespaces so we can have an API endpoint to get all namespaces.

_Note: Not sure what other metadata we would like to return._

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/926

### DIFF
--- a/README.md
+++ b/README.md
@@ -825,6 +825,14 @@ factory.getTemplate(fullTemplateName).then(model => {
 | :-------------   | :----- | :-------------|  :-------------|
 | fullTemplateName | String | Yes | Name of the template and the version or tag (e.g. chef/publish@1.2.3 or chef/publish@latest). Can also be just name of the template (e.g. chef/publish) |
 
+#### Get Namespaces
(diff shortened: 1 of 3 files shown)


jithine:
We need to think about how we want to render the templates page in UI. Today we show all the templates in the page, but when we group based on namespace, should we display all templates grouped by namespace or just show top namespaces and their templates by default and have rest collapsed by default with search functionality.

tkyi:
I was thinking a Netflix-like format where namespaces are the categories and templates are the movies. We could order the namespaces alphabetically or by popularity. Clicking a namespace would take you to a larger page with only templates with that namespace.

We could add additional namespaces such as "My Templates" and "Featured".

tkyi:
Closing this PR until we get a mock from the design team so we can better plan.

PR #260, opened 2018-06-05
fix(854) [1]: fix getFirstAdmin to update pipline.admins table

yoshwata:
## Context

This pr fixes the issue [854](https://github.com/screwdriver-cd/screwdriver/issues/854)

## Objective

The `pipeline.admin`, which is a getter added in the pr https://github.com/screwdriver-cd/models/pull/247, does not update db, so it does not delete admin who does not have sufficient permission. This PR fix the problem.

## References

https://github.com/screwdriver-cd/screwdriver/issues/854

## Related PRs
https://github.com/screwdriver-cd/models/pull/260
https://github.com/screwdriver-cd/screwdriver/pull/1106

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -714,7 +714,6 @@ class PipelineModel extends BaseModel {
         } else {
             // This is needed to make admins dirty and update db
             this.admins = newAdmins;
-            this.update();
 
             const result = await factory.get({
                 username: Object.keys(this.admins)[0],
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -7,6 +7,7 @@ const BaseModel = require('./base');
 const parser = require('screwdriver-config-parser');
 const workflowParser = require('screwdriver-workflow-parser');
 const hoek = require('hoek');
+const _ = require('lodash');
 const { PR_JOB_NAME, EXTERNAL_TRIGGER } = require('screwdriver-data-schema').config.regex;
 const PAGINATE_PAGE = 1;
 const PAGINATE_COUNT = 50;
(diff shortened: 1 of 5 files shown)


lib/pipeline.js
```diff
@@ -701,15 +703,26 @@ class PipelineModel extends BaseModel {
             // eslint-disable-next-line no-await-in-loop
             const permission = await user.getPermissions(this.scmUri);
 
-            if (permission.push) {
-                return user;
+            if (!permission.push) {
+                delete newAdmins[username];
```
yoshwata, line 710: All admins who doesn't have permission will removed here. The [reference of executor queue](https://github.com/screwdriver-cd/executor-queue/pull/28/files) is no problem because all users in admins table have proper permission.

minzcmu (approved):
👍 Please rebase with master before merging it

lib/pipeline.js
```diff
@@ -701,15 +703,26 @@ class PipelineModel extends BaseModel {
...
         }
         /* eslint-enable no-restricted-syntax */
 
-        throw new Error('Pipeline has no admin');
+        if (Object.keys(newAdmins).length === 0) {
+            throw new Error('Pipeline has no admin');
+        } else {
+            // This is needed to make admins dirty and update db
+            this.admins = newAdmins;
+            this.update();
```
minzcmu: Just a question to make sure this is what we want. So we don't want to wait for this update to finish before we return?

yoshwata: Thank you! I think I don't need to `update` at here. This function is called inside of `update`

https://github.com/screwdriver-cd/models/blob/master/lib/pipeline.js#L908

kumada626 (approved):
LGTM

tk3fftk (approved):

PR #261, opened 2018-06-06
feat(937): Pass config pipeline's secrets to child pipeline

Filbird:
## Context
Since child pipelines inherit the configuration of their config pipeline, they should also inherit their config pipeline's secrets.

## Objective
`pipeline.secrets` should return a Promise that includes the config pipeline's secrets if the pipeline is a child pipeline.

In the case that secrets between the two pipeline's share the same name, the child pipeline's secrets should take precedence.

## Reference
https://github.com/screwdriver-cd/screwdriver/issues/937

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -7,6 +7,7 @@ const BaseModel = require('./base');
 const parser = require('screwdriver-config-parser');
 const workflowParser = require('screwdriver-workflow-parser');
 const hoek = require('hoek');
+const _ = require('lodash');
 const { PR_JOB_NAME, EXTERNAL_TRIGGER } = require('screwdriver-data-schema').config.regex;
 const PAGINATE_PAGE = 1;
 const PAGINATE_COUNT = 50;
@@ -799,11 +800,33 @@ class PipelineModel extends BaseModel {
         /* eslint-enable global-require */
         const factory = SecretFactory.getInstance();
 
-        const secrets = factory.list(listConfig);
+        let secrets = factory.list(listConfig);
 
-        // ES6 has weird getters and setters in classes,
-        // so we redefine the pipeline property here to resolve to the
-        // resulting promise and not try to recreate the factory, etc.
+        // Fetch and merge config pipeline's secrets
+        if (this.configPipelineId && this.configPipelineId !== this.id) {
+            const configPipelineListConfig = hoek.clone(listConfig);
+
+            configPipelineListConfig.params.pipelineId = this.configPipelineId;
+
+            return Promise.all([
+                secrets,
+                factory.list(configPipelineListConfig)
+            ]).then((results) => {
+                // Merge config pipeline's secrets into this pipeline's
+                secrets = _.uniqBy([...results[0], ...results[1]], 'name');
+
+                // ES6 has weird getters and setters in classes,
+                // so we redefine the pipeline property here to resolve to the
+                // resulting promise and not try to recreate the factory, etc.
+                Object.defineProperty(this, 'secrets', {
+                    enumerable: true,
+                    value: secrets
+                });
+
+                return secrets;
+            });
+        }
+
         Object.defineProperty(this, 'secrets', {
             enumerable: true,
             value: secrets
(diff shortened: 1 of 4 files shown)


lib/pipeline.js
```diff
@@ -799,11 +800,33 @@ class PipelineModel extends BaseModel {
...
+            const configPipelineListConfig = hoek.clone(listConfig);
+
+            configPipelineListConfig.params.pipelineId = this.configPipelineId;
+
+            return Promise.all([
+                secrets,
+                factory.list(configPipelineListConfig)
+            ]).then((results) => {
+                // Merge config pipeline's secrets into this pipeline's
+                secrets = _.uniqBy([...results[0], ...results[1]], 'name');
```
minzcmu, line 816: Don't think uniqBy is what we want based on this example:
```
_.uniqBy([{ 'x': 1 }, { 'x': 2 }, { 'x': 1 }], 'x');
// => [{ 'x': 1 }, { 'x': 2 }]
```
https://lodash.com/docs#uniqBy
Can probably use `hoek.applyToDefaults` instead. https://github.com/hapijs/hoek/blob/master/API.md#applytodefaultsdefaults-options-isnulloverride

test/lib/pipeline.test.js
```diff
@@ -1037,6 +1037,56 @@ describe('Pipeline Model', () => {
...
+                },
+                paginate
+            };
+
+            secretFactoryMock.list.resolves([
+                [childPipelineSecrets],
+                [configPipelineSecrets]
+            ]);
+            // when we fetch secrets it resolves to a promise
+            assert.isFunction(childPipeline.secrets.then);
```
minzcmu: In the test, can we assert that the values we're getting is what we expect? not sure why other getter tests are structured like this, but can we try something like `return pipeline.secrets.then((secrets) => assert the value)`

minzcmu (changes requested):

lib/pipeline.js
```diff
@@ -799,11 +800,33 @@ class PipelineModel extends BaseModel {
...
+        let secrets = factory.list(listConfig);
+
+        // Fetch and merge config pipeline's secrets
+        if (this.configPipelineId && this.configPipelineId !== this.id) {
+            const configPipelineListConfig = hoek.clone(listConfig);
+
+            configPipelineListConfig.params.pipelineId = this.configPipelineId;
+
+            return Promise.all([
+                secrets,
```
tkyi, line 812: can probably delete line 803 and just put ~secrets~`factory.list(listConfig),` here

Filbird: Sadly, in the case of regular pipelines we will still need to return `secrets` on line `835`

tkyi: Ah I missed that. Nvm!

lib/pipeline.js
Filbird: I may be wrong, but if the goal is to only override any secrets with the same name, then I think `_.uniqBy` is the more appropriate approach.

For example, given these arrays of secret objects:
```
const list1 = [{name: 'secret1'}, {name: 'secret2'}];
const list2 = [{name: 'secret3'}, {name: 'secret4'}];
```

We expect the resulting list to include all of the secret objects since they all have unique names.

Output of `_.uniqBy([...list1, ...list2], 'name');`:
```
[ { name: 'secret1' },
  { name: 'secret2' },
  { name: 'secret3' },
  { name: 'secret4' } ]
``` 

Output of `hoek.applyToDefaults(list1, list2);`:
```
[ { name: 'secret3' }, { name: 'secret4' } ]
```

minzcmu: loll I get the secrets model wrong! yep, _uniq is the right way to go!

minzcmu:
Don't know why the tests failed for token model... But this is good to go after tests  pass! 👍

minzcmu (approved):

Filbird:
Believe it's because of this change:
https://github.com/screwdriver-cd/data-schema/pull/250

Made a small fix to cover those test cases!

minzcmu:
Em... interesting... instead of you fixing the tests, they should fix the data-schema to make it backward compatible, otherwise all old tokens will be broken...
https://github.com/screwdriver-cd/data-schema/pull/250/files
@s-yoshika can you help fixing it? our functional tests are also failing for this... https://cd.screwdriver.cd/pipelines/1/builds/41213

s-yoshika:
Oh... sorry, I'll revert it.

PR #265, opened 2018-06-18
feat(1113): Parse meta from config on event creation

Filbird:
## Context
Events should be created with the meta specified in the config.

## Objective
Create an `event` with `meta` if the field is populated.

## Reference
https://github.com/screwdriver-cd/screwdriver/issues/1113

### DIFF
--- a/test/lib/buildFactory.test.js
+++ b/test/lib/buildFactory.test.js
@@ -190,6 +190,11 @@ describe('Build Factory', () => {
             }
         };
 
+        const meta = {
+            foo: 'bar',
+            one: 1
+        };
+
         let saveConfig;
 
         beforeEach(() => {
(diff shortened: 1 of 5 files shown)


minzcmu (approved):
tests?

Filbird:
:^)

d2lam (approved):

minzcmu:
Can you also add meta as an optional field to the jsDoc for buildFactory.create function? https://github.com/screwdriver-cd/models/blob/master/lib/buildFactory.js#L155
And probably set the default value `{}` there too. Just to be consistent

Filbird:
Ooh. Yep! Good catch

minzcmu (approved):

test/lib/buildFactory.test.js
```diff
@@ -133,7 +133,7 @@ describe('Build Factory', () => {
         });
     });
 
-    describe('create', () => {
```
jithine: remove `only`

jithine (changes requested):

jithine (approved):

PR #268, opened 2018-06-25
fix(1011): Always update repo name and branch

tkyi:
## Context
If a user changes their repo name, Screwdriver isn't currently updating the scm repo information stored and displaying it correctly in the pipeline page.

## Objective
This PR changes it so calling `update()` always updates the scmRepo information.

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/1011

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -975,22 +975,18 @@ class PipelineModel extends BaseModel {
      * @return {Promise}
      */
     update() {
-        if (this.isDirty('scmUri')) {
-            return this.admin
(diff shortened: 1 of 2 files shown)


minzcmu (approved):

d2lam (approved):

d2lam:
Actually the test case for this should be it will still call `decorateUrl` even if `scmUri` has not changed. We should do assertion in the current test.

d2lam (approved):

PR #271, opened 2018-06-27
fix(1152): update isDone for UNSTABLE

d2lam:
The problem is `UNSTABLE` is not considered `done` so `executor stop` will not be called and `running_key` will not be cleaned up. 

`UNSTABLE` can be `done` or `not done` depending on the `endTime`. Thank you @jithin1987  for the suggestion.

Related: https://github.com/screwdriver-cd/screwdriver/issues/1152


jithine (approved):
👍

Filbird (approved):

minzcmu (approved):

PR #273, opened 2018-07-03
fix: [small] inherit previous configPipelineSha

tk3fftk:
## Context
I found a bug of External Config feature.  The reproduce flow is below:
1. Create pipelines same as https://github.com/screwdriver-cd-test/external-config-example and https://github.com/screwdriver-cd-test/external-config-child1
2. Start a child pipeline (this called `event1`).
3. Restart from `event1` and create `event2`. `event1` has `configPipelineSha` so it should succeed.
4. Restart from `event2`. `event2` does not have `configPipelineSha` so it should fail.

image

## Objective
This PR fixed to inherit previous `configPipelineSha`.

## Reference
Related: https://github.com/screwdriver-cd/screwdriver/issues/937

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -323,11 +323,15 @@ class EventFactory extends BaseFactory {
             .then((p) => {
                 // Sync pipeline with the parentEvent sha and create jobs based on that sha
                 if (config.parentEventId) {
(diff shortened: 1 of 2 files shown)


kumada626 (approved):
LGTM 👍

minzcmu (approved):

PR #276, opened 2018-07-13
fix(1179): Fix namespace parsing in template tag create

tkyi:
## Context
Template tag creation is broken because the namespace is not getting created properly.

## Objective
This PR refactors name/namespace parsing in template creation and uses it in template tag creation.

## Related links
Issue: https://github.com/screwdriver-cd/screwdriver/issues/1179

### DIFF
--- a/lib/helper.js
+++ b/lib/helper.js
@@ -1,5 +1,8 @@
 'use strict';
 
+const schema = require('screwdriver-data-schema');
+const TEMPLATE_NAME_REGEX_WITH_NAMESPACE = schema.config.regex.FULL_TEMPLATE_NAME_WITH_NAMESPACE;
+
 /**
  * Get the value of the annotation that matches name
  * @method getAnnotations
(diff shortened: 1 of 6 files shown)


jithine (approved):
What will we do about all published templates where `namespace` is not populated in database ?

tkyi:
@jithin1987 Only template tags are currently missing namespace. Moving forward, all templates and template tags should have namespace.

lib/helper.js
```diff
@@ -11,6 +14,41 @@ function getAnnotations(perm, name) {
...
+
+            return {
+                namespace,
+                name
+            };
+        }
+
+        // Set namespace to default if no slash in name
+        return {
+            namespace: 'default'
```
d2lam, line 43: Since this is a helper function, might be useful for later to return an object with both `namespace` and `name` for consistency. That way we won't need to know that this should be used in combination with a `merge`.

d2lam (approved):
LGTM 👍

PR #277, opened 2018-07-20
fix(1161): Remove templates on pipeline remove

Filbird:
## Context
Templates are left in an "orphaned" state if their underlying pipelines are removed. They should be removed as soon as their pipeline is.

## Objective
On `pipeline.remove()` remove all templates belonging to the pipeline.

## Context
https://github.com/screwdriver-cd/screwdriver/issues/1161

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -714,6 +714,42 @@ class PipelineModel extends BaseModel {
     }
 
     /**
+     * Fetch all templates that belong to this pipeline
+     * @property  {Templates}
+     * @return    {Promise}   Resolves to an array of templates
+    */
+    get templates() {
+        const listConfig = {
+            params: {
+                pipelineId: this.id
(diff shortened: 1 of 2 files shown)


tkyi (approved):

petey:
What if someone is still using that template? Won't their builds just suddenly start breaking? Would it be better to immediately mark the template as deprecated and provide a warning to the user(s) that the template is not supported, and will be removed after (X) time? That way they have some time to look into alternatives before their builds just start breaking.

Filbird:
@petey I agree, this change has the potential to be unintentionally disruptive. 

I'll bring up your suggestion on the issue page so we can discuss this further.

Filbird:
Closed in favor of an alternative approach as discussed in https://github.com/screwdriver-cd/screwdriver/issues/1161.

PR #280, opened 2018-08-06
feat(1114): add step model

minzcmu:
### Context
To avoid race condition when updating step meta when build's running, we separate step into a different model. API will be modified to update the step model directly instead of the whole `steps` field of a build.

### Objectives
This PR contains the following change: 

- add step model
- create step models when creating new build
- `buildFactory.get` now returns a build with merged step data with step models

### Related links
This bug is related to changes made for https://github.com/screwdriver-cd/screwdriver/issues/1114

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -11,6 +11,8 @@ const tokenGen = Symbol('tokenGen');
 const uiUri = Symbol('uiUri');
 const ABORT_CODE = 130; // 128 + SIGINT 2 (^C)
 const TEMPORAL_JWT_TIMEOUT = 12 * 60; // 12 hours in minutes
+const PAGINATE_PAGE = 1;
+const PAGINATE_COUNT = 50;
 
 /**
  * Get the array of ids for jobs that match the names passed in
@@ -121,6 +123,33 @@ class BuildModel extends BaseModel {
 
             return this.scm.updateCommitStatus(config);
         });
+    }
+
+    /**
+    * Get models for all steps
+    * @method getStepsModel
+    * @return {Promise}
+    */
+    getStepsModel() {
+        const listConfig = {
+            params: {
+                buildId: this.id
+            },
+            paginate: {
+                count: PAGINATE_COUNT,
+                page: PAGINATE_PAGE
+            }
+        };
+
+        // Lazy load factory dependency to prevent circular dependency issues
+        // https://nodejs.org/api/modules.html#modules_cycles
+        /* eslint-disable global-require */
+        const StepFactory = require('./stepFactory');
+        /* eslint-enable global-require */
+
+        const factory = StepFactory.getInstance();
+
+        return factory.list(listConfig);
     }
 
     /**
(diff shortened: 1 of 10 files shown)


lib/build.js
```diff
@@ -11,6 +11,8 @@ const tokenGen = Symbol('tokenGen');
 const uiUri = Symbol('uiUri');
 const ABORT_CODE = 130; // 128 + SIGINT 2 (^C)
 const TEMPORAL_JWT_TIMEOUT = 12 * 60; // 12 hours in minutes
+const PAGINATE_PAGE = 1;
+const PAGINATE_COUNT = 50;
```
jithine, line 15: does this mean, by default `50` steps for a build will be fetched and caller has to handler if there are more ?

lib/build.js
```diff
@@ -249,21 +278,34 @@ class BuildModel extends BaseModel {
...
+                            step.endTime = now;
+                            step.code = ABORT_CODE;
+                        }
+
+                        return step.update();
+                    }));
                 }
 
-                return step;
+                this.steps = this.steps.map((step) => {
```
jithine, line 293: which case is this ? For older builds ? Does previous `this.getStepsModel()` need to be wrapped in `try-catch` ?

lib/buildFactory.js
```diff
@@ -226,6 +229,12 @@ class BuildFactory extends BaseFactory {
                         return super.create(modelConfig);
                     });
                 })
+                    .then(build => Promise.all(modelConfig.steps.map(step =>
```
jithine, line 231: do we still populate `steps` field for build model ? Is that required ?

lib/buildFactory.js
```diff
@@ -236,6 +245,31 @@ class BuildFactory extends BaseFactory {
...
+     * @method get
+     * @param  {Mixed}   config    The configuration from which an id is generated or the actual id
+     * @return {Promise}           Resolve build model after merging with step models
+     */
+    get(config) {
+        return super.get(config)
+            .then(build => build.getStepsModel()
+                .then((stepsModel) => {
+                    build.steps = build.steps.map((step) => {
+                        const stepModel = stepsModel.find(s => s.name === step.name);
```
jithine, line 258: If `stepModel` exists can't we do `build.steps = stepModel` ?

lib/build.js
minzcmu: Yes. But pagination is not working now, this is doing nothing. Even we don't pass it here, it has a default value in the base.

lib/buildFactory.js
```diff
@@ -169,8 +169,11 @@ class BuildFactory extends BaseFactory {
         // https://nodejs.org/api/modules.html#modules_cycles
         /* eslint-disable global-require */
         const JobFactory = require('./jobFactory');
+        const StepFactory = require('./stepFactory');
         /* eslint-enable global-require */
         const factory = JobFactory.getInstance();
+        const stepFactory = StepFactory.getInstance();
+        /* eslint-enable global-require */
```
tkyi: don't need this line since it's already on line 173

lib/build.js
minzcmu: This is for backward compatibility. `getStepsModel` is inside `abortSteps`.  The whole `abortSteps` function is in the promise chain of `update`. If anything fail will be caught at api.

lib/buildFactory.js
minzcmu: No, need to get every single model and do `model.toJson()` before returning it otherwise it will contain functions and other stuff. I added this in the latest commit.

lib/step.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+const BaseModel = require('./base');
+
+class StepModel extends BaseModel {
+    /**
+     * Construct a BannerModel object
```
tkyi: Need to replace all the "banner" docs with "step"

lib/buildFactory.js
minzcmu: don't really need it. i can delete it from modelConfig before we save the build. But there is no harm to keep it? If ppl try to get `build/{id}` before step model populated, they can still see the steps info.

tkyi: I think it would be good to leave it just in case.

tkyi (approved):
👍 LGTM

jithine (approved):

PR #282, opened 2018-08-09
feat(1204): Add pagination for getting pipeline events only [2]

tkyi:
## Context
The pipeline page is taking a long time to load since there are so many events per pipeline. Pagination is only actually done through the UI, not from the API side. It would be nice to do pagination from the API side to speed up UI loading for the pipeline page.

## Objective
This PR will:
- remove pagination for all function calls except `getEvents` (since most functions rely on everything being returned)
- remove default pagination from baseFactory except when `config.pagination` is passed in

2 questions:
1. Not sure if we should default to `sort: descending` (the API will force that default anyways using the data-schema (https://github.com/screwdriver-cd/data-schema/blob/master/api/pagination.js#L18))
2. Not sure if we should always paginate pipeline `getEvents` since *A. it's unlikely people will want more than 50 events* and *B. it'll make the page load a lot faster* or we should default to no pagination for backwards compatibility.

## Related links
Issue: https://github.com/screwdriver-cd/screwdriver/issues/1204

### DIFF
--- a/lib/baseFactory.js
+++ b/lib/baseFactory.js
@@ -116,15 +116,18 @@ class BaseFactory {
     list(config) {
         const scanConfig = {
             table: this.table,
-            params: config.params || {},
-            paginate: {
-                count: hoek.reach(config, 'paginate.count', { default: PAGINATE_COUNT }),
-                page: hoek.reach(config, 'paginate.page', { default: PAGINATE_PAGE })
-            }
+            params: config.params || {}
         };
 
         if (config.sort) {
             scanConfig.sort = config.sort;
+        }
+
+        if (config.paginate) {
+            scanConfig.paginate = {
+                count: hoek.reach(config, 'paginate.count', { default: PAGINATE_COUNT }),
+                page: hoek.reach(config, 'paginate.page', { default: PAGINATE_PAGE })
+            };
         }
 
         return this.datastore.scan(scanConfig)
--- a/lib/baseFactory.js
+++ b/lib/baseFactory.js
@@ -105,8 +105,8 @@ class BaseFactory {
     /**
      * List records with pagination and filter options
      * @method list
-     * @param  {Object}   config                    Config object
-     * @param  {Object}   config.params             Parameters to filter on
+     * @param  {Object}   [config]
+     * @param  {Object}   [config.params]           Parameters to filter on
      * @param  {Object}   [config.paginate]         Pagination parameters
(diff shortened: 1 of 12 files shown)


minzcmu (approved):

lib/pipeline.js
```diff
@@ -963,12 +952,16 @@ class PipelineModel extends BaseModel {
...
+        };
+
+        // Do not force pagination by default (for backwards compatibility)
+        if (config && config.paginate) {
+            defaultConfig.paginate = {
                 count: PAGINATE_COUNT,
                 page: PAGINATE_PAGE
-            }
-        };
+            };
```
minzcmu: Don't think we need this default pagination option here. Base factory will handle the default values.

minzcmu (changes requested):

minzcmu (approved):

lib/baseFactory.js
```diff
@@ -116,17 +116,20 @@ class BaseFactory {
...
         };
 
-        if (config.sort) {
+        if (config && config.sort) {
             scanConfig.sort = config.sort;
         }
 
+        if (config && config.paginate) {
+            scanConfig.paginate = {
+                count: hoek.reach(config, 'paginate.count', { default: PAGINATE_COUNT }),
```
minzcmu: This hoek.reach behaves weird when paginate.count = undefined.
```
> var config = { paginate: { count: undefined } }
>  hoek.reach(config, 'paginate.count', { default: 5 })
undefined
```
See it's not using the default value when paginate.count's value is `undefined`. We need to take care of that case as well.  need to do `hoek.reach(config, 'paginate.count', { default: PAGINATE_COUNT }) || PAGINATE_COUNT`. Not sure if there is any other way to do it to make it cleaner

minzcmu (changes requested):

minzcmu (approved):

Filbird (approved):

lib/baseFactory.js
```diff
@@ -116,17 +116,22 @@ class BaseFactory {
...
         };
 
-        if (config.sort) {
+        if (config && config.sort) {
             scanConfig.sort = config.sort;
         }
 
+        if (config && config.paginate) {
+            scanConfig.paginate = {
+                count: hoek.reach(config, 'paginate.count', { default: PAGINATE_COUNT })
```
d2lam, line 128: `hoek.reach` doesn't seem to be useful here. Can just do `paginate.count || PAGINATE_COUNT`

PR #283, opened 2018-08-14
fix: fix not to throw error when build not found

catto:
I found build API returns 500 error if specified build doesn't exist. I think it should return 404 in this case.
```javascript
{"statusCode ":500, "error": "Internal Server Error", "message": "Cannot read property 'getStepsModel' of null"}
```
And I fixed it by checking null and returning empty value in `BuildFactory.get`.
It is the same behavior as https://github.com/screwdriver-cd/models/blob/master/lib/baseFactory.js#L93-L96

### DIFF
--- a/lib/buildFactory.js
+++ b/lib/buildFactory.js
@@ -98,6 +98,10 @@ function dockerImageName({ container, dockerRegistry }) {
  * @return {Promise}                Build model with merged steps
  */
 function mergeStepsModel(build) {
+    if (!build) {
+        return Promise.resolve(build);
+    }
+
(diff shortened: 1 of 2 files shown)


jithine (dismissed):

jithine (commented):
We could make Promise to `reject` here and let caller handle this ?

minzcmu:
@jithin1987 API will check if the returned obj is undefined / null and throw 404

minzcmu (approved):

tkyi (approved):

PR #284, opened 2018-08-14
fix(1244): Pass correct parent pipeline sha to startBuild

Filbird:
## Context
Currently in `createBuild`, we pass in the value of `configPipelineSha` that belongs to `eventConfig`. In the case where a build is triggered via a webhook, this value will improperly be set to a sha belonging to the child pipeline rather than the intended value.

## Objective
Pass in the correct value of `configPipelineSha` to `startBuild`.

## References
https://github.com/screwdriver-cd/screwdriver/issues/1244


pranavrc (approved):
image

jithine (approved):

PR #285, opened 2018-08-28
fix(1220): Reduce for-loop in getFirstAdmin

kumada626:
## Context
I found the waste API calls in the getFirstAdmin logic. There is no need to find other admins after first admin is found.

## Objective
I fixed getFristAdmin to reduce scm API calls.

## References
screwdriver-cd/screwdriver#1220


wdstar (approved):
:+1:

catto (approved):

PR #286, opened 2018-08-28
[2][small] fix(1220):  Add scmInfo

kumada626:
## Context
Screwdriver calls `lookupScmUri` many times during a single PR event and it should be reduced.

## Objective
I added scmInfo property to the pipeline model to cache scmInfo. And use it to reduce API calls.

## References
original issue: screwdriver-cd/screwdriver#1220
related PRs:
screwdriver-cd/data-schema#273
screwdriver-cd/scm-github#101

### DIFF
--- a/test/lib/pipeline.test.js
+++ b/test/lib/pipeline.test.js
@@ -979,6 +979,18 @@ describe('Pipeline Model', () => {
             // We want to cache scmInfo to reduce API call
             assert.calledOnce(scmMock.lookupScmUri);
         });
+
+        it('rejects when lookupScmUri rejects', () => {
+            const testError = new Error('someGithubCommError');
+
+            scmMock.lookupScmUri.rejects(testError);
+
+            return pipeline.scmInfo.then(() => {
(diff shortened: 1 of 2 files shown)


test/lib/pipeline.test.js
```diff
@@ -943,6 +963,24 @@ describe('Pipeline Model', () => {
         });
     });
 
+    describe('get scmInfo', () => {
+        beforeEach(() => {
+            scmMock.lookupScmUri.resolves({
```
minzcmu, line 968: Can we have a test for the failure case e.g. when lookupScmUri rejects with err? In that case will the getter also throw err?

tkyi (approved):

test/lib/pipeline.test.js
kumada626: In that case, the getter should also throw err. I added the test case for the failure case when lookupScmUri rejects with err.

minzcmu (approved):

PR #293, opened 2018-10-05
fix(1271): Add option to return raw data for datastore scan [3]

tkyi:
## Context
When trying to get all distinct namespaces in templates, you should get an array of namespace strings, instead of an array of template objects.

## Objective
This PR changes the `list` function to take in a flag to return raw data.

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/1271, https://github.com/screwdriver-cd/datastore-sequelize/pull/35

### DIFF
--- a/lib/baseFactory.js
+++ b/lib/baseFactory.js
@@ -110,6 +110,7 @@ class BaseFactory {
      * @param  {Object}   [config.paginate]         Pagination parameters
      * @param  {Number}   [config.paginate.count]   Number of items per page
      * @param  {Number}   [config.paginate.page]    Specific page of the set to return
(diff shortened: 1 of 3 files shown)


lib/baseFactory.js
```diff
@@ -144,12 +145,17 @@ class BaseFactory {
             };
         }
 
+        if (config && config.raw) {
```
minzcmu, line 148: Let's document this behavior in the readme too

minzcmu (changes requested):

DekusDenial (approved):

minzcmu (approved):

PR #300, opened 2018-11-17
fix(1319): call list with params

minzcmu:
call list with params otherwise it won't work.

example: https://github.com/screwdriver-cd/models/blob/master/lib/templateTagFactory.js#L57


parthasl (approved):
👍

PR #301, opened 2018-11-27
[4] feat(1010): enable prChain

tk3fftk:
## Context
We want to implement `prChain` feature. If `prChain` is true, a job name does not have `PR-` prefix and will trigger subsequent jobs. 

## Objective
Pass to `prChain` to workflow-parser and store `prChain` to a pipeline.

## Misc.
Related Issue: https://github.com/screwdriver-cd/screwdriver/issues/1010
Related PRs:
- https://github.com/screwdriver-cd/data-schema/pull/296
- https://github.com/screwdriver-cd/config-parser/pull/81
- https://github.com/screwdriver-cd/workflow-parser/pull/15

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -91,9 +91,6 @@ function getJobsFromPR(config) {
         prChain
     });
 
-    console.log('===== nextJobs ======');
-    console.log(nextJobs);
-
     const jobsToStart = jobs.filter(
         j => j.name.startsWith(`PR-${prNum}:`) || nextJobs.includes(j.name)
     );
@@ -145,8 +142,6 @@ function startBuild(config) {
                 return isMatch;
             }));
     }
-
-    console.log(buildConfig);
 
     return hasChangeInSourcePaths ? buildFactory.create(buildConfig) : null;
 }
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -67,16 +67,17 @@ function getJobsFromJobName(config) {
 /**
  * Get PR jobs to start that are enabled
  * @method getJobsFromPR
- * @param  {Object}   config
- * @param  {Array}    config.jobs                           Array of job objects
- * @param  {Number}   config.prNum                          PR number
- * @param  {Object}   config.pipelineConfig
- * @param  {Object}   config.pipelineConfig.workflowGraph   Object with nodes and edges that represent the order of jobs
- * @param  {String}   config.startFrom                      Startfrom (e.g. ~commit, ~pr, etc)
- * @resolves {Array}                                        Array of PR jobs to start
+ * @param  {Object}    config
+ * @param  {Array}     config.jobs                           Array of job objects
+ * @param  {Number}    config.prNum                          PR number
+ * @param  {Object}    config.pipelineConfig
+ * @param  {Object}    config.pipelineConfig.workflowGraph   Object with nodes and edges that represent the order of jobs
+ * @param  {String}    config.startFrom                      Startfrom (e.g. ~commit, ~pr, etc)
+ * @param  {Boolean}   config.prChain                        Flag of triggering subsequent job after pull request job
+ * @resolves {Array}                                         Array of PR jobs to start
  */
 function getJobsFromPR(config) {
-    const { jobs, prNum, pipelineConfig, startFrom } = config;
+    const { jobs, prNum, pipelineConfig, startFrom, prChain } = config;
     const isPRTrigger = PR_TRIGGER.test(startFrom);
 
     if (!isPRTrigger) {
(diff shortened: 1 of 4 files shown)


lib/pipeline.js
```diff
@@ -342,17 +343,28 @@ class PipelineModel extends BaseModel {
...
-                const jobsToArchive = prJobs.filter(p => !nextJobs.includes(p.name));
-                const jobsToUnarchive = prJobs.filter(p => nextJobs.includes(p.name));
+                const existingPRJobNames = prJobs.map(prJob => prJob.name);
+                const missingPRJobNames =
+                    nextJobs.filter(nextJob => !existingPRJobNames.includes(nextJob));
+                let jobsToCreate;
+                let jobsToArchive;
+                let jobsToUnarchive;
+
+                if (this.prChain) {
```
wdstar, line 375: We should add logging of `prChain` flag for inspect, debugging and so on .

lib/pipeline.js
```diff
@@ -316,8 +317,8 @@ class PipelineModel extends BaseModel {
      * Update the permutations to be correct
      * Create missing PR jobs
      * @param   {Number}   prNum        PR Number
-     * @method syncPR
-     * @return {Promise}
+     * @method  syncPR
```
d2lam, line 336: I think we also need to change `syncPRs` method

d2lam:
I'm a little confused. Can I get some clarification? If `prChain` is true, then it will just run the actual pipeline job as if it was merged?

lib/pipeline.js
sakka2: I added logging for `prChain`.

lib/pipeline.js
sakka2: I fixed _checkPRState() method, not to be unarchived when sync button is pushed.
https://github.com/screwdriver-cd/models/pull/301/files#diff-679e90b5b15ab1b73ddb7ed2377e7413R192

sakka2:
@d2lam 
> If prChain is true, then it will just run the actual pipeline job as if it was merged?

That's right. If prChain is true, PR jobs will be just treated as commit job.

wdstar (approved):
LGTM

lib/pipeline.js
```diff
@@ -342,17 +343,29 @@ class PipelineModel extends BaseModel {
...
-                const jobsToArchive = prJobs.filter(p => !nextJobs.includes(p.name));
-                const jobsToUnarchive = prJobs.filter(p => nextJobs.includes(p.name));
+                const existingPRJobNames = prJobs.map(prJob => prJob.name);
+                const missingPRJobNames =
+                    nextJobs.filter(nextJob => !existingPRJobNames.includes(nextJob));
+                let jobsToCreate;
+                let jobsToArchive;
+                let jobsToUnarchive;
+
+                winston.info(`prChain flag is ${this.prChain}.`);
```
tk3fftk: we should add more information for logging like pipeline ID

sakka2: I added pipelineId to this info log.

lib/eventFactory.js
```diff
@@ -85,9 +87,13 @@ function getJobsFromPR(config) {
 
     const nextJobs = workflowParser.getNextJobs(pipelineConfig.workflowGraph, {
         trigger: startFrom,
-        prNum
+        prNum,
+        prChain
     });
 
+    console.log('===== nextJobs ======');
```
d2lam: Remove console logs

lib/pipeline.js
```diff
@@ -342,17 +343,29 @@ class PipelineModel extends BaseModel {
...
+                const existingPRJobNames = prJobs.map(prJob => prJob.name);
+                const missingPRJobNames =
+                    nextJobs.filter(nextJob => !existingPRJobNames.includes(nextJob));
+                let jobsToCreate;
+                let jobsToArchive;
+                let jobsToUnarchive;
+
+                winston.info(`pipelineId:${this.id}: prChain flag is ${this.prChain}.`);
+                if (this.prChain) {
+                    jobsToCreate = [];
```
d2lam, line 376: so we don't create new PR jobs? Just use the actual pipeline job?

sakka2: Yes, we don't create PR jobs if prChain is true.

lib/eventFactory.js
sakka2: Sorry, I removed.

lib/pipeline.js
tkyi: I think it's better to have separate jobs for PR workflow. If you run the actual pipeline job for PRs, it may modify configuration and code of real production jobs.

d2lam:
Discussed with @sakka2 offline. We need to create different jobs for different PRs for `prChain:true`. Otherwise, PR builds might modify actual job config. Also,  multiple PRs might conflict and override each other. 
However, since the flag is off right now. I'm ok with merging this and fix it in the second pass. It should not affect current build flow.

d2lam (approved):

PR #306, opened 2018-12-13
fix(1405): Add logging around child pipelines events

tkyi:
## Context
When child pipelines of a parent pipeline are getting removed it's almost impossible to track where and when it happened.

## Objective
This PR adds logging around child pipeline events and attempts to standardize log output from the pipeline factory.

example logs:
```
2018-12-14T01:14:31.407Z - info: pipelineId:123: Updating child pipeline foo.git with pipelineId:2.
2018-12-14T01:14:31.408Z - info: pipelineId:123: Removing child pipeline for baz.git with pipelineId:2.
2018-12-14T01:14:31.408Z - info: pipelineId:123: Creating child pipeline for bar.git with pipelineId:98765.
```

## Related links
https://github.com/screwdriver-cd/screwdriver/issues/1405

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -16,6 +16,11 @@ const REGEX_CAPTURING_GROUP = {
 };
 
 winston.level = process.env.LOG_LEVEL || 'info';
+const logger = new (winston.Logger)({
+    transports: [
+        new (winston.transports.Console)({ timestamp: true })
+    ]
+});
 
 /**
  * Sync external triggers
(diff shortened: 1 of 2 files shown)


lib/pipeline.js
```diff
@@ -504,15 +510,21 @@ class PipelineModel extends BaseModel {
...
-            // Child pipline does not belong to this parent, return
-            winston.error(`Pipeline ${scmUrl} already exists: ${pipeline.id}`);
+            // Child pipeline does not belong to this parent, return
+            logger.error(`pipelineId:${this.id}: Pipeline ${scmUrl} already ` +
+                `exists: ${pipeline.id}.`);
 
             return null;
         }
 
+        logger.info(`pipelineId:${this.id}: Creating child pipeline for ${scmUrl}.`);
```
d2lam: should we log this after the `create` so that we can include the child pipeline id in the log?

d2lam (approved):
Small comment. The rest LGTM 👍

lib/pipeline.js

PR #316, opened 2019-01-22
fix: fix no build event response

yokawara:
### Context
Events which doesn't trigger any jobs return error to github webhooks.
It would be better to return 204.

### Objective
* Return null when events starts no jobs.

### References
relates to:
* https://github.com/screwdriver-cd/screwdriver/pull/1445

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -208,6 +208,7 @@ function createBuilds(config) {
             // No jobs to start (eg: when jobs are disabled or startFrom is not valid)
             if (jobsToStart.length === 0) {
                 winston.info('No jobs to start.');
+
                 return null;
             }
 
@@ -231,6 +232,7 @@ function createBuilds(config) {
 
                 if (builds.length === 0) {
                     winston.info('No jobs to start.');
+
                     return null;
                 }
 
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -8,6 +8,7 @@ const {
     PR_TRIGGER
 } = require('screwdriver-data-schema').config.regex;
 const workflowParser = require('screwdriver-workflow-parser');
+const winston = require('winston');
 
 let instance;
 
@@ -206,6 +207,7 @@ function createBuilds(config) {
         .then((jobsToStart) => {
             // No jobs to start (eg: when jobs are disabled or startFrom is not valid)
             if (jobsToStart.length === 0) {
+                winston.info('No jobs to start.');
                 return null;
             }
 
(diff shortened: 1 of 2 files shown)


lib/eventFactory.js
```diff
@@ -206,7 +207,9 @@ function createBuilds(config) {
         .then((jobsToStart) => {
             // No jobs to start (eg: when jobs are disabled or startFrom is not valid)
             if (jobsToStart.length === 0) {
-                throw new Error('No jobs to start');
+                winston.info('No jobs to start.');
```
minzcmu: Can we log out the `eventId` together with this line? that will be easier to track later on

minzcmu (changes requested):

tkyi (approved):

lib/eventFactory.js
yokawara: @minz1027 
That makes sense, I added `eventId` in the log. thanks!

yokawara:
Could you restart PR job? 🙏

minzcmu:
@yokawara lol now you can restart the pr build, @tkyi implement a feature to allow PR author to restart their builds even if they don't have permissions against the repo

https://github.com/screwdriver-cd/screwdriver/issues/1246

yokawara:
@minz1027 I thought I could, but I can't sign in with my github account, so I can't restart the build...
It seems that I need the right to access your SD cluster.

lib/eventFactory.js
```diff
@@ -206,7 +207,9 @@ function createBuilds(config) {
         .then((jobsToStart) => {
             // No jobs to start (eg: when jobs are disabled or startFrom is not valid)
             if (jobsToStart.length === 0) {
-                throw new Error('No jobs to start');
+                winston.info(`No jobs to start in event ${eventId}.`);
```
d2lam, line 210: So what happens if all jobs are disabled and user clicks start? Previously they will see the error message. We should have provide some feedback to users still right?

yokawara: @d2lam You're right, thank you. Feedbacks should be provided to users as before.
So I fixed api to return message when events don't start any job.
https://github.com/screwdriver-cd/screwdriver/pull/1445/commits/d5e5c8850f143944f7a6342349d8fe63236e8786

yoshwata (approved):

minzcmu:
@yokawara @yoshwata added you guys to the whitelist for sd.cd, so that you guys can actually see your builds there!

minzcmu (approved):

minzcmu:
Good to go after fixing the test

lib/eventFactory.js
d2lam: Looks good 👍

d2lam (approved):

PR #319, opened 2019-01-25
feat(1412): pipeline metrics

d2lam:
Pipeline metrics should have events between a time range and these events duration (to show the trend in the UI). 

**Notes:** Need to observe the performance of this implementation after adding indexes. If it's too slow, we might need a separate table. 

**Blocked By:** 
- Indexes added to DB
- Modify `scan` method inside `datastore-sequelize` for time range: https://github.com/screwdriver-cd/datastore-sequelize/pull/41

**Related:** https://github.com/screwdriver-cd/screwdriver/issues/1412

### DIFF
--- a/lib/event.js
+++ b/lib/event.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const hoek = require('hoek');
 const BaseModel = require('./base');
 
 class EventModel extends BaseModel {
@@ -15,14 +16,18 @@ class EventModel extends BaseModel {
 
     /**
      * Return builds that belong to this event
-     * @return {Promise}                            List of builds
+     * @param  {String}   [config.startTime]     Search for builds after this startTime
+     * @param  {String}   [config.endTime]       Search for builds before this endTime
+     * @return {Promise}  Resolves to an array of builds
      */
-    getBuilds() {
-        const listConfig = {
+    getBuilds(config) {
+        const defaultConfig = {
             params: {
                 eventId: this.id
             }
         };
+
+        const listConfig = config ? hoek.applyToDefaults(defaultConfig, config) : defaultConfig;
 
         // Lazy load factory dependency to prevent circular dependency issues
         // https://nodejs.org/api/modules.html#modules_cycles
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -1155,7 +1155,7 @@ class PipelineModel extends BaseModel {
      * @param  {String}   [config.endTime]      Only look at events created before this endTime
      * @return {Promise}  [description]
      */
-    async getReports({ startTime, endTime }) {
+    async getMetrics({ startTime, endTime }) {
         // Get events during this time range
         const events = this.getEvents({
             startTime,
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -1040,6 +1040,8 @@ class PipelineModel extends BaseModel {
      * @param  {Object}   [config.paginate]                     Pagination parameters
      * @param  {Number}   [config.paginate.count]               Number of items per page
      * @param  {Number}   [config.paginate.page]                Specific page of the set to return
+     * @param  {String}   [config.startTime]                    Search for events after this startTime
+     * @param  {String}   [config.endTime]                      Search for events before this endTime
      * @return {Promise}  Resolves to an array of events
      */
     getEvents(config) {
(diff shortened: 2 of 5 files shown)


lib/pipeline.js
```diff
@@ -1145,6 +1147,66 @@ class PipelineModel extends BaseModel {
...
+                startTime,
+                endTime
+            });
+
+            // Somehow this event doesn't have any builds
+            if (!builds) {
+                return 0;
+            }
+
+            let minStartTime = new Date(builds[0].startTime);
```
d2lam, line 1182: Can make this block of code cleaner using `map` and `sort`, but will not be as efficient.

lib/pipeline.js
```diff
@@ -1147,6 +1149,59 @@ class PipelineModel extends BaseModel {
...
+
+    /**
+     * getMetrics for this pipeline
+     * @method getMetrics
+     * @param  {Object}   [config]              Configuration object
+     * @param  {String}   [config.startTime]    Look at events created after this startTime
+     * @param  {String}   [config.endTime]      Look at events created before this endTime
+     * @return {Promise}  Resolves to array of metrics for events belong to this pipeline
+     */
+    async getMetrics(config) {
```
tkyi, line 1161: Can probably change to 
```suggestion
    async getMetrics({ startTime = null, endTime = null }) {
```
That way you don't need the next couple lines.

d2lam: Will get `startTime already declared` if do it that way.

lib/pipeline.js
```diff
@@ -1147,6 +1149,59 @@ class PipelineModel extends BaseModel {
...
+        // if no config pass in then get all (will not pass startTime/endTime to datastore)
```
parthasl, line 1162: if no config, can it be defaulted to current quarter ?

lib/event.js
```diff
@@ -15,15 +16,19 @@ class EventModel extends BaseModel {
...
+     * @param  {String}   [config.startTime]     Search for builds after this startTime
+     * @param  {String}   [config.endTime]       Search for builds before this endTime
+     * @return {Promise}  Resolves to an array of builds
      */
-    getBuilds() {
-        const listConfig = {
+    getBuilds(config) {
+        const defaultConfig = {
             params: {
                 eventId: this.id
```
parthasl, line 26: this will not use index, as index is combination of eventId + createTime; need to change data-schema if access pattern is only eventId

lib/pipeline.js
```diff
@@ -1147,6 +1149,59 @@ class PipelineModel extends BaseModel {
...
+
+            let minStartTime = new Date(builds[0].startTime);
+            let maxEndTime = new Date(builds[0].endTime);
+
+            for (let i = 1; i < builds.length; i += 1) {
+                const s = new Date(builds[i].startTime);
+                const e = new Date(builds[i].endTime);
+
+                minStartTime = s < minStartTime ? s : minStartTime;
+                maxEndTime = e > maxEndTime ? e : maxEndTime;
```
parthasl, line 1190: sequelize aggregated functions to get min n max ?

lib/event.js
parthasl: retested, not big difference; we can go with current index and add later

lib/pipeline.js
d2lam: I think UI can have that option. And UI can also have the option "All Time", which will have no startTime, no endTime so that will be no config.

parthasl: 👍

lib/pipeline.js
parthasl: checked db, based on max # of builds per event, we can go with for loop for now

parthasl (approved):
👍

lib/pipeline.js
```diff
@@ -1147,6 +1149,59 @@ class PipelineModel extends BaseModel {
...
+            }
+
+            return Math.round((maxEndTime - minStartTime) / 1000); // round in seconds
+        };
+
+        // Generate metrics
+        const promiseArray = events.map(async (e) => {
```
minzcmu, line 1197: when we pull this in, should probably also increase the connection pool size(current number is 10), this can be a lot of concurrent calls

minzcmu (approved):

lib/pipeline.js
d2lam: will increase pool size 👍

PR #320, opened 2019-01-26
fix(1440): Convert string to JSON for custom Git status

tkyi:
## Context
Currently, you need to set each field in `meta.status` separately:
```
steps:
      - status: |
          meta set meta.status.coverage.status SUCCESS
          meta set meta.status.coverage.message 'Coverage is above 80%'
```
It'd be nicer if we could JSON parse a string for the user so the command can be done in one line:
```
steps:
      - status: |
           meta set --format json meta.summary.coverage '{"status":"SUCCESS","message":"Coverage is above 80%"}'
```

## Objective
This PR tries to JSON parse the string.

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/1440

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -2,6 +2,7 @@
 
 const BaseModel = require('./base');
 const hoek = require('hoek');
+const winston = require('winston');
 const { EXTERNAL_TRIGGER } = require('screwdriver-data-schema').config.regex;
 
 // Symbols for private members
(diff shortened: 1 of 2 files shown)


lib/build.js
```diff
@@ -133,18 +133,31 @@ function getStatusConfig({ metadata }) {
...
+        if (typeof checkObject !== 'object' && typeof checkObject !== 'string') {
+            return statusConfigs;
         }
+
+        // Parse JSON string
+        if (typeof checkObject === 'string') {
+            try {
+                checkObject = JSON.parse(checkObject);
+            } catch (e) {
+                return statusConfigs;
```
jithine: instead can it do `continue` here such that rest of status fields can be processed ?

tkyi: updated

jithine (approved):

PR #324, opened 2019-02-07
feat(1412): metrics for events & builds (pipeline scope)

d2lam:
- Adds build duration trend of an event
- Adds step duration trend of a build
- Removes unnecessary fields from returned metrics for event duration trend of a pipeline

Blocked by: https://github.com/screwdriver-cd/datastore-sequelize/pull/45

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -274,15 +274,19 @@ class BuildModel extends BaseModel {
 
     /**
     * Get models for all steps
-    * @method getStepsModel
+    * @param  {String}   [config.startTime]     Search for builds after this startTime
+    * @param  {String}   [config.endTime]       Search for builds before this endTime
+    * @method getSteps
     * @return {Promise}
     */
-    getStepsModel() {
-        const listConfig = {
+    getSteps(config) {
+        const defaultConfig = {
             params: {
                 buildId: this.id
             }
         };
+
+        const listConfig = config ? hoek.applyToDefaults(defaultConfig, config) : defaultConfig;
 
         // Lazy load factory dependency to prevent circular dependency issues
         // https://nodejs.org/api/modules.html#modules_cycles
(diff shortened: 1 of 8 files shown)


lib/build.js
```diff
@@ -517,6 +527,36 @@ class BuildModel extends BaseModel {
...
+
+    /**
+     * getMetrics for this build
+     * @method getMetrics
+     * @param  {Object}   [config]              Configuration object
+     * @param  {String}   [config.startTime]    Look at steps created after this startTime
+     * @param  {String}   [config.endTime]      Look at steps created before this endTime
+     * @return {Promise}  Resolves to array of metrics for steps belong to this event
+     */
+    async getMetrics(config) {
```
tkyi:
```suggestion
    async getMetrics({ startTime=null, endTime=null }) {
```

Does this work?
And you can remove lines 541, 542

lib/build.js
```diff
@@ -517,6 +527,36 @@ class BuildModel extends BaseModel {
...
+
+        // Get steps during this time range
+        const steps = await this.getSteps({
+            startTime,
+            endTime
+        });
+
+        // Generate metrics
+        const metrics = steps.map((s) => {
+            const { id, name, code } = s;
```
tkyi:
```suggestion
            const { id, name, code, endTime, startTime } = s;
            const duration = Math.round((new Date(endTime) - new Date(startTime)) / 1000);
```
not sure if this is any better

lib/build.js
d2lam: Ah doesn't let me do exactly that. It seems default param value in es6 will not work on JSON, it will give error `Unable to deconstruct endTime` if the field `endTime` is missing. 
However, found something similar:
```
async getMetrics(config = { startTime: null, endTime: null }) {
```
then need to reference it as `config.startTime` instead of `startTime`. 
Let me know if there are better ways.

lib/build.js
d2lam: ah couldn't do that before because it will conflict with the `startTime` and `endTime` outside. But can do it now 👍

lib/build.js
d2lam: Addressed here: https://github.com/screwdriver-cd/models/pull/324/commits/3b0ffbd5ce81304543b0142127426523148e6183

tkyi: Aw that's weird that it didn't work :\
Your workaround looks good to me

tkyi (approved):
👍

PR #331, opened 2019-02-21
feat(1447): skip ci creates empty event

d2lam:
Skip creating & starting builds if `skipMessage` is set

Related: https://github.com/screwdriver-cd/screwdriver/issues/1447

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -375,10 +375,10 @@ class EventFactory extends BaseFactory {
                 if (p.configPipelineId) {
                     // eslint-disable-next-line no-underscore-dangle
                     return this._getCommitSha(p.configPipelineId)
-                        .then((commitsha) => {
-                            modelConfig.configPipelineSha = commitsha;
+                        .then((commitSha) => {
+                            modelConfig.configPipelineSha = commitSha;
 
-                            return p.sync(commitsha);
+                            return p.sync(commitSha);
                         });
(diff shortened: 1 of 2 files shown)


lib/eventFactory.js
```diff
@@ -373,47 +375,45 @@ class EventFactory extends BaseFactory {
                 if (p.configPipelineId) {
                     // eslint-disable-next-line no-underscore-dangle
                     return this._getCommitSha(p.configPipelineId)
-                        .then((sha) => {
-                            modelConfig.configPipelineSha = sha;
+                        .then((commitsha) => {
```
tkyi: camel case?
```suggestion
                        .then((commitSha) => {
```

lib/eventFactory.js
```diff
@@ -426,9 +426,15 @@ class EventFactory extends BaseFactory {
...
                     return super.create(modelConfig);
                 })
                 .then((event) => {
+                    // Skip creating & starting builds
+                    if (skipMessage) {
+                        return event;
+                    }
+
                     if (modelConfig.type === 'pipeline') {
                         pipeline.lastEventId = event.id;
```
jithine, line 434: is this logic not required if skipping ?

d2lam: You're right. We should still update this field. Will move it

jithine (approved):

PR #333, opened 2019-02-22
fix: add fetchSteps flag to list builds function

minzcmu:
Currently we will get steps info from steps table for each build during list builds, this creates a lot of unnecessary queries to our DB since usually we do not need steps info when listing builds.

This PR will add a `fetchSteps` flag to list builds function and set it to `false` by default. Later on, if we really need to fetch steps for listing builds, we can still set the flag and get it.

### DIFF
--- a/lib/buildFactory.js
+++ b/lib/buildFactory.js
@@ -388,11 +388,20 @@ class BuildFactory extends BaseFactory {
      * @param  {Object}   config.paginate         Pagination parameters
      * @param  {Number}   config.paginate.count   Number of items per page
      * @param  {Number}   config.paginate.page    Specific page of the set to return
+     * @param  {Boolean}  [config.fetchSteps]     Fetch steps for all the builds, default is true
      * @return {Promise}                          Resolve builds after merging with step models
      */
     list(config) {
+        const fetchSteps = config.fetchSteps ? config.fetchSteps : false;
+
         return super.list(config)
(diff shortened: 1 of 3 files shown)


lib/buildFactory.js
```diff
@@ -388,11 +388,20 @@ class BuildFactory extends BaseFactory {
      * @param  {Object}   config.paginate         Pagination parameters
      * @param  {Number}   config.paginate.count   Number of items per page
      * @param  {Number}   config.paginate.page    Specific page of the set to return
+     * @param  {Boolean}  [config.fetchSteps]     Fetch steps for all the builds, default is true
      * @return {Promise}                          Resolve builds after merging with step models
      */
     list(config) {
+        const fetchSteps = config.fetchSteps ? config.fetchSteps : false;
```
jithine: does this change has to go the very last ? Alternatively I would just default this to `true` so that no-one breaks.

minzcmu: that's what i think in the first place to default it to `true` and that requires a lot of changes in our api and models to pass in the flag. Went through the code with Dao and we're not using / modifying steps when list builds. So it should be safe. Make it a flag so that later on if we need it we can always pass it

tkyi:
```suggestion
        const fetchSteps = config.fetchSteps || false;
```
should also work

tkyi (approved):

d2lam (approved):

PR #332, opened 2019-02-22
fix(1010): pr branch's prChain flag does not override base branch's one.

wdstar:
## Context

Currently, PR branch's `prChain` flag can override base branch's one.
It is contrary to the specifications. This PR fixes it.

## References

- https://github.com/screwdriver-cd/screwdriver/issues/1010#issuecomment-448142968
> #### Precondition:
> * Only the prChain flag of main branch will affect the job chain.
> * Even if prChain flag is changed in PRs, behavior of job chain does not change.
> * This is because, it is quite dangerous that anyone can change prChain flag in PR and run subsequent jobs easily.

PR #335, opened 2019-02-27
fix(1010): [small] update the workflowGraph of the event with prChain enabled.

wdstar:
## Context

Workflow graphs of PR chain enabled event have no job ids.
This PR adds job ids to those graphs.
And I add `rewire` package for unit tests. Let's write unit tests for each function!

## Note

- Currently `PR-#:` prefixed name is not allowed by the workflow graph validator. We should consider the addition of the prefix carefully.
```
[request,server,error] data: Error: child "workflowGraph" fails because [child "nodes" fails because ["nodes" at position 2 fails because [child "name" fails because ["name" with value "PR-1:first-job" fails to match the required pattern: /^~([\w-]+)$/, "name" with value "PR-1:first-job" fails to match the required pattern: /^[\w-]+$/, "name" with value "PR-1:first-job" fails to match the required pattern: /^~(sd@\d+:[\w-]+|(pr|commit)(:(.+))?)$/]]]]
    at postValidate (/usr/src/app/node_modules/hapi/lib/validation.js:214:30)
...
```

## References

- https://github.com/screwdriver-cd/screwdriver/issues/1010#issuecomment-462615428

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -265,6 +265,48 @@ function getLatestWorkflowGraph(config) {
     return Promise.resolve(pipeline.workflowGraph);
 }
 
+/**
+ * Update the workflowGraph
+ * @method updateWorkflowGraph
+ * @param  {Object}         config
+ * @param  {Pipeline}       pipeline            Pipeline
+ * @param  {Number}         pipeline.id         Pipeline id
+ * @param  {Object}         eventConfig
+ * @param  {Number}         [eventConfig.prNum] PR number
+ * @param  {Object}         workflowGraph       WorkflowGraph
+ * @resolves {Object}                           Resolves with workflowGraph
+ */
+function updateWorkflowGraph(config) {
+    const { pipeline, eventConfig, workflowGraph } = config;
+
+    if (pipeline.prChain) {
+        // eslint-disable-next-line global-require
+        const JobFactory = require('./jobFactory');
+        const jobFactory = JobFactory.getInstance();
+
+        return jobFactory.list({
+            params: { pipelineId: pipeline.id, archived: false },
+            search: { field: 'name', keyword: `PR-${eventConfig.prNum}:%` }
+        })
+            .then((prChainedJobs) => {
+                const nodes = workflowGraph.nodes;
+
+                prChainedJobs.forEach((job) => {
+                    // Add jobId to workflowGraph.nodes
+                    nodes.forEach((node) => {
+                        if (`PR-${eventConfig.prNum}:${node.name}` === job.name) {
+                            node.id = job.id;
+                        }
+                    });
+                });
+
+                return workflowGraph;
+            });
+    }
+
+    return Promise.resolve(workflowGraph);
+}
+
 class EventFactory extends BaseFactory {
     /**
      * Construct a EventFactory object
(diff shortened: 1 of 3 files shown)


lib/eventFactory.js
```diff
@@ -265,6 +265,48 @@ function getLatestWorkflowGraph(config) {
     return Promise.resolve(pipeline.workflowGraph);
 }
 
+/**
+ * Update the workflowGraph
+ * @method updateWorkflowGraph
+ * @param  {Object}         config
+ * @param  {Pipeline}       pipeline            Pipeline
```
tk3fftk: These JSDoc of params should be `config.*`

wdstar: I will fix it. Thanks.

wdstar: @tk3fftk fixing done.

tk3fftk: Sorry, my intention is `config.pipeline`, `config.eventConfig`... like other JSDoc.

lib/eventFactory.js
```diff
@@ -265,6 +265,48 @@ function getLatestWorkflowGraph(config) {
...
+ * @param  {Number}         pipeline.id         Pipeline id
+ * @param  {Object}         eventConfig.*
+ * @param  {Number}         [eventConfig.prNum] PR number
+ * @param  {Object}         workflowGraph       WorkflowGraph
+ * @resolves {Object}                           Resolves with workflowGraph
+ */
+function updateWorkflowGraph(config) {
+    const { pipeline, eventConfig, workflowGraph } = config;
+
+    if (pipeline.prChain) {
```
yuichi10: Don't you need to check `eventConfig.prRef` here?
Only event PR has prChain?

lib/eventFactory.js
wdstar: OK, I copied the existing `getLatestWorkflowGraph()`'s JSDoc, which was invalid.

lib/eventFactory.js
wdstar: No, we must check the base branch (`pipeline` object)'s `prChain` flag.
Please see https://github.com/screwdriver-cd/models/pull/332

tk3fftk (approved):
LGTM:+1:

lib/eventFactory.js
yuichi10: When the event come from commit, is there possibility the `pipeline.prChain` has true? 
If so maybe you also need to check the event is came from pr.
Then the if sentense will be `if (eventConfig.prRef && pipline.prChain)`.

sakka2: I agree with `if (eventConfig.prRef && pipline.prChain)` .
We need check if the event is pr or not.

wdstar: @yuichi10 Your advice is very correct. I fixed it. Thanks!

yuichi10 (approved):
LGTM

sakka2 (approved):
LGTM!

tkyi (approved):
Wondering if it would be better to combine `getLatestWorkflowGraph` and `updateWorkflowGraph`, since they take in similar arguments

Otherwise LGTM 👍

wdstar:
> Wondering if it would be better to combine `getLatestWorkflowGraph` and `updateWorkflowGraph`, since they take in similar arguments

@tkyi We should keep a function simple and let it to do one thing well. If it is not so, it will be more difficult to test a function.

minzcmu (approved):

PR #337, opened 2019-03-01
fix: do not fail build status update if git commit status update failed

minzcmu:
- make `build.updateCommitStatus` always resolve and print out the err otherwise it may fail `build.update()` and `executor.start()`. But that's probably not what we want. Git status update shouldn't be a blocker for these functions.

`executor.start()`: https://github.com/screwdriver-cd/models/blob/master/lib/build.js#L455
`build.update()`: https://github.com/screwdriver-cd/models/blob/master/lib/build.js#L508

Related to: https://github.com/screwdriver-cd/screwdriver/issues/1527

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -270,7 +270,12 @@ class BuildModel extends BaseModel {
             }
 
             return Promise.all(updateTasks);
-        });
+        })
+            .catch((err) => {
+                winston.error(`Failed to update commit status: ${err}`);
+
+                return Promise.resolve();
+            });
     }
 
     /**
--- a/test/lib/build.test.js
+++ b/test/lib/build.test.js
@@ -197,13 +197,12 @@ describe('Build Model', () => {
                 })
         );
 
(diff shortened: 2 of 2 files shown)


tkyi (approved):

lib/build.js
```diff
@@ -270,7 +270,12 @@ class BuildModel extends BaseModel {
             }
 
             return Promise.all(updateTasks);
-        });
+        })
+            .catch((err) => {
+                winston.error(`Failed to update commit status: ${err}`);
+
+                return Promise.resolve();
```
jithine, line 277: if the caller does not care about return values, this should be good. If `all` passes, it returns an array, in this case it returns nothing.

jithine (approved):

lib/build.js
minzcmu: yeah, i checked where it calls the function, it doesn't use the results resolve by this

PR #338, opened 2019-03-01
fix: do not fail pipeline.getJobs if cannot decorate open PRs

minzcmu:
If pipeline admin's git account is suspended, the pipeline page cannot load and other user cannot sync to update admin because `pipeline.getJobs` will failed at `scm.getOpenedPRs`.


jithine (approved):
👏 ooh this is why some pipelines were mysteriously not loading

DekusDenial (approved):

PR #339, opened 2019-03-05
fix: allow get step metrics for all steps for a single job

minzcmu:
- No actual changes is required, `build.getSteps()` already supports get all steps if `name` is undefined.

https://github.com/screwdriver-cd/models/blob/master/lib/build.js#L310

Related to: https://github.com/screwdriver-cd/screwdriver/issues/1412

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -280,36 +280,15 @@ class BuildModel extends BaseModel {
 
     /**
     * Get models for all steps
-    * @param  {String}   [config.startTime]     Search for builds after this startTime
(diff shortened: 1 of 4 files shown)


jithine (approved):

d2lam:
Should we change this so it doesn't need to actually query the steps: https://github.com/screwdriver-cd/models/blob/master/lib/build.js#L569-L573

Our discussion from last time was we can just call `build.steps` and do filtering instead of making calls to database.

minzcmu:
@d2lam makes sense! fixed!

d2lam (approved):

PR #340, opened 2019-03-06
feat(1412): more metrics

d2lam:
Adding `imagePullTime` and `queuedTime` to metrics

### DIFF
--- a/lib/event.js
+++ b/lib/event.js
@@ -59,22 +59,29 @@ class EventModel extends BaseModel {
         // Generate metrics
         const metrics = builds.map((b) => {
             const { id, jobId, eventId, createTime, status, startTime, endTime, stats } = b;
-            const { queueEnterTime, imagePullStartTime } = stats;
-            const queuedTime = findDuration(queueEnterTime, imagePullStartTime);
-            const imagePullTime = findDuration(imagePullStartTime, startTime);
             const duration = findDuration(startTime, endTime);
+            let queuedTime;
+            let imagePullTime;
+
+            // for backward compatibility, some old builds don't have stats field
+            if (stats) {
+                const { queueEnterTime, imagePullStartTime } = stats;
+
+                queuedTime = findDuration(queueEnterTime, imagePullStartTime);
+                imagePullTime = findDuration(imagePullStartTime, startTime);
+            }
 
             return {
                 id,
                 jobId,
                 eventId,
                 createTime,
-                queuedTime,
-                imagePullTime,
                 startTime,
                 endTime,
                 duration,
-                status
+                status,
+                queuedTime,
+                imagePullTime
             };
         });
 
--- a/lib/event.js
+++ b/lib/event.js
@@ -54,12 +54,28 @@ class EventModel extends BaseModel {
             endTime: config.endTime
         });
 
+        const findDuration = (start, end) => Math.round((new Date(end) - new Date(start)) / 1000);
+
         // Generate metrics
         const metrics = builds.map((b) => {
-            const { id, jobId, eventId, createTime, status, startTime, endTime } = b;
-            const duration = Math.round((new Date(endTime) - new Date(startTime)) / 1000);
+            const { id, jobId, eventId, createTime, status, startTime, endTime, stats } = b;
+            const { queueEnterTime, imagePullStartTime } = stats;
+            const queuedTime = findDuration(queueEnterTime, imagePullStartTime);
+            const imagePullTime = findDuration(imagePullStartTime, startTime);
+            const duration = findDuration(startTime, endTime);
 
-            return { id, jobId, eventId, createTime, status, duration };
+            return {
+                id,
+                jobId,
+                eventId,
+                createTime,
+                queuedTime,
+                imagePullTime,
+                startTime,
+                endTime,
+                duration,
+                status
+            };
         });
 
         return metrics;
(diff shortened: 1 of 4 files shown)


lib/event.js
```diff
@@ -54,12 +54,35 @@ class EventModel extends BaseModel {
...
+            const { id, jobId, eventId, createTime, status, startTime, endTime, stats } = b;
+            const duration = findDuration(startTime, endTime);
+            let queuedTime;
+            let imagePullTime;
+
+            // for backward compatibility, some old builds don't have stats field
+            if (stats) {
+                const { queueEnterTime, imagePullStartTime } = stats;
+
+                queuedTime = findDuration(queueEnterTime, imagePullStartTime);
```
tkyi, line 70: Should we make these metrics more consistent with what we're showing in the UI and just change this to `blockedTime = findDuration(createTime, imagePullStartTime)`? Or we can change UI to match. Not sure what makes more sense
https://github.com/screwdriver-cd/ui/blob/master/app/build/model.js#L81-L92

lib/pipeline.js
```diff
@@ -1172,34 +1172,58 @@ class PipelineModel extends BaseModel {
...
+                if (e > maxEndTime) {
+                    maxEndTime = e;
+                    eventStatus = status;
+                }
 
                 minStartTime = s < minStartTime ? s : minStartTime;
-                maxEndTime = e > maxEndTime ? e : maxEndTime;
+                if (queuedTime) {
+                    totalQueuedTime += queuedTime;
+                    totalImagePullTime += imagePullTime;
```
minzcmu, line 1210: better have a separate check for `imagePullTime `. Build can have `queueEnterTime`, `imagePullStartTime` but no `startTime` (failed when pulling image), in that case `imagePullTime` will be invalid

minzcmu (changes requested):

lib/pipeline.js
```diff
@@ -1172,34 +1172,58 @@ class PipelineModel extends BaseModel {
...
 
-            let minStartTime = new Date(builds[0].startTime);
-            let maxEndTime = new Date(builds[0].endTime);
-
-            for (let i = 1; i < builds.length; i += 1) {
-                const s = new Date(builds[i].startTime);
-                const e = new Date(builds[i].endTime);
+            let minStartTime = new Date(buildMetrics[0].startTime);
+            let maxEndTime = new Date(buildMetrics[0].endTime);
+            let eventStatus = buildMetrics[0].status;
```
tkyi, line 1189: Is it possible there's no status if `[skip ci]` was used?

lib/pipeline.js
```diff
@@ -1172,34 +1172,58 @@ class PipelineModel extends BaseModel {
...
-                maxEndTime = e > maxEndTime ? e : maxEndTime;
+                if (queuedTime) {
+                    totalQueuedTime += queuedTime;
+                    totalImagePullTime += imagePullTime;
+                }
             }
 
-            return Math.round((maxEndTime - minStartTime) / 1000); // round in seconds
+            return {
+                duration: Math.round((maxEndTime - minStartTime) / 1000), // round in seconds
```
minzcmu, line 1215: this can be `Nah` if last build of the event hasn't finished yet. Not sure if that's the expected behavior

lib/pipeline.js
```diff
@@ -1172,34 +1172,58 @@ class PipelineModel extends BaseModel {
...
+                const s = new Date(startTime);
+                const e = new Date(endTime);
+
+                // last build in the event
+                if (e > maxEndTime) {
+                    maxEndTime = e;
+                    eventStatus = status;
+                }
 
                 minStartTime = s < minStartTime ? s : minStartTime;
```
minzcmu: same problem here, if the last build has no startTime, e.g. `Frozen`.  `s < minStartTime` will always be false, and `minStartTime ` will be invalid

lib/pipeline.js
d2lam: `[skip ci]` won't have a build so it will not be part of `buildMetrics`

lib/pipeline.js
d2lam: I think `NaN` is expected if the build hasn't finished. It was part of my test here: https://github.com/screwdriver-cd/models/blob/master/test/lib/job.test.js#L445

lib/event.js
d2lam: Hmm yea, from the mock it looks like we includes that as `queuedTime`, which might be confusing when their builds are not really queued, but it was blocked. Should I change the key to `blockedTime`?

minzcmu (approved):

lib/event.js
tkyi: To me, it makes more sense as `queuedTime`, since when it's blocked, it gets put back in the queue.
But I think @jithin1987 was worried people would think the queue was really slow if blocked jobs are shown as in the queue the whole time. 

What do you think makes more sense? Or should we just split to show both `queuedTime` and `blockedTime`?

d2lam: Let's get this merged and we can change the key later, so that it's not blocked by Jithin.

tkyi (approved):
Wondering if it would be confusing to have only `queuedTime = findDuration(queueEnterTime, imagePullStartTime)` and `imagePullTime`
Might be better to have `queuedTime = findDuration(createTime, imagePullStartTime)` so that the times add up to total build duration
But we can address that later

Otherwise, LGTM 👍

PR #341, opened 2019-03-08
fix: [small] fix getJobsFromPR function

sakka2:
## Context
When I made PR with chainPR flag true, all jobs had started at same time unintentionally.
This is because jobs are filtered by job name which prefix is `PR-${prNum}:`.

k9a6a5c80a734df812067c924793b

## Objective
We don't need to filter jobs by job name. 
`workflowParser.getNextJobs()` method can get proper jobs which is triggered by `startFrom`, whatever its value is `~pr`, `~pr:branch`, `PR-4:jobname`, `jobA`, and so on...

### DIFF
--- a/test/lib/eventFactory.test.js
+++ b/test/lib/eventFactory.test.js
@@ -508,11 +508,6 @@ describe('Event Factory', () => {
                         eventId: model.id,
                         jobId: 9
                     }));
-                    assert.calledWith(buildFactoryMock.create, sinon.match({
-                        parentBuildId: 12345,
-                        eventId: model.id,
-                        jobId: 10
-                    }));
                 });
             });
 
@@ -775,7 +770,7 @@ describe('Event Factory', () => {
             }, {
                 id: 2,
                 pipelineId: 8765,
-                name: 'PR-1:test',
+                name: 'PR-1:publish',
                 permutations: [{
                     requires: ['~pr'],
                     sourcePaths: ['src/test/']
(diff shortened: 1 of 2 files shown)


test/lib/eventFactory.test.js
```diff
@@ -775,7 +770,7 @@ describe('Event Factory', () => {
             }, {
                 id: 2,
                 pipelineId: 8765,
-                name: 'PR-1:test',
+                name: 'PR-1:publish',
```
sakka2, line 910: `PR-1:publish` is proper job name if trigger `startFrom` is `~pr`.
https://github.com/screwdriver-cd/models/blob/master/test/lib/eventFactory.test.js#L207

test/lib/eventFactory.test.js
```diff
@@ -508,11 +508,6 @@ describe('Event Factory', () => {
                         eventId: model.id,
                         jobId: 9
                     }));
-                    assert.calledWith(buildFactoryMock.create, sinon.match({
-                        parentBuildId: 12345,
-                        eventId: model.id,
-                        jobId: 10
-                    }));
```
sakka2, line 532: The value of trigger `startFrom` is `~pr:branch`, so `jobId 10` should not be selected.
https://github.com/screwdriver-cd/models/blob/master/test/lib/eventFactory.test.js#L306

tk3fftk (approved):
LGTM👍

yuichi10 (approved):

s-yoshika (approved):

tkyi (approved):

PR #350, opened 2019-03-18
fix: not to call sync with SHA of parent event if it is pr event

sakka2:
## Context
Now, pipeline `sync()` method is called with parent event sha even if it is PR event.
This cause to create a new job with parent PR event config as a commit event.
Created new job doesn't have `PR-` prefix.

## Objective
Do not call `sync()` with parent event sha if it is PR event.

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -421,7 +421,7 @@ class EventFactory extends BaseFactory {
             // Sync pipeline to make sure workflowGraph is generated
             .then((p) => {
                 // Sync pipeline with the parentEvent sha and create jobs based on that sha
-                if (parentEventId) {
+                if (parentEventId && !prRef) {
                     modelConfig.parentEventId = parentEventId;
 
                     // for child pipelines restart event, sync with configPipelineSha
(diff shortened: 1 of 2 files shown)


lib/eventFactory.js
```diff
@@ -421,7 +421,7 @@ class EventFactory extends BaseFactory {
             // Sync pipeline to make sure workflowGraph is generated
             .then((p) => {
                 // Sync pipeline with the parentEvent sha and create jobs based on that sha
-                if (parentEventId) {
+                if (parentEventId && !prRef) {
```
tk3fftk, line 424: It's ok to check the event is PR or not. In my opinion, the intention looks unclear.

tk3fftk (approved):

wdstar (approved):

minzcmu (approved):

d2lam:
Snyk failed. Closing to retrigger.

lib/eventFactory.js
jithine: I am not clear on PR title vs code change, shouldn't this change be on line [428](https://github.com/screwdriver-cd/models/pull/350/files#diff-26258d311a2f870653c61d02805df225R428) below ?

sakka2: What I wanted to say is pipeline sync method with the parent event sha should not be called when PR event.
I will change the title.

PR #353, opened 2019-03-21
fix(1412): Aggregate job metrics daily

d2lam:
Accept `config.aggregate`. If `true`, will aggregate it daily. 
Result will be an array where each element is `{createTime, duration}`, where `duration` is the average of all builds run on that day. 

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -241,19 +241,20 @@ class Job extends BaseModel {
                 count: MAX_COUNT
             }
         };
-        const findDuration = (start, end) => dayjs(end).diff(dayjs(start), 'second');
+        const findMetrics = ({ id, jobId, eventId, createTime, status, startTime, endTime }) => {
+            const duration = startTime && endTime
+                ? dayjs(endTime).diff(dayjs(startTime), 'second')
+                : null;
+
+            return { id, jobId, eventId, createTime, status, duration };
+        };
 
         if (!config.aggregate) {
             // Get builds during this time range
             const builds = await this.getBuilds(options);
 
             // Generate metrics
-            return builds.map((b) => {
-                const { id, jobId, eventId, createTime, status, startTime, endTime } = b;
-                const duration = findDuration(startTime, endTime);
-
-                return { id, jobId, eventId, createTime, status, duration };
-            });
+            return builds.map(b => findMetrics(b));
         }
 
         const formatDate = dateTime => dayjs(dateTime).format('YYYY-MM-DD');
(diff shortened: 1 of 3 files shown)


minzcmu (approved):

lib/job.js
```diff
@@ -232,25 +233,84 @@ class Job extends BaseModel {
...
+                const currentBuild = builds[i];
+                const buildDate = formatDate(currentBuild.createTime);
+
+                if (!dayjs(buildDate).isSame(dayjs(currentDate))) {
+                    currentIndex += 1;
+                    currentDate = buildDate;
+                    buildArray.push([]);
+                }
+
+                buildArray[currentIndex].push(currentBuild);
```
pranavrc, line 285: Not sure I understand here. If the buildDate is the same as current date, the subsequent builds are pushed to the same index every time?

d2lam: If buildDate is same as currentDate then push to the same index. If it's different, then increment the index & set currentDate to the new buildDate. 

It will be an array of array 
```
[[all builds on 3/22],[all builds on date 3/23],[all builds on 3/24]...]
```
so that we can compute the daily avg more easily later.

DekusDenial:
Is it really not desirable to perform this daily aggregation on the db directly? I think I know the complication with sequelize, but still

parthasl:
Yes, agree; aggregations can be done directly in DB via MV and MV can be refreshed periodically n queried from api. I need generated sql's to do some tests. It also depends on the # of hits for this api and the underlying queries, joins n index.

Thanks,Lakshmi

 

    On Friday, March 22, 2019, 11:35:43 AM PDT, Dekus <notifications@github.com> wrote:  
 
 
Is it really not desirable to perform this daily aggregation on the db directly? I think I know the complication with sequelize, but still
cc: @parthasl thought on this?

d2lam:
@parthasl  Can you try the query and send it to me if it works? It might be complicated in sequelize. Thanks.
We can keep that as a separate track and not block this.

parthasl:
@d2lam what query u want me to fire ? In sequelize, i guess it prints the query in debug mode.


    On Monday, March 25, 2019, 1:56:14 PM PDT, Dao Lam <notifications@github.com> wrote:  
 
 
@parthasl Can you try the query and send it to me if it works? It might be complicated in sequelize. Thanks.
We can keep that as a separate track and not block this.

d2lam:
@parthasl  The query to do the same thing as this PR: select records within time range, calculate the duration for each record, group them by the "date" part only,  then average the duration for each date, since we want to do the same thing in the DB instead of Models. 
Thank you! 

Are we ok with keeping that as a separate track so that it doesn't block this PR?

PR #354, opened 2019-03-22
fix: Make sure user exists when checking admin permission

tkyi:
Fixing a small bug
Only happens if an admin is added manually in the DB to the Screwdriver admins list that does not exist as a Screwdriver user

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -460,6 +460,10 @@ class PipelineModel extends BaseModel {
                 scmContext: this.scmContext
             });
 
+            if (!user) {
+                return false;
+            }
+
             // eslint-disable-next-line no-await-in-loop
             const permission = await user.getPermissions(scmUri);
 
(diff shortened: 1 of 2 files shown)


d2lam:
Hmmm what's the purpose of adding the admin manually?
It will go through this flow and get removed from the admin list anyway.

tkyi:
If we're trying to bulk-add admins, it can be nice to do it manually through the DB rather than by asking each potential admin to Start a build or Sync a pipeline through the UI.
This PR will prevent that issue where a manually-added admin is not actually a Screwdriver user and puts the pipeline in a bad state.

@d2lam Are you saying we should never manually add an admin? Or that this change prevents adding them?

d2lam:
Two cases when adding an admin manually:
- If he/she is in the database, this problem will not happen anyway
- If he/she is not in the database, this change will remove him/her from the admin list. So what is the point of adding him/her in the first place?

tkyi:
I think it's just a safety for the case where we manually add an admin that is not in the database. If the user does not exist, this code will remove them and we will not run into the error we were seeing before.

tkyi:
I'm okay with closing this PR if it doesn't seem to have much value

tkyi:
Closing PR since it doesn't seem to be of value.

PR #355, opened 2019-03-27
fix(1412): include steps metrics in builds

d2lam:
This is so that UI will not need to make separate calls to get steps.
Step metrics will be included as part of `job/id/metrics`. The results will look like this:
```node
// builds for this job
[{
id: 1234 //build id
duration: ...
steps: [{
   id: 1 // step id
   duration: ...
}, {
   id: 2
   duration ... 
}]
}, ... ]
```

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -1,14 +1,14 @@
 'use strict';
 
 const BaseModel = require('./base');
-const hoek = require('hoek');
 const dayjs = require('dayjs');
+const hoek = require('hoek');
 const getAnnotations = require('./helper').getAnnotations;
-const START_INDEX = 3;
-const MAX_COUNT = 1000;
 const executor = Symbol('executor');
 const tokenGen = Symbol('tokenGen');
 const apiUri = Symbol('apiUri');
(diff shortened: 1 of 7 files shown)


lib/job.js
```diff
@@ -241,20 +242,26 @@ class Job extends BaseModel {
                 count: MAX_COUNT
             }
         };
-        const findMetrics = ({ id, jobId, eventId, createTime, status, startTime, endTime }) => {
+        const findMetrics = (build, aggregate) => {
```
jithine: can this function `findMetrics` be moved outside of this function ?

d2lam: Can do. 
But what's the reason? The function is not too long. Will need to pass in additional param (stepName) if move it outside :(

jithine: I generally avoid nested functions. Readability wise great, but not so for performance since these function objects are created on each run.

jithine (approved):

lib/job.js
d2lam: Okay 👍 
I refactored

PR #357, opened 2019-04-01
fix(1412): Aggregate event metrics daily

d2lam:

### DIFF
--- a/lib/helper.js
+++ b/lib/helper.js
@@ -54,7 +54,7 @@ function parseTemplateConfigName(config) {
  * recursively fetching records (ex: builds/events) until the end
  * @method getAllRecords
  * @param  {String}       funcName    Function to fetch. For example: 'getBuilds'
- * @param  {opts}         opts        opts to pass into scan function
+ * @param  {Object}       opts        opts to pass into scan function
  * @param  {Array}        resultArray Result so far
  * @param  {String}       date        Current date for aggregation
  * @param  {Number}       index       Current index (each index is a different date)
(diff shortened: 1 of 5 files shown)


lib/helper.js
```diff
@@ -48,7 +50,53 @@ function parseTemplateConfigName(config) {
     return {};
 }
 
+/**
+ * recursively fetching records (ex: builds/events) until the end
+ * @method getAllRecords
+ * @param  {String}       funcName    Function to fetch. For example: 'getBuilds'
+ * @param  {opts}         opts        opts to pass into scan function
```
tkyi:
```suggestion
 * @param  {Object}         opts        opts to pass into scan function
```
?

lib/helper.js
```diff
@@ -48,7 +50,53 @@ function parseTemplateConfigName(config) {
...
+ * @param  {Array}        resultArray Result so far
+ * @param  {String}       date        Current date for aggregation
+ * @param  {Number}       index       Current index (each index is a different date)
+ * @return {Array}                    Array where each element is an array of records belong to the same date
+ */
+async function getAllRecords(funcName, opts, resultArray, date, index) {
+    const builds = await this[funcName](opts);
+    const length = builds.length;
```
tkyi, line 65: do you need make sure `builds` is an array?

lib/helper.js
```diff
@@ -48,7 +50,53 @@ function parseTemplateConfigName(config) {
...
+        }
+
+        resultArray[currentIndex].push(currentBuild);
+    }
+    // last page
+    if (length < opts.paginate.count) {
+        return resultArray;
+    }
+
+    // might have more data, continue fetchign
```
tkyi:
```suggestion
    // might have more data, continue fetching
```

DekusDenial (approved):
LGTM, looking forwards to seeing the query option for agg duration

lib/helper.js
d2lam: it's from `list` methods so should always be an array?

tkyi (approved):

PR #361, opened 2019-04-10
fix(1412): allow week/month/year aggregation

d2lam:
This will allow aggregation option like `day/week/month/year`

### DIFF
--- a/lib/helper.js
+++ b/lib/helper.js
@@ -53,18 +53,18 @@ function parseTemplateConfigName(config) {
 /**
  * recursively fetching records (ex: builds/events) until the end
  * @method getAllRecords
- * @param  {String}       funcName    Function to fetch. For example: 'getBuilds'
- * @param  {String}       aggregate   Aggregate data by day/month/week/year
- * @param  {Object}       opts        opts to pass into scan function
- * @param  {Array}        resultArray Result so far
- * @param  {String}       date        Current date for aggregation
- * @param  {Number}       index       Current index (each index is a different date)
- * @return {Array}                    Array where each element is an array of records belong to the same date
+ * @param  {String}       funcName            Function to fetch. For example: 'getBuilds'
+ * @param  {String}       aggregateInterval   Aggregate data by day/month/week/year
+ * @param  {Object}       opts                opts to pass into scan function
+ * @param  {Array}        resultArray         Result so far
+ * @param  {String}       date                Current date for aggregation
+ * @param  {Number}       index               Current index (each index is a different date)
+ * @return {Array}                            Array where each element is an array of records belong to the same date
  */
-async function getAllRecords(funcName, aggregate, opts, resultArray, date, index) {
-    if (!['day', 'week', 'month', 'year'].includes(aggregate)) {
+async function getAllRecords(funcName, aggregateInterval, opts, resultArray, date, index) {
+    if (!['day', 'week', 'month', 'year'].includes(aggregateInterval)) {
         throw new Error('Invalid aggregation option. ' +
-            `${aggregate} is not one of these values ['day', 'week', 'month', 'year']`);
+            `${aggregateInterval} is not one of these values ['day', 'week', 'month', 'year']`);
     }
     const builds = await this[funcName](opts);
     const length = builds.length;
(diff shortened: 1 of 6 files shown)


lib/job.js
```diff
@@ -275,7 +275,8 @@ class Job extends BaseModel {
         }
 
         options.paginate.page = 1;
-        const allBuilds = await getAllRecords.apply(this, ['getBuilds', options, [[]]]);
+        const allBuilds = await getAllRecords.apply(this,
+            ['getBuilds', config.aggregate, options, [[]]]);
```
DekusDenial: wondering if u know all the args, why not use `.call`?

lib/helper.js
```diff
@@ -54,13 +54,18 @@ function parseTemplateConfigName(config) {
...
+ * @param  {String}       aggregate   Aggregate data by day/month/week/year
  * @param  {Object}       opts        opts to pass into scan function
  * @param  {Array}        resultArray Result so far
  * @param  {String}       date        Current date for aggregation
  * @param  {Number}       index       Current index (each index is a different date)
  * @return {Array}                    Array where each element is an array of records belong to the same date
  */
-async function getAllRecords(funcName, opts, resultArray, date, index) {
+async function getAllRecords(funcName, aggregate, opts, resultArray, date, index) {
+    if (!['day', 'week', 'month', 'year'].includes(aggregate)) {
```
DekusDenial: https://github.com/screwdriver-cd/screwdriver/pull/1582#discussion_r274465814

lib/pipeline.js
```diff
@@ -1292,11 +1292,12 @@ class PipelineModel extends BaseModel {
 
         options.paginate.page = 1;
 
-        const allEvents = await getAllRecords.apply(this, ['getEvents', options, [[]]]);
+        const allEvents = await getAllRecords.apply(this,
+            ['getEvents', config.aggregate, options, [[]]]);
```
DekusDenial: same here

lib/job.js
d2lam: I guess I can use either. Are there any difference performance difference?

DekusDenial: Technical answer: yes
Is it a lot? Depends on number of args
Is this blocking? No, hence this is just a comment ;-P

d2lam: Just curious, which one has better performance?

DekusDenial: Call is bit faster than apply

lib/job.js
```diff
@@ -13,19 +13,19 @@ const MAX_COUNT = 1000;
 /**
  * Find metrics and step metrics related to the build
  * @method findMetrics
- * @param  {Object}    build      build object
- * @param  {String}    stepName   only include this step
- * @param  {Boolean}   aggregate  Whether to include step metrics
- * @return {Array}                Array of metrics
+ * @param  {Object}    build              build object
+ * @param  {String}    stepName           only include this step
+ * @param  {Boolean}   aggregateInterval  Include step metrics if no aggregation
```
tkyi: ~Boolean~ String

lib/job.js
```diff
@@ -249,11 +249,11 @@ class Job extends BaseModel {
...
-     * @param  {Object}   [config]              Configuration object
-     * @param  {String}   [config.startTime]    Look at builds created after this startTime
-     * @param  {String}   [config.endTime]      Look at builds created before this endTime
-     * @param  {String}   [config.stepName]     Only include this step
-     * @param  {Boolean}  [config.aggregate]    Whether to aggregate the data
+     * @param  {Object}   [config]                    Configuration object
+     * @param  {String}   [config.startTime]          Look at builds created after this startTime
+     * @param  {String}   [config.endTime]            Look at builds created before this endTime
+     * @param  {String}   [config.stepName]           Only include this step
+     * @param  {Boolean}  [config.aggregateInterval]  Whether to aggregateInterval the data
```
tkyi: ~Boolean~ String

tkyi (approved):

lib/helper.js
```diff
@@ -53,14 +53,19 @@ function parseTemplateConfigName(config) {
...
+ * @param  {String}       aggregateInterval   Aggregate data by day/month/week/year
+ * @param  {Object}       opts                opts to pass into scan function
+ * @param  {Array}        resultArray         Result so far
+ * @param  {String}       date                Current date for aggregation
+ * @param  {Number}       index               Current index (each index is a different date)
+ * @return {Array}                            Array where each element is an array of records belong to the same date
  */
-async function getAllRecords(funcName, opts, resultArray, date, index) {
+async function getAllRecords(funcName, aggregateInterval, opts, resultArray, date, index) {
+    if (!['day', 'week', 'month', 'year'].includes(aggregateInterval)) {
```
minzcmu: do we still need to check here? seeing API has this check already for the query param

minzcmu (approved):
small comment, 👍

lib/helper.js
d2lam: probably don't need. I wrote this before the API, so that's why I had this check. But API already checks so I can remove it 👍

PR #362, opened 2019-04-11
fix(1412): use sortBy to sort

d2lam:
Need to use `sortBy` since if there are indexes, it will override the default sortKey: https://github.com/screwdriver-cd/datastore-sequelize/blob/master/index.js#L397-L398


DekusDenial (approved):
Lol

pranavrc (approved):
image

PR #365, opened 2019-04-12
fix: sorting metrics

d2lam:

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -1276,19 +1276,19 @@ class PipelineModel extends BaseModel {
 
         if (!config.aggregate) {
             const events = await this.getEvents(options);
-            const metrics = [];
-            const promiseArray = events.map(async (e) => {
+            const metrics = await Promise.all(events.map(async (e) => {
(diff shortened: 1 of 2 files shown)


lib/pipeline.js
```diff
@@ -1276,19 +1276,19 @@ class PipelineModel extends BaseModel {
 
         if (!config.aggregate) {
             const events = await this.getEvents(options);
-            const metrics = [];
-            const promiseArray = events.map(async (e) => {
+            const metrics = await Promise.all(events.map(async (e) => {
```
DekusDenial, line 1279: what's the purpose of `Promise.all` here `events.map` already resolve to a series of event metrics?

d2lam: cannot use `await` inside regular `events.map` because it's not async: https://stackoverflow.com/a/42497383/1522257

minzcmu (approved):

lib/pipeline.js
DekusDenial: oh right, forgot async function return promise, my bad

PR #367, opened 2019-04-15
fix(1058): Allow creator to be passed into event create (2)

tkyi:
## Context
When an event is started by a periodic build, it can be hard for users to know if the build was started manually by a user, through a commit, or by Screwdriver automatically.

## Objective
This PR allows `creator` to be passed into event create. We can pass in creator as `sd:scheduler` instead of defaulting to the commit author for periodic builds.

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/1058
Blocked by https://github.com/screwdriver-cd/data-schema/pull/329


lib/eventFactory.js
```diff
@@ -471,20 +473,31 @@ class EventFactory extends BaseFactory {
...
             })
             .then(pipeline => pipeline.token
-                .then(token => this.scm.decorateAuthor({ // decorate user who creates this event
-                    username,
-                    scmContext,
-                    token
-                }).then((creator) => {
-                    modelConfig.creator = creator;
+                .then((token) => {
+                    if (!modelConfig.creator) {
```
jithine, line 477: is there anything else in the event which marks it as a periodic build ? wondering if creator should be configurable or should be inferred from another field which stands for periodic build.

tkyi: just going to use `creator` and `causeMessage`
the executor-queue will pass this info in when starting an event

jithine (approved):

pranavrc (approved):
image

PR #368, opened 2019-04-15
fix(1412): edge cases for aggregated metrics

d2lam:
Address some edge cases for aggregation:

If one of the events in the aggregate interval has duration `null`, then it makes the whole thing `null`
Similar for job metrics.

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -289,7 +289,7 @@ class Job extends BaseModel {
                 findMetrics(b, config.stepName, config.aggregateInterval).duration);
             let emptyCount = 0;
 
-            const avg = (metrics.reduce((acc, current) => {
+            const totalDuration = metrics.reduce((acc, current) => {
                 if (current) {
                     return acc + current;
                 }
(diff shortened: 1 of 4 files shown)


lib/pipeline.js
```diff
@@ -95,21 +95,37 @@ const sumAggregatedEventMetrics = async (events) => {
...
     let totalDuration = 0;
-    let emptyEvent = 0;
+    const emptyCount = {
+        event: 0,
+        queuedTime: 0,
+        imagePullTime: 0,
+        duration: 0
+    };
+
+    const checkForEmpty = (fieldName, time) => {
```
tkyi, line 106: could you add a comment or some js docs for this?

lib/job.js
```diff
@@ -284,13 +284,23 @@ class Job extends BaseModel {
...
+
+            const avg = (metrics.reduce((acc, current) => {
+                if (current) {
+                    return acc + current;
+                }
+
+                emptyCount += 1;
+
+                return acc;
+            })) / (metrics.length - emptyCount);
```
jithine: might want to check for division by zero here, in case metrics is empty.

lib/pipeline.js
```diff
@@ -1296,16 +1313,25 @@ class PipelineModel extends BaseModel {
...
 
-        return Promise.all(allEvents.map(async (sameDateEvents) => {
-            const { totalQueuedTime, totalImagePullTime, totalDuration, emptyEvent }
-                = await sumAggregatedEventMetrics(sameDateEvents);
-            const length = sameDateEvents.length - emptyEvent;
+        return Promise.all(allEvents.map(async (sameIntervalEvents) => {
+            const length = sameIntervalEvents.length;
+            const { totalQueuedTime, totalImagePullTime, totalDuration, emptyCount }
+                = await sumAggregatedEventMetrics(sameIntervalEvents);
+            const duration = length > emptyCount.duration
```
tkyi, line 1320: wondering if this logic would be better in a helper function, but we can refactor that at a different time

tkyi (approved):

PR #372, opened 2019-04-23
feat(1333): Add function to get triggers given a pipeline id (1)

tkyi:
## Context
It would be nice if users could see what pipelines/jobs run after jobs in their pipeline are done.

## Objective
This PR adds a method `getTriggers` that returns an object with jobName and destination trigger data.

Example output:
```
[{
  jobName: 'main',
  triggers: ['~sd@12345:main']
}, {
  jobName: 'beta',
  triggers: ['~sd@12345:test', '~sd@223344:test']
}, {
  jobName: 'prod',
  triggers: ['~sd@556677:deploy']
}]
```

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/1333

### DIFF
--- a/lib/triggerFactory.js
+++ b/lib/triggerFactory.js
@@ -44,20 +44,21 @@ class TriggerFactory extends BaseFactory {
             .then((pipeline) => {
                 if (pipeline) {
                     // Get pipeline job names
-                    return pipeline.jobs.then(jobs => Promise.all(
-                        jobs.map(j =>
-                            // Get dest triggers for each src job
-                            this.list({
-                                params: {
-                                    src: `~sd@${pipelineId}:${j.name}`
-                                }
-                            // Push jobName and dest triggers to result
-                            }).then((triggersArr) => {
-                                const triggers = triggersArr.map(t => t.dest);
+                    return pipeline.getJobs()
+                        .then(jobs => Promise.all(
+                            jobs.map(j =>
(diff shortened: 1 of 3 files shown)


lib/triggerFactory.js
```diff
@@ -28,6 +28,43 @@ class TriggerFactory extends BaseFactory {
...
+        const pipelineFactory = PipelineFactory.getInstance();
+        const result = [];
+
+        return pipelineFactory.get(pipelineId)
+            .then((pipeline) => {
+                if (pipeline) {
+                    // Get pipeline job names
+                    return pipeline.getJobs()
+                        .then(jobs => Promise.all(
+                            jobs.map(j =>
```
jithine, line 50: do we need to filter out PR jobs ?

tkyi: if PRs never have external triggers, yes

tkyi: updated

jithine: that seems to be the ask here https://github.com/screwdriver-cd/screwdriver/issues/1606 but we can get there much later, it needs a lot of design changes,

jithine (approved):

lib/triggerFactory.js
tkyi: yeah I think that can be addressed at a later time

PR #373, opened 2019-04-23
fix: sync PRs when chainPR was changed

yuichi10:
## Context
When change chainPR of pipeline, UI does not change automatically.
This PR sync PRs and change UI automatically when chainPR was changed.

## Objective
- syncPRs function call syncPR functions. 
   To sync chainPR's status, we need to get pipeline configuration to check next job of workflowParser. Now syncPRs function does not see pipeline configuration. If syncPRs get pipeline configuration, there are almost no difference to call syncPR function.

- syncPRs when chainPR was changed
  Maybe it took many time to sync PRs, because it get pipeline configuration and check next job. Thus I call syncPRs function only chainPR was changed.

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -425,12 +425,12 @@ class EventFactory extends BaseFactory {
             pr: {},
             prNum
         };
-        let beforeChainPR = '';
+        let prevChainPR = '';
 
         return pipelineFactory.get(pipelineId)
             // Sync pipeline to make sure workflowGraph is generated
             .then((p) => {
-                beforeChainPR = p.chainPR;
+                prevChainPR = p.chainPR;
                 // Sync pipeline with the parentEvent sha and create jobs based on that sha
                 if (parentEventId && !prRef) {
                     modelConfig.parentEventId = parentEventId;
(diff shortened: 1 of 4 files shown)


lib/pipeline.js
```diff
@@ -441,14 +361,17 @@ class PipelineModel extends BaseModel {
...
-                ]);
-            }));
+                openedPRs.forEach((openedPR) => {
+                    const prNum = openedPR.name.split('-')[1];
+
+                    synced.push(this.syncPR(prNum));
+                });
+
+                return Promise.all(synced)
+                    .then(() => this._archiveClosePRs(existingJobs, openedPRs));
```
sakka2, line 373: [ask] why do we need to call this function after finishing syncPR? It seems that closed PR job will be archived in syncPR function.

yuichi10: To archive jobs which are in closed PR.
SyncPR function does not have this feature, thus I add this function.

sakka2: 👍

sakka2 (approved):

lib/eventFactory.js
```diff
@@ -425,10 +425,12 @@ class EventFactory extends BaseFactory {
             pr: {},
             prNum
         };
+        let beforeChainPR = '';
```
s-yoshika: [imo] I prefer prevChainPR.

s-yoshika (approved):
Add a small comment, but LGTM.

lib/eventFactory.js
yuichi10: I will fix that.

jithine (approved):

minzcmu (approved):

yuichi10:
@minz1027 
Thank you!!!

PR #375, opened 2019-04-27
feat(1455): Add rootDir to sourcePaths (4)

tkyi:
## Context
Would be nice if users could put screwdriver.yaml not at root.

## Objective
This PR adjusts startBuild params so that if rootDir is set:
- when no sourcePaths: start build if changedFile matches `rootDir`
- when sourcePaths: start build if changedFile matches `<rootDir>/<sourcePath>`

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/1455

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -10,7 +10,6 @@ const {
     RELEASE_TRIGGER,
     TAG_TRIGGER
 } = require('screwdriver-data-schema').config.regex;
-const path = require('path');
 const workflowParser = require('screwdriver-workflow-parser');
 const winston = require('winston');
 
@@ -146,16 +145,11 @@ function startBuild(config) {
             throw new Error('Your SCM does not support Source Paths');
         }
 
-        let paths = sourcePaths || [];
+        const paths = sourcePaths || [];
 
         // Add rootDir as a sourcePath if no sourcePaths
-        // Or prefix all sourcePaths with rootDir
-        if (rootDir) {
-            if (paths.length === 0) {
-                paths.push(`${rootDir}/`);
-            } else {
-                paths = paths.map(p => path.join(rootDir, p));
-            }
+        if (rootDir && paths.length === 0) {
+            paths.push(`${rootDir}/`);
         }
 
         hasChangeInSourcePaths = changedFiles.some(file =>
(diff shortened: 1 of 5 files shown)


lib/pipeline.js
```diff
@@ -139,6 +139,33 @@ const sumAggregatedEventMetrics = async (events) => {
...
+        rootDir: 3
+    };
+
+    if (!scmUri) {
+        return '';
+    }
+    const uriInfos = scmUri.split(':');
+    const position = fieldMap[field];
+
+    if (uriInfos.length < position + 1) {
```
minzcmu: this can just be `position >= uriInfos.length`

minzcmu (commented):
small comment, otherwise LGTM 👍

minzcmu (approved):

lib/eventFactory.js
```diff
@@ -115,35 +116,59 @@ function getJobsFromPR(config) {
...
+
+        let paths = sourcePaths || [];
+
+        // Add rootDir as a sourcePath if no sourcePaths
+        // Or prefix all sourcePaths with rootDir
+        if (rootDir) {
+            if (paths.length === 0) {
+                paths.push(`${rootDir}/`);
+            } else {
+                paths = paths.map(p => path.join(rootDir, p));
```
jithine: I think we should not prefix `sourcePaths` with `rootDir` If user want to use `sourcePaths` along with `rootDir` let that be implicitly set in `sourcePaths`. I can see use cases where these 2 need not be under same path.

tkyi: I'm wondering if that's confusing, since all `steps` in a `screwdriver.yaml` file would be in the context of root. I would prefer if the whole yaml was consistent.

jithine: `steps` scoped to `rootDir` is perfectly fine. But consider a use case for a mono repo like this

```
src/[cpp/php/...]
build/[cpp/php/...]
```

making `sourcePaths` tie to `rootDir` will prevent users from using `sourcePaths` for above repo structure.

tkyi: Okay if we do it this way we need to make sure we're very explicit in the documentation about it.

jithine (approved):

PR #376, opened 2019-05-01
fix: improve DB query efficiency

minzcmu:
- in event create, call `pipeline.getJobs()` instead of `pipeline.jobs` to get only unarchived jobs.
- in template factory, only fetch the latest template instead of fetching them all and pick the first one when user has no tag and version in template config. 

Related to: https://github.com/screwdriver-cd/screwdriver/issues/1603

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -205,7 +205,7 @@ function createBuilds(config) {
 
     return Promise.all([
         pipeline.branch,
-        pipeline.jobs
+        pipeline.getJobs()
     ]).then(([branch, jobs]) => {
         // When startFrom is ~commit, ~release, ~tag, ~sd@
         const jobsFromTrigger = getJobsFromTrigger({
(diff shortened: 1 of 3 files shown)


lib/eventFactory.js
```diff
@@ -205,7 +205,7 @@ function createBuilds(config) {
 
     return Promise.all([
         pipeline.branch,
-        pipeline.jobs
+        pipeline.getJobs()
```
tkyi, line 208: should we pass in `type` based on `isPR`?

tkyi (approved):

lib/eventFactory.js
minzcmu: yeah, we can but just found out that the getJobs is fetching all the jobs anyway. And if you set the type, it will parse the list and filter it out. So don't think the extra logic will help a lot...

tkyi: ah okay

tkyi:
Need to fix a test

tkyi (approved):

PR #379, opened 2019-05-06
fix: remove steps when remove build

minzcmu:
remove steps of the build when removing build

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -558,6 +558,28 @@ class BuildModel extends BaseModel {
     }
 
     /**
+     * Remove all steps associated with this build and the build itself
+     * @return {Promise}        Resolves to null if remove successfully
+     */
+    remove() {
+        const removeSteps = (() =>
+            // Iterate through the steps and remove them
+            this.getSteps().then((steps) => {
+                if (steps.length === 0) {
+                    // Done removing steps
+                    return null;
+                }
+
+                return Promise.all(steps.map(step => step.remove()))
+                    .then(() => removeSteps());
+            }));
+
+        // Remove steps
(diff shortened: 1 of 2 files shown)


parthasl (approved):
👍

minzcmu:
this is not really efficient, when you delete a pipeline with a lot of builds and a lot of steps. Need to fire a lot of DB request to do that. convert this to a story so that we can delete steps by bulk

lib/build.js
```diff
@@ -557,6 +557,28 @@ class BuildModel extends BaseModel {
...
+                    // Done removing steps
+                    return null;
+                }
+
+                return Promise.all(steps.map(step => step.remove()))
+                    .then(() => removeSteps());
+            }));
+
+        // Remove steps
+        return removeSteps()
```
jithine, line 578: should do this in `finally` rather than `then`

lib/build.js
```diff
@@ -557,6 +557,28 @@ class BuildModel extends BaseModel {
...
+        const removeSteps = (() =>
+            // Iterate through the steps and remove them
+            this.getSteps().then((steps) => {
+                if (steps.length === 0) {
+                    // Done removing steps
+                    return null;
+                }
+
+                return Promise.all(steps.map(step => step.remove()))
+                    .then(() => removeSteps());
```
jithine, line 574: why is this being done recursively ? is `steps` dynamic ?

minzcmu: previously we have default pagination. That's why it's recursive. Closed this anyway in favor of https://github.com/screwdriver-cd/screwdriver/issues/1614

PR #383, opened 2019-05-15
fix(1333): Make getTriggers db call more efficient

tkyi:
## Context
Need to make getTriggers call more efficient. 
Before, we were making a call to the DB per pipeline job, which would potentially be a lot of calls depending on number of jobs. Now, we make one call total to the DB to get triggers for all pipeline jobs.

## Objective
This PR changes DB logic to make the call more efficient.

This is probably what the translated query becomes:
_Before:_
1.
```
SELECT
 trigger
FROM
 triggers
WHERE src='~sd@123:main';
```
2.
```
SELECT
 trigger
FROM
 triggers
WHERE src='~sd@999:test';
```
3.
```
SELECT
 trigger
FROM
 triggers
WHERE src='~sd@555:publish';
```

_After:_
```
SELECT
 trigger
FROM
 triggers
WHERE
 src IN ('~sd@123:main', '~sd@999:test', '~sd@555:publish');
```

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/1333

### DIFF
--- a/lib/triggerFactory.js
+++ b/lib/triggerFactory.js
@@ -46,20 +46,28 @@ class TriggerFactory extends BaseFactory {
                 if (pipeline) {
                     // Get pipeline job names
                     return pipeline.getJobs({ type: type || 'pipeline' })
-                        .then(jobs => Promise.all(
-                            jobs.map(j =>
-                                // Get dest triggers for each src job
-                                this.list({
-                                    params: {
-                                        src: `~sd@${pipelineId}:${j.name}`
-                                    }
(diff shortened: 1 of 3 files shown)


DekusDenial (approved):

DekusDenial:
would be nice to see the translated downstream query and the reduction in observed response time

tkyi:
@DekusDenial I updated my description to put what I think the translated downstream query is

minzcmu (approved):

PR #386, opened 2019-06-18
fix(1364): fix to consider ~pr:pipelineBranch trigger

yokawara:
## Context
Branch filtering doesn't work with `~pr` trigger when use pipeline's branch for the filter.

## Objective
Fix triggers like `~pr:pipelineBranch` to work properly.

## References
https://github.com/screwdriver-cd/screwdriver/issues/1364

## License

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -82,21 +82,30 @@ function getJobsFromJobName(config) {
  * @param  {Object}    config.pipelineConfig.workflowGraph   Object with nodes and edges that represent the order of jobs
  * @param  {String}    config.startFrom                      Startfrom (e.g. ~commit, ~pr, etc)
  * @param  {Boolean}   config.chainPR                        Flag of triggering subsequent job after pull request job
+ * @param  {String}    config.branch                         triggered branch name
  * @resolves {Array}                                         Array of PR jobs to start
  */
 function getJobsFromPR(config) {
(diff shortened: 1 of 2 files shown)


yoshwata (approved):
LGTM

kkokufud (approved):
👍

tk3fftk:
Do we need add tests?

jithine:
Is this always the case or some corner case ? I've seen a case internally where for a job

```
jobs:
    qa:
        requires: [~pr:qa, ~commit:qa]
```

there are 2 pipelines Pipeline A on branch A & Pipeline B on branch B

A pull request on `qa` branch triggered only one of them

yokawara:
@jithin1987 
This PR fixes PR branch filtering with pipeline's branch. (I fixed the Context to clearify the problem.)
This problem is caused because, we are removing branch when the branch is pipeline's branch in [api](https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/webhooks/index.js#L341), but we don't consider pipeline's branch filtering after that.

In your case, `~pr:A` and `~pr:B` are supposed not to start builds on each pipelines.
> A pull request on qa branch triggered only one of them

~~I have not seen the case, there might be a new bug ... ?~~ I couldn't reproduce the case.
If it happenes regulary, I would like to work on it as another issue.

jithine (approved):

PR #389, opened 2019-09-11
fix: builds list should use createTime as default sortBy

jithine:
## Context

Use `createTime` as default sortBy for builds list, so that it uses index.

## Objective

Address performance issue with `job.builds` list
## References

https://github.com/screwdriver-cd/screwdriver/issues/1737

## License

### DIFF
--- a/test/lib/collectionFactory.test.js
+++ b/test/lib/collectionFactory.test.js
@@ -6,7 +6,7 @@ const sinon = require('sinon');
 
 sinon.assert.expose(assert, { prefix: '' });
 
-describe.only('Collection Factory', () => {
+describe('Collection Factory', () => {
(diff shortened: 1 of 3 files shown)


test/lib/collectionFactory.test.js
```diff
@@ -17,7 +17,8 @@ describe('Collection Factory', () => {
         userId,
         name,
         description,
-        pipelineIds
+        pipelineIds,
+        type: undefined
```
tkyi: I don't think `undefined` is allowed
can you switch this to something else?
https://github.com/screwdriver-cd/data-schema/blob/master/models/collection.js#L49

tkyi (approved):

PR #390, opened 2019-09-18
fix: job.getBuilds should use default pagination

jithine:
## Context

`job.getBuilds()` is expensive for jobs with large number of builds and should always be using with pagination. 

## Objective

Reduce number of data records fetched from DB for `job.getBuilds()`

## References

https://github.com/screwdriver-cd/screwdriver/issues/1737

## License

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -9,6 +9,7 @@ const tokenGen = Symbol('tokenGen');
 const apiUri = Symbol('apiUri');
 const START_INDEX = 3;
 const MAX_COUNT = 1000;
+const DEFAULT_COUNT = 10;
 
 /**
  * Find metrics and step metrics related to the build
@@ -145,6 +146,9 @@ class Job extends BaseModel {
             params: {
                 jobId: this.id
             },
+            paginate: {
+                count: DEFAULT_COUNT
+            },
             sort: sort ? sort.toLowerCase() : 'descending' // Sort by primary sort key
         };
 
@@ -227,8 +231,8 @@ class Job extends BaseModel {
      */
     remove() {
         const removeBuilds = (() =>
-            // Iterate through the builds and remove them
-            this.getBuilds().then((builds) => {
+            // Iterate through the builds and remove MAX_COUNT at a time
+            this.getBuilds({ paginate: { count: MAX_COUNT } }).then((builds) => {
                 if (builds.length === 0) {
                     // Done removing builds
                     return null;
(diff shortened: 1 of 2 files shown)


parthasl (approved):
👍

d2lam:
Don't we need to modify these functions to recursively get the builds: https://github.com/screwdriver-cd/models/blob/master/lib/job.js#L183-L198
otherwise it will miss some data since `DEFAULT_COUNT` is only 10

jithine:
@d2lam 
> Don't we need to modify these functions to recursively get the builds: https://github.com/screwdriver-cd/models/blob/master/lib/job.js#L183-L198
otherwise it will miss some data since DEFAULT_COUNT is only 10

`getLastSuccessfulBuild` is already using only 1 record.   https://github.com/screwdriver-cd/models/blob/master/lib/job.js#L196 So this won't be affected.

As for `getLastSuccessfulBuild` there should be only one record in either `RUNNING` or `QUEUED` (unless someone has chosen to turn off COLLAPSE)  categories..  And these are 2 parallel queries and used only from PR `stopJob` context - https://github.com/search?q=org%3Ascrewdriver-cd+getRunningBuilds&type=Code

tkyi (approved):

d2lam:
> getLastSuccessfulBuild` is already using only 1 record. https://github.com/screwdriver-cd/models/blob/master/lib/job.js#L196 So this won't be affected.

Not necessarily. getBuilds() will try to get the builds first, then look for `SUCCESS`
So if the builds are like `[FAIL, FAIL, FAIL, FAIL, FAIL, FAIL, FAIL, FAIL, FAIL, FAIL, FAIL, FAIL, SUCCESS]`, then `getLastSuccessfulBuild` will return nothing.

jithine:
@d2lam 
>  So if the builds are like [FAIL, FAIL, FAIL, FAIL, FAIL, FAIL, FAIL, FAIL, FAIL, FAIL, FAIL, FAIL, SUCCESS], then getLastSuccessfulBuild will return nothing.

how ? It's filtering by `SUCCESS` https://github.com/screwdriver-cd/models/blob/master/lib/job.js#L195-L196

d2lam:
Yeah actually I'm not sure, it depends on sequelize. Maybe it does take care of it in that order. In that case then it's ok. 

For `getRunningBuilds` case, are we not going to take care of the case where Collapse is off then?

jithine:
> For getRunningBuilds case, are we not going to take care of the case where Collapse is off then?

There won't be more than 1 active build for a job in PR context. Because this is called on PR sync on webhook delivery and since it happens for each webhook delivery, this should happen sequentially and cleanup older builds. 

https://github.com/screwdriver-cd/screwdriver/blob/359b89f3befc47f4a6c331fc402cd93c90030815/plugins/webhooks/index.js#L455-L459

it cleans up first before creating new.

PR #392, opened 2019-09-24
fix: add parameters to top level

adong:
## Context

Allow users to send `{ meta: parameters: {} }` via API, uisng `meta get` to react in their pipelines later. We need to be able to only allow parameters defined in `screwdriver.yaml` to pass thru.

## Objective

This PR does exactly that, pass thru the user defined parameters in `screwdriver.yaml` file down to the `lib/eventFactory.js` to trigger build.

## License

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -215,6 +215,8 @@ function startBuild(config) {
  * @param  {Boolean}  [config.webhooks]                      If the create came from a webhook (pr or push) or not
  * @param  {Boolean}  [config.isPR]                          Is it PR?
  * @param  {Object}   [config.decoratedCommit]               Decorated commit object
+ * @param  {Object}   [config.pipeline]                      Default pipeline config from databse
+ * @param  {Object}   [config.pipelineConfig]                Current Pipeline config
  */
 function createBuilds(config) {
     const {
@@ -237,10 +239,12 @@ function createBuilds(config) {
     // enable pipeline parameters feature
     if (pipeline.parameters && pipelineConfig.meta && pipelineConfig.meta.parameters) {
         const allowedPipelineParameters = _.keys(pipeline.parameters);
-        const currentPipelineParamters = _.assign({}, pipeline.parameters,
+        const currentPipelineParameters = _.assign({},
+            pipeline.parameters,
             pipelineConfig.meta.parameters);
 
-        pipelineConfig.meta.paramters = _.pick(currentPipelineParamters, allowedPipelineParameters);
+        pipelineConfig.meta.parameters = _.pick(currentPipelineParameters,
+            allowedPipelineParameters);
     }
 
     return Promise.all([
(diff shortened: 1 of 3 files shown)


adong:
Rebased with latest master

adong:
Rebased with latest master

lib/eventFactory.js
```diff
@@ -232,6 +233,15 @@ function createBuilds(config) {
         return null;
     }
 
+    // enable pipeline parameters feature
+    if (pipeline.parameters && pipelineConfig.meta && pipelineConfig.meta.parameters) {
+        const allowedPipelineParameters = _.keys(pipeline.parameters);
+        const currentPipelineParamters = _.assign({}, pipeline.parameters,
+            pipelineConfig.meta.parameters);
+
+        pipelineConfig.meta.paramters = _.pick(currentPipelineParamters, allowedPipelineParameters);
```
d2lam: mispelled `parameters` here. 
can you add test for this?

lib/eventFactory.js
```diff
@@ -12,6 +12,7 @@ const {
 } = require('screwdriver-data-schema').config.regex;
 const workflowParser = require('screwdriver-workflow-parser');
 const winston = require('winston');
+const _ = require('lodash');
```
d2lam, line 15: `lodash` is heavy. Since what we do is not too complicated, I would suggest to not include it.

lib/eventFactory.js
adong: Corrected it.

lib/eventFactory.js
adong: I think its okay, since we already have loadsh in the `screwdriver-cd/model`. Plus, its in the backend, I would have some reservation in the front end though.

```javascript
lib/eventFactory.js
15:const _ = require('lodash');

lib/pipeline.js
12:const _ = require('lodash');
```

test/lib/eventFactory.test.js
```diff
@@ -1506,6 +1506,50 @@ describe('Event Factory', () => {
                 });
             });
         });
+
+        it.only('should have parameters if create with parameters', () => {
```
tkyi:
```suggestion
        it('should have parameters if create with parameters', () => {
```

test/lib/eventFactory.test.js
```diff
@@ -1506,6 +1506,50 @@ describe('Event Factory', () => {
...
+            config.startFrom = 'main';
+            config.meta = currentPipelineConfig;
+            syncedPipelineMock.update = sinon.stub().resolves(fixedPipelineConfig);
+
+            return eventFactory.create(config).then((model) => {
+                assert.equal(model.meta.parameters.user, 'actualValue');
+            });
+        });
+
+        it.only('should not have parameters if create without parameters', () => {
```
tkyi:
```suggestion
        it('should not have parameters if create without parameters', () => {
```

adong: Fixed

test/lib/eventFactory.test.js
adong: Fixed

d2lam (approved):

lib/eventFactory.js
DekusDenial: Include the module u needs

PR #393, opened 2019-09-26
feat(1717): get latest build

klu909:
## Context

This is initial work for adding an endpoint that return latest build

## Objective

Modify getLastSuccessfulBuild to accept other status

## References

https://github.com/screwdriver-cd/screwdriver/issues/1717

## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #394, opened 2019-09-27
fix: getLatestBuild return default if no build found

klu909:
## Context

in my previous PR I overlooked case for builds that doesn't exist

## Objective

making change so that getLatestBuild return an empty object when request build doesn't exist

## References

https://github.com/screwdriver-cd/models/pull/393

## License

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -199,7 +199,7 @@ class Job extends BaseModel {
      */
     getLatestBuild(config = {}) {
         return this.getBuilds({ status: config.status })
-            .then(latestBuilds => latestBuilds[0]);
+            .then(latestBuilds => latestBuilds[0] || {});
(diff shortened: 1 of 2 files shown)


lib/job.js
```diff
@@ -199,7 +199,11 @@ class Job extends BaseModel {
      */
     getLatestBuild(config = {}) {
         return this.getBuilds({ status: config.status })
-            .then(latestBuilds => latestBuilds[0]);
+            .then((latestBuilds) => {
+                const latestBuild = latestBuilds[0] ? latestBuilds[0] : {};
```
tkyi:
```suggestion
                return latestBuilds[0] || {};
```

tkyi (approved):

PR #399, opened 2019-10-10
fix: allow the pipeline model's getMetrics method to fetch metrics by pagination

Code-Beast:
## Context
Currently on the frontend the dashboard/show page fetches metrics of last month, which may retrieve nothing for some pipeline that hasn't been triggered for a while and cause pipeline-card component or collection-table-row component to show nothing (because lastEventInfo will be empty too). If we allow pipeline metrics to be fetched by pagination, we can always get the latest metrics info as long as the pipeline has been triggered before.

## Objective
Pipeline model's getMetrics method can fetch metrics by pagination

## References
[Screwdriver-1784](https://github.com/screwdriver-cd/screwdriver/pull/1784)

## License

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -17,6 +17,9 @@ const REGEX_CAPTURING_GROUP = {
     job: 2 // main or undefined if using legacy
 };
 const MAX_COUNT = 1000;
+// The default page for fetching not by aggregatedInteval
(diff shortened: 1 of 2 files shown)


Code-Beast:
Need some test changes

tkyi (approved):
LGTM after adding a test

tkyi (approved):

PR #402, opened 2019-10-21
fix: use name.value instead of name for parameters

adong:
## Context

For this PR, unify the way of parameters defined in the screwdriver and used in the backend. For example:

Before

```
parameters: 
  p1: 'a'
  p2: 
    value: 'b'
    description: 'b description'
```

You can get value 

```
echo $(meta get parameters.p1) # a
echo $(meta get parameters.p2) # object 
```

After

```
echo $(meta get parameters.p1.value) # a
echo $(meta get parameters.p2.value) # b
```

`p1` 
```
parameters: 
  p1: 'a'
```

is now a shorthand of 

```
parameters: 
  p1: 
    value: 'a'
    description: ''
```

and users must use `parameter.name.value` to get its `value`

## Objective

This PR implements above behavior

## License

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -244,17 +244,19 @@ function createBuilds(config) {
             if (typeof val === 'object') {
                 defaultParameterValue = val.value;
             }
-            allowedParameters[name] = defaultParameterValue;
 
+            allowedParameters[name] = { value: defaultParameterValue };
             if (pipelineConfig.meta.parameters) {
-                const actualParameterValue = pipelineConfig.meta.parameters[name];
+                let actualParameterValue = pipelineConfig.meta.parameters[name];
 
-                // handle input as 0, false and other intented falsy value as user input
+                if (typeof actualParameterValue === 'object') {
+                    actualParameterValue = actualParameterValue.value;
+                }
+
                 if (actualParameterValue !== undefined) {
-                    allowedParameters[name] = actualParameterValue;
+                    allowedParameters[name] = { value: actualParameterValue };
                 }
             }
-
         });
 
         if (pipelineConfig.meta) {
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -12,8 +12,6 @@ const {
 } = require('screwdriver-data-schema').config.regex;
 const workflowParser = require('screwdriver-workflow-parser');
 const winston = require('winston');
-const _ = require('lodash');
-
 let instance;
 
 /**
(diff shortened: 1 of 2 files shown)


lib/eventFactory.js
```diff
@@ -236,15 +234,36 @@ function createBuilds(config) {
...
+        const allowedParameters = Object.create(null);
+
+        Object.entries(pipeline.parameters).forEach(([name, val]) => {
+            let defaultParameterValue = val;
+
+            if (typeof val === 'object') {
+                defaultParameterValue = val.value;
+            }
+
+            allowedParameters[name] = { value: defaultParameterValue };
```
jithine, line 248: why do we need to make this an object with key`value` ?
why can't this be simply ` allowedParameters[name] = defaultParameterValue;`

adong: Unified behavior for end users when they access parameters. Users can only access `parameter.user.value`, while `parameter.user` directly is not permitted.

DekusDenial:
may i ask what's the point of this convention? is your key-value mapping only one level deep? if i have more than one level deep parameter mapping, will i be i doing `parameters.a.value.b.value.c.value`? if so, this is weird

adong:
> may i ask what's the point of this convention? is your key-value mapping only one level deep? if i have more than one level deep parameter mapping, will i be i doing `parameters.a.value.b.value.c.value`? if so, this is weird

Yep, it's a flat dictionary, only 1 level deep.

lib/eventFactory.js
tkyi: what if a user wants both `value` and `description`?

jithine: we should have a limit on description. But to be consistent with yaml data, let's keep this proposal.

jithine (approved):

tkyi (approved):

lib/eventFactory.js
```diff
@@ -236,15 +234,36 @@ function createBuilds(config) {
...
+
+                if (actualParameterValue !== undefined) {
+                    allowedParameters[name] = { value: actualParameterValue };
+                }
+            }
+        });
 
-        pipelineConfig.meta.parameters = _.pick(currentPipelineParameters,
-            allowedPipelineParameters);
+        if (pipelineConfig.meta) {
```
d2lam, line 262: Can just do 
```
pipelineConfig.meta = Object.assign({}, pipelineConfig.meta, { parameters: allowedParameters });
```

so you don't need if/else block here

d2lam (approved):
small comment. logic LGTM 👍

lib/eventFactory.js
adong: I actually went back and forth on it, and eventually pick readability over shorter codes. 🤦‍♂️

PR #406, opened 2019-10-31
fix(1801): update collection when removing pipeline take 2

klu909:
## Context

in previous pull request, collection.pipelineIds was not mark dirty thus db is not actually updated

## Objective

assign new array to collection so collection.pipelineIds would be marked as dirty

## License

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -1273,7 +1273,11 @@ class PipelineModel extends BaseModel {
                         collection.pipelineIds.includes(this.id));
 
                     return Promise.all(filteredCollections.map((collection) => {
-                        collection.pipelineIds.splice(collection.pipelineIds.indexOf(this.id), 1);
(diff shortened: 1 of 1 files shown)


lib/pipeline.js
```diff
@@ -1273,7 +1273,11 @@ class PipelineModel extends BaseModel {
                         collection.pipelineIds.includes(this.id));
 
                     return Promise.all(filteredCollections.map((collection) => {
-                        collection.pipelineIds.splice(collection.pipelineIds.indexOf(this.id), 1);
+                        const newPipelineIds = collection.pipelineIds.filter(pipelineId =>
```
adong, line 1276: Nice finding!
https://stackoverflow.com/questions/37805412/remove-an-element-from-an-array-in-javascript-with-slice-or-spread

adong (approved):

jithine (approved):

PR #407, opened 2019-11-04
feat(1610): Add cleanup method in job factory

pritamstyz4ever:
## Context

Currently when the node process terminates it happens abruptly and the in process messages are lost

## Objective

This PR adds a cleanup function

## References

https://github.com/screwdriver-cd/screwdriver/issues/1610

## License

### DIFF
--- a/lib/baseFactory.js
+++ b/lib/baseFactory.js
@@ -211,6 +211,13 @@ class BaseFactory {
     }
 
     /**
+     * Cleanup
+     */
+    cleanUp() {
+        // no-op when not implemted by extender
+    }
+
+    /**
      * Get an instance of a Factory
      * @method getInstance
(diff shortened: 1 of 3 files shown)


lib/baseFactory.js
```diff
@@ -210,6 +210,13 @@ class BaseFactory {
             });
     }
 
+    /**
+     * Cleanup
+     */
+    cleanUp() {
+        // no-op when not implemted by extender
```
DekusDenial, line 217: what else would implement this method? example?

DekusDenial (approved):
small comment

lib/baseFactory.js
pritamstyz4ever: so job factory cleanup method would be called by tasks set in the plugin. Added to base if any other factory classes intend to call this

PR #412, opened 2019-11-26
fix(1339): event not standardizing parameters:

klu909:
## Context

Currently we are standardizing parameters when we create builds but not events. This delays display of parameters until a build finishes.

## Objective

standardizing parameters when we create events

## References

https://github.com/screwdriver-cd/screwdriver/issues/1339

## License

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -234,38 +234,6 @@ function createBuilds(config) {
         return null;
     }
 
-    // enable pipeline parameters feature, if parameters exist in screwdriver.yaml
-    if (pipeline.parameters) {
-        const allowedParameters = Object.create(null);
-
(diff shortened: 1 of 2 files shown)


lib/eventFactory.js
```diff
@@ -411,6 +379,49 @@ function updateWorkflowGraph(config) {
...
+                actualParameterValue = actualParameterValue.value;
+            }
+
+            if (actualParameterValue !== undefined) {
+                allowedParameters[name] = { value: actualParameterValue };
+            }
+        }
+    });
+
+    return Promise.resolve(allowedParameters);
```
jithine: why return promise here ?

klu909: yeah we don't need it, I will change that

jithine (approved):

PR #416, opened 2019-12-10
fix(1339): add null check on evenConfig.meta

klu909:
## Context

in previous pull request I used different config object where one did not replace missing meta with empty object, which lead to internal server error.

## Objective

add null check against it

## References

https://github.com/screwdriver-cd/models/pull/412

## License


parthasl (approved):
👍

PR #417, opened 2019-12-16
feat(1881): build cluster to support multiple scm contexts

parthasl:
## Context

Build cluster supports only one SCM context, which leads to error "This pipeline is not authorized to use this build cluster." if the same build cluster is used for different SCM context.

## Objective

This PR is intended for a build cluster to support multiple SCM contexts

## References

[1881](https://github.com/screwdriver-cd/screwdriver/issues/1881)

## License


DekusDenial (approved):
Though I’m bit confused about whether we should expect an array of strings or just a JSON array string

parthasl:
> Though I’m bit confused about whether we should expect an array of strings or just a JSON array string

array of strings similar to scm organizations

vnugopal (approved):

PR #423, opened 2020-01-10
fix(1865): [1] Remove builds.steps 

kkokufud:
## Context
As mentioned here screwdriver-cd/screwdriver#1865 , `builds.steps` and `steps.command` are both same data.
And these steps are defined by user's yaml and it tends to have a big size.

## Objective
This PR reduces DB size by remove builds.steps in build table.
And did some refactoring.

## References
https://github.com/screwdriver-cd/screwdriver/issues/1865

## License

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -484,27 +484,18 @@ class BuildModel extends BaseModel {
             const now = (new Date()).toISOString();
 
             return this.getSteps().then((steps) => {
-                if (steps.length !== 0) {
-                    return Promise.all(steps.map((step) => {
-                        if (step.startTime && !step.endTime) {
-                            step.endTime = now;
-                            step.code = ABORT_CODE;
-                        }
-
-                        return step.update();
-                    }));
+                if (steps.length === 0) {
+                    return new Promise();
                 }
 
-                this.steps = this.steps.map((step) => {
+                return Promise.all(steps.map((step) => {
                     if (step.startTime && !step.endTime) {
                         step.endTime = now;
                         step.code = ABORT_CODE;
                     }
 
-                    return step;
-                });
-
-                return this.steps;
+                    return step.update();
+                }));
             });
         };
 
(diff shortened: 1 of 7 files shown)


kkokufud:
We will make PR of API and data-schema changes. Please wait.

jithine:
This is awesome. Thanks for addressing this. cc @parthasl

test/lib/buildFactory.test.js
```diff
@@ -844,83 +847,15 @@ describe('Build Factory', () => {
...
-        const stepsData = steps.map(step => Object.assign({ code: 0 }, step));
-        const stepsMock = stepsData.map((step) => {
-            const mock = hoek.clone(step);
-
-            mock.toJson = sinon.stub().returns(step);
-
-            return mock;
-        });
-
         it('should list builds without step models', () => {
```
kkisic: I think this title does not match with these test codes.

kkokufud: I fixed it.

sugarnaoming:
Is it better to call `toJsonWithSteps` function in `toJson` function in `models/build.js` ?

sakka2:
> Is it better to call `toJsonWithSteps` function in `toJson` function in `models/build.js` ?

If do so, `toJson` function becomes to return promise object, it breaks compatibility.

sugarnaoming:
> > Is it better to call `toJsonWithSteps` function in `toJson` function in `models/build.js` ?
> 
> If do so, `toJson` function becomes to return promise object, it breaks compatibility.

I see OK

yuichi10 (approved):
LGTM

lib/job.js
```diff
@@ -19,15 +19,15 @@ const DEFAULT_COUNT = 10;
  * @param  {String}    aggregateInterval  Include step metrics if no aggregation
  * @return {Array}                        Array of metrics
```
tk3fftk: (small comment) this function returns `Promise`.

tk3fftk (approved):
LGTM

lib/job.js
sugarnaoming: thx
fixed

jithine (approved):

lib/build.js
```diff
@@ -594,29 +586,62 @@ class BuildModel extends BaseModel {
...
+            .then((steps) => {
+                if (!steps.length) {
+                    throw new Error('Steps do not exist');
+                }
+
+                // This if statement should be removed after enough time has passed since build.steps removed.
+                // Make orders of steps in completed builds sure,
+                // because steps in old builds in DB have the order not sorted.
+                if (this.endTime) {
+                    steps.sort((a, b) =>
```
jithine, line 638: `because steps in old builds in DB have the order not sorted.` can you elaborate this ? isn't `id` always ascending ? won't `sortBy` guarantee this ?

sugarnaoming: In old builds, the order of records in steps table are not guaranteed. because it used `Promise.all` to save steps records, so each records is inserted asynchronously.
[fixed before](https://github.com/screwdriver-cd/models/blob/d380c78410630d57238d82c95f89a6d0f642c3f4/lib/buildFactory.js#L288)

After this fixed is merged, it guarantees the order.
[fixed after](https://github.com/sonic-screwdriver-cd/models/blob/8284c55cb627880130a81bc047c5d8262c6bbfdf/lib/buildFactory.js#L263)

So this is workaround for old builds.

PR #428, opened 2020-01-17
[Don't Merge]feat(1716): enhance latest build end point

klu909:
## Context

currently latest build end point only returns latest build, we should allow fetching build in relative position to latest build, such as second to latest build.

## Objective

fetching build in relative position to latest build

## References

https://github.com/screwdriver-cd/screwdriver/issues/1716

## License

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -204,7 +204,7 @@ class Job extends BaseModel {
         const parsedPoistion = Number(position);
         let relativePosition = -1;
 
-        if (parsedPoistion && parsedPoistion > 0) {
+        if (parsedPoistion === 0 || parsedPoistion > 0) {
             relativePosition = parsedPoistion;
         } else if (position === 'latestBuild') { // to retain old behavior
             relativePosition = 0;
--- a/lib/job.js
+++ b/lib/job.js
@@ -193,13 +193,25 @@ class Job extends BaseModel {
 
     /**
      * Return latest build that belong to this job
-     * @param  {Object}   [config]                  Configuration object
-     * @param  {String}   [config.status]           Return latest build with this status
-     * @return {Promise}                            Lastest build
+     * @param  {Object}   [config]              Configuration object
+     * @param  {String}   [config.status]       Return build with this status
+     * @param  {String}   [config.position]     Return build with this position relative to latest build
+     * @return {Promise}                        Lastest build
      */
     getLatestBuild(config = {}) {
-        return this.getBuilds({ status: config.status })
-            .then(latestBuilds => latestBuilds[0] || {});
+        const { position, status } = config;
+        // passing string instead of number to retain old behavior
+        const parsedPoistion = Number(position);
+        let relativePosition = -1;
+
+        if (parsedPoistion && parsedPoistion > 0) {
+            relativePosition = parsedPoistion;
+        } else if (position === 'latestBuild') { // to retain old behavior
+            relativePosition = 0;
+        }
+
+        return this.getBuilds({ status })
+            .then(latestBuilds => latestBuilds[relativePosition] || {});
     }
 
     /**
--- a/test/lib/job.test.js
+++ b/test/lib/job.test.js
@@ -371,10 +371,70 @@ describe('Job Model', () => {
                 sort: 'descending'
             };
 
-            return job.getLatestBuild()
+            return job.getLatestBuild({ position: 'latestBuild' })
                 .then((latestBuild) => {
                     assert.calledWith(buildFactoryMock.list, expected);
                     assert.equal(latestBuild, build3);
+                });
+        });
+
+        it('gets second to latest build', () => {
+            buildFactoryMock.list.resolves([build3, build2]);
+
+            const expected = {
+                paginate: {
+                    count: 10
+                },
+                params: {
+                    jobId: 1234
+                },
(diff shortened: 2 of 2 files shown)


lib/job.js
```diff
@@ -193,13 +193,25 @@ class Job extends BaseModel {
...
+     * @param  {String}   [config.status]       Return build with this status
+     * @param  {String}   [config.position]     Return build with this position relative to latest build
+     * @return {Promise}                        Lastest build
      */
     getLatestBuild(config = {}) {
-        return this.getBuilds({ status: config.status })
-            .then(latestBuilds => latestBuilds[0] || {});
+        const { position, status } = config;
+        // passing string instead of number to retain old behavior
+        const parsedPoistion = Number(position);
```
adong, line 204: [`Number`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number) could return `NaN`

```
> Number('123')
123
> Number('x123')
NaN
> Number({a: 123})
NaN
```

So, we need to guard and set a default value for it.

lib/job.js
```diff
@@ -193,13 +193,25 @@ class Job extends BaseModel {
...
+        let relativePosition = -1;
+
+        if (parsedPoistion && parsedPoistion > 0) {
+            relativePosition = parsedPoistion;
+        } else if (position === 'latestBuild') { // to retain old behavior
```
adong, line 209: > `position === 'latestBuild'`

If this is user input that passed from url, could we change `latestBuild` to `latest`? 

Also, I don't have much comment on `position` used here, or its synonyms: order, revision.
Not sure what term people usually use.

@jithine @tkyi Anyone suggestions on the naming?

lib/job.js
```diff
@@ -193,13 +193,25 @@ class Job extends BaseModel {
...
+     * @param  {String}   [config.position]     Return build with this position relative to latest build
+     * @return {Promise}                        Lastest build
      */
     getLatestBuild(config = {}) {
-        return this.getBuilds({ status: config.status })
-            .then(latestBuilds => latestBuilds[0] || {});
+        const { position, status } = config;
+        // passing string instead of number to retain old behavior
+        const parsedPoistion = Number(position);
+        let relativePosition = -1;
```
adong, line 205: Not sure if this helps with the developer experience though. 🤷‍♂ 

```
        // default revision in asscending order
        // i.e. [0, 1, 2, 3, 4, 5]
        // 0 is the latest, which is 0
        // 1 is second-to-last, which is 1
        // -1 is the first, which is 5

        let absolutePosition = 0;

        if (relativePosition === 'latest') {
          absolutePosition = 0;
        } else {
          const parsedPoistion = Number(relativePosition);

          if (!isNaN(parsedPoistion) {
            if (parsedPoistion < 0) {
              // totalNumberOfRevision, not sure if we have it or not.
              absolutePosition = totalNumberOfRevision + parsedPoistion;
            } else {
              absolutePosition = parsedPoistion;
            }
          }
        }
```

adong (changes requested):

lib/job.js
adong:
```
When `parsedPoistion = '0.1'`;

        if (parsedPoistion && parsedPoistion > 0) {
            relativePosition = parsedPoistion;
        }
```

`relativePosition <= (Number(0.01) && Number(0.01) > 0)` would be `0.1`

jithine (commented):
why do we need this logic ?
since it's a filter on top of results returned from DB, can't caller just fetch 10 or so records, sorted descending and then pick required item the list ?

adong:
> why do we need this logic ?
> since it's a filter on top of results returned from DB, can't caller just fetch 10 or so records, sorted descending and then pick required item the list ?

I think this change is needed for a follow-up changes of  
PR: https://github.com/screwdriver-cd/ui/pull/515/files

i.e.

``` 
1) /pipelines/1/job/main/latest
2) /pipelines/1/job/main/latest?status=SUCCESS
3) /pipelines/1/job/main/latest?status=FAILURE
4) /pipelines/1/job/main/3?status=SUCCESS
5) /pipelines/1/job/main/4?status=FAILURE
6) /pipelines/1/job/main/-1?status=SUCCESS
```


To support issue: https://github.com/screwdriver-cd/screwdriver/issues/1717

lib/job.js
klu909: thing like this will just make function return an empty object, we can do a Math.floor but I am not sure if that's what we want

klu909: it's user error IDK if we want to recover from it at all

lib/job.js
klu909: As comment suggest, this is to retain old behavior. So this change does not break any current build that happen to rely on previous behavior. Changing it to latest WILL BREAK IT

lib/job.js
adong: In that case, just throw exception is good enough.

adong (approved):

lib/job.js
adong: Looks like `totalNumberOfRevision` is not gonna be available, negative query is not gonna be implemented for the time being.

klu909:
As per discussion with Jithin, this change was not necessary. Will pull this back for now

PR #429, opened 2020-01-23
fix: prevent updating pipeline for each job

jithine:
## Context

Pipeline update is getting called for each job in the workflow graph. This should happen only once.

## Objective

When adding job id to workflow graph in pipeline, we are currently calling `update` for each node. But this should happen only once.

## References

https://github.com/screwdriver-cd/models/pull/424/files#diff-679e90b5b15ab1b73ddb7ed2377e7413R810-R848

## License

### DIFF
--- a/test/lib/pipeline.test.js
+++ b/test/lib/pipeline.test.js
@@ -384,7 +384,7 @@ describe('Pipeline Model', () => {
         });
     });
 
-    describe('sync', () => {
+    describe.only('sync', () => {
         let publishMock;
         let mainMock;
         let externalMock;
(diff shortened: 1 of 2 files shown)


test/lib/pipeline.test.js
```diff
@@ -384,7 +384,7 @@ describe('Pipeline Model', () => {
         });
     });
 
-    describe('sync', () => {
+    describe.only('sync', () => {
```
tkyi:
```suggestion
    describe('sync', () => {
```

tkyi (approved):

test/lib/pipeline.test.js
jithine: fixed

pritamstyz4ever (approved):

PR #431, opened 2020-01-31
fix(1930): handle case where getPermission throws error

klu909:
## Context

currently screwdriver can't delete user from admin list if they revoke their token.

## Objective

correctly handle the thrown error from getPermission()

## References

https://github.com/screwdriver-cd/screwdriver/issues/1930

## License

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -937,7 +937,11 @@ class PipelineModel extends BaseModel {
                 // eslint-disable-next-line no-await-in-loop
                 permission = await user.getPermissions(this.scmUri);
             } catch (err) {
-                permission.push = false;
+                if (err.status === 401) {
+                    permission.push = false;
+                } else {
+                    throw err;
+                }
             }
 
             if (!permission.push) {
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -931,9 +931,14 @@ class PipelineModel extends BaseModel {
                 username,
                 scmContext: this.scmContext
             });
+            let permission = {};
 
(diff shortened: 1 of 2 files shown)


lib/pipeline.js
```diff
@@ -921,9 +921,14 @@ class PipelineModel extends BaseModel {
...
                 scmContext: this.scmContext
             });
+            let permission = {};
 
-            // eslint-disable-next-line no-await-in-loop
-            const permission = await user.getPermissions(this.scmUri);
+            try {
+                // eslint-disable-next-line no-await-in-loop
+                permission = await user.getPermissions(this.scmUri);
+            } catch (err) {
```
jithine, line 939: can we explicitly check for `401` error ? Wondering if this might cause all admins to be wiped out in case git was down or something.

jithine: `err` object should be an HttpError object ```{"name":"HttpError","status":401,"hea```

klu909: updated it

lib/pipeline.js
```diff
@@ -931,9 +931,18 @@ class PipelineModel extends BaseModel {
...
+            let permission = {};
+
+            try {
+                // eslint-disable-next-line no-await-in-loop
+                permission = await user.getPermissions(this.scmUri);
+            } catch (err) {
+                if (err.status === 401) {
+                    permission.push = false;
+                } else {
+                    throw err;
```
jithine, line 943: don't throw `err` here let it just continue. Else API will never work if user is invalid, or some other random error.

jithine (approved):

PR #432, opened 2020-02-05
feat(1091): add templateId to builds, add templateFactory listWithMetrics method

InderH:
## Context

Need to add templateId to builds so metrics about the number of builds using some template can be aggregated. Need to add functionality so that template versions can be listed along with their associated metrics.

## Objective

This PR adds templateId to build on create. This PR also adds the listWithMetrics method to templateFactory; this method returns the different versions of a template and associated popularity metrics.

## References

https://github.com/screwdriver-cd/screwdriver/issues/1091

## License

### DIFF
--- a/lib/templateFactory.js
+++ b/lib/templateFactory.js
@@ -188,6 +188,45 @@ class TemplateFactory extends BaseFactory {
     }
 
     /**
+     * List templates and associated metrics with pagination and filter options
+     * @method listWithMetrics
+     * @param  {Object}   config                  Config object
+     * @param  {Object}   config.params           Parameters to filter on
+     * @param  {Object}   config.paginate         Pagination parameters
+     * @param  {Number}   config.paginate.count   Number of items per page
+     * @param  {Number}   config.paginate.page    Specific page of the set to return
+     * @return {Promise}
+     */
+    listWithMetrics(config) {
+        return this.list(config).then((templates) => {
+            //make this query the read only data store
+            const JobFactory = require('./jobFactory');
+            const JobFactoryInstance = JobFactory.getInstance();
+
+            const templateIds = templates.map((t) => {return t.id});
+
+            const listConfig = {
+                params: { templateId: templateIds },
+                readOnly: true
+            };
+
+            return JobFactoryInstance.list(listConfig).then((jobs) => {
+                return templates.map((t) => {
+                    const numJobs = jobs.filter(j => j.templateId != undefined && j.templateId === t.id).length;
+
+                    t.metrics = {
+                        jobs: {
+                            count: numJobs
+                        }
+                    }
+
+                    return t;
+                });
+            });
+        });
+    }
+
+    /**
      * Get a template
      * @method get
      * @param  {Mixed}     config
--- a/lib/templateFactory.js
+++ b/lib/templateFactory.js
@@ -215,24 +215,25 @@ class TemplateFactory extends BaseFactory {
 
             const listConfig = {
                 params: { templateId: templateIds },
-                readOnly: true
+                readOnly: true,
+                aggregationField: 'templateId'
             };
 
             return Promise.all([
                 jobFactory.list(listConfig),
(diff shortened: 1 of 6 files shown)


test/lib/templateFactory.test.js
```diff
@@ -445,6 +459,234 @@ describe('Template Factory', () => {
...
+                        },
+                        builds: {
+                            count: 3
+                        }
+                    }
+                }
+            ];
+        });
+
+        /* it('should list templates with metrics when namespace is passed in', () => {
```
klu909: remove commented out code

lib/templateFactory.js
```diff
@@ -187,6 +187,64 @@ class TemplateFactory extends BaseFactory {
...
+
+            const templateIds = templates.map(t => t.id);
+
+            const listConfig = {
+                params: { templateId: templateIds },
+                readOnly: true
+            };
+
+            return jobFactory.list(listConfig).then(jobs =>
+                buildFactory.list(listConfig).then(builds =>
```
klu909: you should do it this way
`return Promise.all([
  jobFactory.list(listConfig),
  buildFactory.list(listConfig)
]).then(([jobs, builds]) => {`

lib/templateFactory.js
```diff
@@ -187,6 +187,64 @@ class TemplateFactory extends BaseFactory {
...
+                    templates.map((t) => {
+                        const numJobs = jobs.filter(
+                            j => j.templateId !== undefined &&
```
klu909: this line is redundant, line 226 will take care of undefined case.

klu909 (changes requested):

test/lib/templateFactory.test.js
```diff
@@ -445,6 +459,234 @@ describe('Template Factory', () => {
...
+                    assert.deepEqual(t.id, expected[i].id);
+                    assert.deepEqual(t.metrics.jobs.count, expected[i].metrics.jobs.count);
+                    assert.deepEqual(t.metrics.builds.count, expected[i].metrics.builds.count);
+                    i += 1;
+                });
+            });
+        }); */
+
+        it('should list templates with metrics when no namespace is passed in', () => {
+            expected = [returnValue[3]];
```
klu909, line 602: it's better to add this line assert.expect(x); with x being number of asserts in your test

InderH: assert.expect() is not a function in this repo

lib/templateFactory.js
InderH: updated, thanks

lib/templateFactory.js
InderH: updated, thanks

test/lib/templateFactory.test.js
InderH: uncommented the test, thanks

klu909 (approved):

lib/templateFactory.js
```diff
@@ -187,6 +187,60 @@ class TemplateFactory extends BaseFactory {
...
+                params: { templateId: templateIds },
+                readOnly: true
+            };
+
+            return Promise.all([
+                jobFactory.list(listConfig),
+                buildFactory.list(listConfig)]
+            ).then(([jobs, builds]) =>
+                templates.map((t) => {
+                    const numJobs = jobs.filter(j => j.templateId === t.id).length;
```
jithine: why do we need this check ? Also if we are looking for just count, can't we leverage sequalize/DB to issue a count rather than doing this in API ?

InderH: Some templates have 50+ versions, so rather than doing 50 count queries, I query for jobs where the templateId is in a list of templateIds (ids for the different template versions) and then filter the list of jobs to get the appropriate counts for each individual template version.

InderH: Updated to do an aggregation query, should significantly reduce API computation.

jithine (approved):

PR #434, opened 2020-02-13
[DNM]fix(1865): correct sort order when steps are skipped

jithine:
## Context

As part of fix for https://github.com/screwdriver-cd/screwdriver/issues/1865 we added sorting to handle older builds. But this causes problem with failed bulids where step `startTime` is not defined.

## Objective

For steps which do not have `startTime`, re-use `startTime` from previous step. For builds after the DB change for dropping `build.steps` this will return correct information. But for builds prior to DB change, this could return out of order results, since we don't have the order maintained anymore. While we could lookup `job` table for grabbing proper step order, that will be an overkill just to handle failed builds prior to DB change.

## References

https://github.com/screwdriver-cd/screwdriver/issues/1865
https://github.com/screwdriver-cd/models/pull/423/files#diff-26b681bd4dac10f25d5293c7595c435bR634-R641

## License

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -634,10 +634,19 @@ class BuildModel extends BaseModel {
                 // This if statement should be removed after enough time has passed since build.steps removed.
                 // Make orders of steps in completed builds sure,
                 // because steps in old builds in DB have the order not sorted.
+
                 if (this.endTime) {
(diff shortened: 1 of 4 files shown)


jithine:
This is not happening on newer builds after DB change in node:12

PR #437, opened 2020-02-20
fix(1930): handle case when using private scm instance

kumada626:
## Context
When we using private scm instance and being removed first admin for pipeline, scm returns `404` error instead of `401` error.
We should handle `404` error too.

## Objective

correctly handle the thrown `404` error from getPermission()

## References

#431 
screwdriver-cd/screwdriver#1930

## License

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -929,7 +929,7 @@ class PipelineModel extends BaseModel {
      */
     async getFirstAdmin() {
         const newAdmins = this.admins;
-        const handleStatus = [401, 404];
+        const handleStatuses = [401, 404];
         /* eslint-disable global-require */
         const UserFactory = require('./userFactory');
         /* eslint-enable global-require */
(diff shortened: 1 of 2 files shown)


wahapo (approved):

yoshwata (approved):

ibu1224 (approved):

lib/pipeline.js
```diff
@@ -929,7 +929,7 @@ class PipelineModel extends BaseModel {
      */
     async getFirstAdmin() {
         const newAdmins = this.admins;
-
+        const handleStatuses = [401, 404];
```
jithine: can you define this `const` outside of this function ? Also call this perhaps `SCM_NO_ACCESS_STATUSES`...

tkyi (approved):
LGTM once Jithin's comments are addressed

lib/pipeline.js
kumada626: Fixed.

jithine (approved):

PR #440, opened 2020-03-05
feat(1785): Save groupEventId to event [2]

tkyi:
## Context

When an event is created, the groupEventId should be saved to track the master parent event id.

## Objective

This PR saves the `groupEventId` to the event if it is passed in. If it is not passed in, it will update the `groupEventId` as its own event `id` after the event is created.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/1785

## License

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -481,9 +481,11 @@ class EventFactory extends BaseFactory {
      *                                              Optional for backwards compatibility
      * @param  {String}  [config.causeMessage]      Message that describes why the event was created
      * @param  {Object}  [config.creator]           Creator of the event
-     * @param  {String}  [config.parentBuildId]     Id of the build that starts this event
-     * @param  {String}  [config.parentEventId]     Id of the parent event
-     * @param  {String}  [config.workflowGraph]     workflowGraph of parentEvent if there is a parentEvent
+     * @param  {Number}  [config.parentBuildId]     Id of the build that starts this event
+     * @param  {Object}  [config.parentBuilds]      Parent builds information
+     * @param  {Number}  [config.groupEventId]      Group parent event ID
+     * @param  {Number}  [config.parentEventId]     Id of the parent event
+     * @param  {Object}  [config.workflowGraph]     workflowGraph of parentEvent if there is a parentEvent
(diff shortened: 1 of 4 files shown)


lib/eventFactory.js
```diff
@@ -519,6 +521,10 @@ class EventFactory extends BaseFactory {
         let prevChainPR = '';
         let decoratedCommit;
 
+        if (groupEventId) {
```
klu909, line 524: we will need to add a logic to set groupEventId to current event's id if groupEventId was not passed in

klu909 (changes requested):

lib/eventFactory.js
tkyi: I think it's okay to leave `groupEventId` blank if it should be its own `eventId`.
Unless you think it makes more sense to try to update the event immediately after creation everywhere.

tkyi: Other logic for setting groupEventId is done in screwdriver repo: https://github.com/screwdriver-cd/screwdriver/pull/2010

klu909: main reason I am bringing this up is that we should be able to query all event in same group by just using groupEventId, otherwise we would need to query second time to fetch root event.

tkyi: yeah that makes sense

If we want to update the newly created event with `groupEventId`, I don't think it can be done in the eventFactory model
we would need to duplicate that logic anytime we called `eventFactory.create()`

klu909: since we have newly created `event` here https://github.com/screwdriver-cd/models/blob/9a3313761c94a627b500ac4c71ea6fbab1ff0eb7/lib/eventFactory.js#L637

can we add `groupEventId`  here if it doesn't exist and call `update()`

tkyi: I tried that but I get 
```
TypeError: this[datastore].update is not a function
```

klu909: I think we need to mock update() in test

klu909 (approved):

lib/eventFactory.js
tkyi: updated

PR #441, opened 2020-03-11
feat(1967): add ability to do raw query

InderH:
## Context

Need to add the ability to do raw queries to baseFactory so queries that are otherwise impossible to do through sequelize can be done. Need to add ability to get build status history for jobs in bulk so that they can be displayed in pipeline list view.

## Objective

Add query method to baseFactory. Add getBuildStatuses to buildFactory so build status history for jobs can be queried in bulk.

## References

https://github.com/screwdriver-cd/screwdriver/issues/1967

## License

### DIFF
--- a/lib/baseFactory.js
+++ b/lib/baseFactory.js
@@ -221,6 +221,52 @@ class BaseFactory {
     }
 
     /**
+     * Run raw query
+     * @method query
+     * @param  {Object}         [config]
+     * @param  {Array<Object>}  [config.queries]      Map of database type to query
+     * @param  {Object}         [config.replacements] Parameters to replace in the query
+     * @param  {Boolean}        [config.rawResponse]  Return raw response without binding to model
+     * @param  {Boolean}        [config.readOnly]     Use readOnly datastore
+     * @return {Promise}
+     */
+    query(config) {
+        const queryConfig = {
+            table: this.table,
+            queries: hoek.reach(config, 'queries'),
+            replacements: hoek.reach(config, 'replacements', { default: {}}),
+            rawResponse: hoek.reach(config, 'rawResponse', { default: false})
+        };
+
+        let datastoreForLookup = this.datastore;
+        // use read only datastore if it is configured
+        const readOnly = hoek.reach(config, 'readOnly', { default: false });
+
+        if (readOnly) {
+            datastoreForLookup = this.datastoreRO;
+        }
+
+        return datastoreForLookup.query(queryConfig)
+            .then((data) => {
+
+                if (queryConfig.rawResponse) {
+                    return data;
+                }
+
+                const result = [];
+
+                data.forEach((item) => {
+                    item.datastore = datastoreForLookup;
+                    item.scm = this.scm;
+
+                    result.push(this.createClass(item));
+                });
+
+                return result;
+            });
+    }
+
+    /**
      * Cleanup
      */
     cleanUp() {
--- a/lib/baseFactory.js
+++ b/lib/baseFactory.js
@@ -243,7 +243,7 @@ class BaseFactory {
 
         // use read only datastore if it is configured
         const readOnly = hoek.reach(config, 'readOnly', { default: false });
-        let datastoreForLookup = readOnly ? this.datastoreRO : this.datastore;
+        const datastoreForLookup = readOnly ? this.datastoreRO : this.datastore;
 
         return datastoreForLookup.query(queryConfig).then(data => {
             let formattedData = data;
--- a/lib/baseFactory.js
+++ b/lib/baseFactory.js
@@ -250,10 +250,10 @@ class BaseFactory {
         }
 
         return datastoreForLookup.query(queryConfig).then(data => {
-            let toReturn = data;
+            let formattedData = data;
 
             if (!queryConfig.rawResponse) {
-                toReturn = data.map(item => {
+                formattedData = data.map(item => {
                     item.datastore = datastoreForLookup;
                     item.scm = this.scm;
 
@@ -261,7 +261,7 @@ class BaseFactory {
                 });
             }
 
-            return toReturn;
+            return formattedData;
         });
     }
 
--- a/lib/baseFactory.js
+++ b/lib/baseFactory.js
@@ -234,8 +234,8 @@ class BaseFactory {
         const queryConfig = {
             table: this.table,
             queries: hoek.reach(config, 'queries'),
-            replacements: hoek.reach(config, 'replacements', { default: {}}),
-            rawResponse: hoek.reach(config, 'rawResponse', { default: false})
+            replacements: hoek.reach(config, 'replacements', { default: {} }),
+            rawResponse: hoek.reach(config, 'rawResponse', { default: false })
         };
 
         let datastoreForLookup = this.datastore;
@@ -248,7 +248,6 @@ class BaseFactory {
 
         return datastoreForLookup.query(queryConfig)
             .then((data) => {
-
                 if (queryConfig.rawResponse) {
                     return data;
                 }
--- a/lib/baseFactory.js
+++ b/lib/baseFactory.js
@@ -241,13 +241,9 @@ class BaseFactory {
             rawResponse: hoek.reach(config, 'rawResponse', { default: false })
         };
 
-        let datastoreForLookup = this.datastore;
         // use read only datastore if it is configured
         const readOnly = hoek.reach(config, 'readOnly', { default: false });
-
-        if (readOnly) {
-            datastoreForLookup = this.datastoreRO;
-        }
+        let datastoreForLookup = readOnly ? this.datastoreRO : this.datastore;
 
         return datastoreForLookup.query(queryConfig).then(data => {
             let formattedData = data;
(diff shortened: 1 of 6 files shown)


lib/baseFactory.js
```diff
@@ -220,6 +220,51 @@ class BaseFactory {
...
+            datastoreForLookup = this.datastoreRO;
+        }
+
+        return datastoreForLookup.query(queryConfig)
+            .then((data) => {
+                if (queryConfig.rawResponse) {
+                    return data;
+                }
+
+                const result = [];
```
pritamstyz4ever: Can we use .map() instead of doing the for each?

lib/buildFactory.js
```diff
@@ -295,6 +295,72 @@ class BuildFactory extends BaseFactory {
...
+     * @param  {Number}   config.offset           Number of build statuses to skip
+     * @param  {Number}   config.numBuilds        Number of build statuses to return per job
+     * @return {Promise}
+     */
+    getBuildStatuses(config) {
+        config.offset = config.offset || 0;
+        config.numBuilds = config.numBuilds || 1;
+        config.jobIds = config.jobIds || [];
+
+        const query =
```
pritamstyz4ever: Use string literals instead of concatenating the query

pritamstyz4ever: Also, thinking would it make sense to have a separate file for raw queries which can be required here. Would be easy to compare when we do several changes over the time and also more maintainable.

InderH: If I do a string literal the string exceeds the maximum lint line length, when I tried to do a multi line string I got another lint issue about multi line strings only being supported in es6+.

Is there another way to do this or should I disable the linter for the string?

parthasl (dismissed):
👍

lib/rawQueries.js
```diff
@@ -0,0 +1,23 @@
...
+        WHERE rank > :offset AND rank <= :maxRank
+        ORDER BY r."jobId", r."id" DESC`,
+
+    statusesQueryMySql:
+        `SELECT r.id, r.jobId, j.name as \`jobName\`, r.status, r.startTime, r.endTime
+        FROM (SELECT id, jobId, status, startTime, endTime,
+        RANK() OVER ( PARTITION BY jobId ORDER BY id DESC ) AS \`rank\`
+        FROM builds WHERE jobId in (:jobIds)) r
+        INNER JOIN jobs as j ON j.id = r.jobId
+        WHERE \`rank\` > :offset AND \`rank\` <= :maxRank
```
parthasl, line 15: @InderH for mysql we need ` prefix ?

parthasl: ignore, got it

parthasl:
@InderH - may I know the frequency of this query? this read-only query will be routed to RO instance?

InderH:
> @InderH - may I know the frequency of this query? this read-only query will be routed to RO instance?

@parthasl - This query will run when:
1. the list view is loaded
2. when the list view is scrolled down and there is more to load
3. when the list view's refresh button is pressed

Yes, the query is done with readOnly set to true, so it uses the RO db
https://github.com/screwdriver-cd/models/pull/441/files#diff-43e1dd855c9ebdae88cbf0102c6e05b2R330
https://github.com/screwdriver-cd/models/pull/441/files#diff-c6dfbb25b49dd3e823a2aa3a6f797f96R249

lib/baseFactory.js
```diff
@@ -223,6 +223,48 @@ class BaseFactory {
...
+        let datastoreForLookup = this.datastore;
+        // use read only datastore if it is configured
+        const readOnly = hoek.reach(config, 'readOnly', { default: false });
+
+        if (readOnly) {
+            datastoreForLookup = this.datastoreRO;
+        }
+
+        return datastoreForLookup.query(queryConfig).then(data => {
+            let toReturn = data;
```
tkyi: maybe a better var name than `toReturn`?

tkyi (changes requested):
can you update https://github.com/screwdriver-cd/models/blob/master/README.md#build-factory

lib/baseFactory.js
InderH: What do you think of something like 'formattedData'?

adong: Good call @tkyi 
Yeah, @InderH that `formattedData` would make more sense.

tkyi (approved):

lib/buildFactory.js
```diff
@@ -306,6 +307,51 @@ class BuildFactory extends BaseFactory {
...
+     * Get build statuses for jobs
+     * @method getBuildStatuses
+     * @param  {Object}   config                  Config object
+     * @param  {Array}    config.jobIds           Jobs to get build statuses for
+     * @param  {Number}   config.offset           Number of build statuses to skip
+     * @param  {Number}   config.numBuilds        Number of build statuses to return per job
+     * @return {Promise}
+     */
+    getBuildStatuses(config) {
+        config.offset = config.offset || 0;
```
tkyi: probably shouldn't override function params like this
might be better to create a new var like this:
```suggestion
        const offset = config.offset || 0;
```

InderH: Done, thanks!

tkyi (dismissed):
👍

lib/baseFactory.js
```diff
@@ -223,6 +223,48 @@ class BaseFactory {
...
+            rawResponse: hoek.reach(config, 'rawResponse', { default: false })
+        };
+
+        let datastoreForLookup = this.datastore;
+        // use read only datastore if it is configured
+        const readOnly = hoek.reach(config, 'readOnly', { default: false });
+
+        if (readOnly) {
+            datastoreForLookup = this.datastoreRO;
+        }
```
pritamstyz4ever:
```suggestion
        // use read only datastore if it is configured
        const readOnly = hoek.reach(config, 'readOnly', { default: false });
        let datastoreForLookup = readOnly ? this.datastoreRO : this.datastore;
```

pritamstyz4ever (dismissed):

lib/baseFactory.js
InderH: Done, thanks!

test/lib/buildFactory.test.js
```diff
@@ -1005,4 +1007,175 @@ describe('Build Factory', () => {
...
+                offset: 1,
+                numBuilds: 5
+            };
+
+            returnValue = [
+                [
+                    {
+                        jobId: 1,
+                        jobName: 'name',
+                        Status: 'SUCCESS',
```
tkyi: should be lowercase
```suggestion
                        status: 'SUCCESS',
```

test/lib/buildFactory.test.js
```diff
@@ -1005,4 +1007,175 @@ describe('Build Factory', () => {
...
+            sinon.stub(BuildFactory.prototype, 'query').returns();
+
+            config = {
+                jobIds: [1, 2, 3, 4],
+                offset: 1,
+                numBuilds: 5
+            };
+
+            returnValue = [
+                [
```
tkyi, line 1027: also is this the correct return data type? an array of arrays?

test/lib/buildFactory.test.js
adong: I wonder this too.

test/lib/buildFactory.test.js
adong: @InderH

test/lib/buildFactory.test.js
InderH: updated, thanks!

test/lib/buildFactory.test.js
InderH: yup, it should be an array of arrays - when running a raw query on sequelize without having it map the returning data to a model (which in this case we don't), it returns an array with 2 elements, the first being an array of the data/rows, and the second being meta data.

tkyi: hm but the expected query config has `table: 'builds'`, right?

InderH: Yeah it has `table: 'builds'` but also has `rawResponse: true` - we are only getting a small subset of the data a build has and are using the data in a way which casting it to a build model wouldn't help, so having the query response map back to builds model would just be extra unnecessary computation

tkyi: I see
sounds reasonable

tkyi (approved):

lib/rawQueries.js
adong: I just used a formatter to read, its essentially 
```sql
SELECT "id", 
       "jobid", 
       "status", 
       "starttime", 
       "endtime" 
FROM   (SELECT "id", 
               "jobid", 
               "status", 
               "starttime", 
               "endtime", 
               Rank() 
                 OVER ( 
                   partition BY "jobid" 
                   ORDER BY "id" DESC ) AS rank 
        FROM   builds 
        WHERE  "jobid" IN ( :jobIds )) AS R 
WHERE  rank > :offset 
       AND rank <= :maxRank 
ORDER  BY "jobid", 
          "id" DESC 
```

PR #442, opened 2020-03-19
fix: Step order is broken

ibu1224:
## Context

I found a build where the order of the steps was broken.
It happens several times a day.

image

The expected steps order are: `1 2 3 .... 5 6 7`.
The cause probably occurs when `stepFactory.create` is delayed.

## Objective
In this PR, modify to wait for `stepFactory.create`.

## References

https://stackoverflow.com/questions/41243468/javascript-array-reduce-with-async-await

## License

### DIFF
--- a/lib/buildFactory.js
+++ b/lib/buildFactory.js
@@ -275,7 +275,7 @@ class BuildFactory extends BaseFactory {
                             // eslint-disable-next-line no-await-in-loop
                             await stepFactory.create({ buildId: build.id, ...step });
                         } catch (err) {
-                            logger.error(`Error in steps -BuildId:${build.id}-step:${step}`, err);
+                            logger.error(`Error in BuildFactory create -buildId:${build.id}-step:${step}`, err);
                         }
                     }
 
--- a/lib/buildFactory.js
+++ b/lib/buildFactory.js
@@ -275,7 +275,7 @@ class BuildFactory extends BaseFactory {
                             // eslint-disable-next-line no-await-in-loop
                             await stepFactory.create({ buildId: build.id, ...step });
                         } catch (err) {
-                            logger.error(`Error in BuildFactory create -buildId:${build.id}-step:${step}`, err);
+                            logger.error(`Error in BuildFactory create - buildId:${build.id}-step:${step}`, err);
                         }
                     }
 
--- a/lib/buildFactory.js
+++ b/lib/buildFactory.js
@@ -274,8 +274,8 @@ class BuildFactory extends BaseFactory {
                             // eslint-disable-next-line no-await-in-loop
                             await stepFactory.create({ buildId: build.id, ...step });
                         } catch (err) {
-                            throw new Error(
-                                `Error in steps -BuildId:${build.id}-step:${step}`, err);
+                            // eslint-disable-next-line max-len
+                            throw new Error(`Error in steps -BuildId:${build.id}-step:${step}`, err);
                         }
                     }
 
--- a/lib/buildFactory.js
+++ b/lib/buildFactory.js
@@ -263,13 +263,16 @@ class BuildFactory extends BaseFactory {
 
                     const build = await super.create(modelConfig);
 
-                    await steps.reduce(async (p, step) =>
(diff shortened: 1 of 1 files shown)


lib/buildFactory.js
```diff
@@ -263,8 +263,9 @@ class BuildFactory extends BaseFactory {
 
                     const build = await super.create(modelConfig);
 
-                    await steps.reduce((p, step) =>
-                        p.then(() => stepFactory.create(Object.assign(
+                    await steps.reduce(async (p, step) =>
+                        // eslint-disable-next-line no-return-await
+                        await p.then(() => stepFactory.create(Object.assign(
```
kumada626: Don't we have to await `stepFactory.create() `?

jithine: perhaps switch to `for of` for ensuring order. We do something similar here https://github.com/screwdriver-cd/screwdriver/pull/2011/files#diff-7a514e440aa937beee722c7ac84dfb93R1263-R1271

ibu1224: @jithine 
It is a similar case.
Modify it using the `for of`.

lib/buildFactory.js
```diff
@@ -268,10 +268,16 @@ class BuildFactory extends BaseFactory {
...
-                        Promise.resolve()
-                    );
+                    // eslint-disable-next-line no-restricted-syntax
+                    for (const step of steps) {
+                        try {
+                            // eslint-disable-next-line no-await-in-loop
+                            await stepFactory.create({ buildId: build.id, ...step });
+                        } catch (err) {
+                            // eslint-disable-next-line max-len
+                            throw new Error(`Error in steps -BuildId:${build.id}-step:${step}`, err);
```
kumada626: I think it is also better to log these message using `logger.error`.

ibu1224: ok. I modify to using `logger`

wahapo (approved):

yoshwata (approved):

lib/buildFactory.js
```diff
@@ -268,10 +269,15 @@ class BuildFactory extends BaseFactory {
...
-                        (p, step) => p.then(() => stepFactory.create({ buildId: build.id, ...step })),
-                        Promise.resolve()
-                    );
+                    // eslint-disable-next-line no-restricted-syntax
+                    for (const step of steps) {
+                        try {
+                            // eslint-disable-next-line no-await-in-loop
+                            await stepFactory.create({ buildId: build.id, ...step });
+                        } catch (err) {
+                            logger.error(`Error in steps -BuildId:${build.id}-step:${step}`, err);
```
tkyi: Could have error message something like below to be more consistent with https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/builds/index.js#L1346-L1347, but not a big deal if we leave it as-is

```suggestion
                            logger.error(`Error in buildFactory create steps - buildId:${build.id}-step:${step}`, err);
```

tkyi (approved):
LGTM
small suggestion

lib/buildFactory.js
ibu1224: Thank you review.
I modify error message.

lib/buildFactory.js
```diff
@@ -268,10 +269,15 @@ class BuildFactory extends BaseFactory {
...
-                        (p, step) => p.then(() => stepFactory.create({ buildId: build.id, ...step })),
-                        Promise.resolve()
-                    );
+                    // eslint-disable-next-line no-restricted-syntax
+                    for (const step of steps) {
+                        try {
+                            // eslint-disable-next-line no-await-in-loop
+                            await stepFactory.create({ buildId: build.id, ...step });
+                        } catch (err) {
+                            logger.error(`Error in BuildFactory create -buildId:${build.id}-step:${step}`, err);
```
jithine: can you add a space after first `-` ?
```Error in BuildFactory create - buildId```

ibu1224: OK. I add space in front of `buildId`.

kumada626 (approved):

PR #444, opened 2020-04-03
fix: pass jobId and pipelineId

pritamstyz4ever:
## Context

JobId and PipelineId were undefined in case of periodic builds in log message

## Objective

This PR passes jobId and pipelineId in function to log correctly

## License

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -223,6 +223,8 @@ class Job extends BaseModel {
                         job: newJob,
                         tokenGen: this[tokenGen],
                         apiUri: this[apiUri],
+                        jobId: newJob.id,
+                        pipelineId: pipeline.id,
                         isUpdate: true
                     });
                 })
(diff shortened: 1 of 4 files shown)


klu909 (approved):

tkyi (changes requested):
Shouldn't the job id and the pipeline id already be determined from the `pipeline` and `job` objects?
calling `postBuildEvent` in `startPeriodic`: https://github.com/screwdriver-cd/queue-service/blob/master/plugins/queue/put.js#L156
pipelineId: https://github.com/screwdriver-cd/queue-service/blob/master/plugins/queue/put.js#L25
jobId: https://github.com/screwdriver-cd/queue-service/blob/master/plugins/queue/put.js#L30

jithine:
I think this change is to satisfy the logger arguments, which is no not ideal https://github.com/screwdriver-cd/executor-queue/blob/master/index.js#L193, we could change the logger arguments instead.

tkyi:
Could the logger extrapolate the data from the config? Like 
`..${config.pipeline.id}:${config.job.id}`

pritamstyz4ever:
Fixed in executor-queue.  https://github.com/screwdriver-cd/executor-queue/pull/86
Closing PR

PR #445, opened 2020-04-06
fix: Fix job remove

yuichi10:
## Context
Recently, the screwdriver which we operate got the situation that worker get busy when big pipelines have been deleted.
Now `job.remove` call `executor.stop` to all builds, even if the build was finished and it will make worker busy.

## Objective
`job.remove` will not call `executor.stop` to builds which have 'ABORTED', 'FAILURE' or 'SUCCESS' status.

## License

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -244,17 +244,22 @@ class Job extends BaseModel {
                         return pipeline.id;
                     }
 
-                    return Promise.all(
-                        builds.map(build => {
-                            this[executor].stop({
-                                buildId: build.id,
-                                pipelineId: pipeline.id,
-                                jobId: this.id
-                            });
+                    return Promise.all([
+                        builds
+                            .filter(build => {
+                                const stoppedStatuses = ['ABORTED', 'FAILURE', 'SUCCESS'];
 
-                            return build.remove();
-                        })
-                    )
+                                return !stoppedStatuses.includes(build.status);
+                            })
+                            .map(build =>
+                                this[executor].stop({
+                                    buildId: build.id,
+                                    pipelineId: pipeline.id,
+                                    jobId: this.id
+                                })
+                            ),
+                        builds.map(build => build.remove())
+                    ])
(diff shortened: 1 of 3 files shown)


lib/job.js
```diff
@@ -246,11 +246,15 @@ class Job extends BaseModel {
 
                     return Promise.all(
                         builds.map(build => {
-                            this[executor].stop({
-                                buildId: build.id,
-                                pipelineId: pipeline.id,
-                                jobId: this.id
-                            });
+                            const stoppedStatuses = ['ABORTED', 'FAILURE', 'SUCCESS'];
```
tk3fftk: Do we need to add `UNSTABLE` and `COLLAPSED` ? These are also kind of finished status.

tk3fftk: `UNSTABLE` seems not to be only finished status.
https://github.com/screwdriver-cd/ui/blob/0111bc864056fca3d53c5461e9576e65d45be7f3/app/utils/build.js#L11

pritamstyz4ever (approved):

jithine:
can we build a bulk end point where all builds can be stopped based on ` jobId` ?

lib/job.js
jithine: we should skip unstable, since buld status can be set as unstable in the middle of a build and still have rest of steps running.

jithine: collapsed should be added though

sakka2: I added `COLLAPSED` status.

sakka2:
@jithine 
> can we build a bulk end point where all builds can be stopped based on ` jobId` ?

Thanks, for review. we have some questions about this.
- Why do you want to add the end point? Do you want to remove all job data in redis server which is left unintentionally?
- Where do we add this end point? Is it in queue-service?
- Can we add this end point in other PR?

tkyi (approved):

pritamstyz4ever (approved):

PR #448, opened 2020-05-07
fix(2056): Add webhook after create child pipeline

kumada626:
## Context

We should add webhook after create child pipeline.

## Objective

Add webhook and sync the child pipeline after create it.

## References

screwdriver-cd/screwdriver#2056

## License

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -24,6 +24,7 @@ const SCM_NO_ACCESS_STATUSES = [401, 404];
 const { getAllRecords, getBuildClusterName } = require('./helper.js');
 
 const JOB_CHUNK_SIZE = process.env.JOBS_PARALLEL_COUNT || 5;
+const SD_API_URI = process.env.URI;
 
 // Process jobs JOB_CHUNK_SIZE at at time
 const getJobChunks = jobs => _.chunk(jobs || [], JOB_CHUNK_SIZE);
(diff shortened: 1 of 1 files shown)


wahapo (approved):

yoshwata (approved):

ibu1224 (approved):

lib/pipeline.js
```diff
@@ -639,6 +640,11 @@ class PipelineModel extends BaseModel {
             })
             .then(p => {
                 logger.info(`pipelineId:${this.id}: Creating child pipeline for ${scmUrl} with pipelineId:${p.id}.`);
+                // sync pipeline to create jobs
+                p.sync();
```
tkyi: does it need `await`?

lib/pipeline.js
```diff
@@ -639,6 +640,11 @@ class PipelineModel extends BaseModel {
...
+                if (SD_API_URI) {
+                    p.addWebhook(`${SD_API_URI}/v4/webhooks`);
```
tkyi: same here
does it also need `await`?

jithine: were we not adding webhooks before ? I wonder how it's working in our setup 🤔

lib/pipeline.js
kumada626: Yes, we should add `await`. I fixed it.

lib/pipeline.js
kumada626: Currently, child pipelines are only created or updated when the parent pipeline is synced.
https://github.com/screwdriver-cd/models/blob/56e3ce9b02c68b99ca5848f3e9e029702f742f11/lib/pipeline.js#L730
However, child pipeline's webhooks are not added in this method.
The parent pipeline's webhook is added after it is created, but the child pipeline's ones are not.
https://github.com/screwdriver-cd/screwdriver/blob/cd5c538ba159f0229b7307f491eb4000892a47f4/plugins/pipelines/create.js#L120
So we should add webhooks after child pipeline is created.

tkyi (approved):

PR #449, opened 2020-05-11
fix: do not ignore failure to fetch screwdriver.yaml

jithine:
## Context

Error fetching `screwdriver.yaml` is currently ignored. This leads to a situation where during `pipeline.sync` generic error configuration is returned by `parser` and overwrites jobs for the pipeline. 

## Objective

Errors when fetching `screwdriver.yaml` should not be ignored. This lets caller handle the error and eventually fail any  API call in progress, rather than causing pipeline config to be overwritten 

## References

https://github.com/screwdriver-cd/models/blob/master/lib/pipeline.js#L721-L722

## License

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -295,9 +295,10 @@ class PipelineModel extends BaseModel {
                     }
 
                     return this.scm.getFile(config).catch(err => {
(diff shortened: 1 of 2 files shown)


DekusDenial (approved):

tkyi (changes requested):
Need to fix tests

tkyi (approved):

PR #450, opened 2020-05-12
feat(1785): Add query to get latest builds for groupEventId [3.5]

tkyi:
## Context

We will want to get the latest build for each job in an event that has a matching `groupEventId`. It would be nice to have a sql query to get the data directly instead of making multiple calls to get the latestBuilds for each event, then processing each event's builds to get the desired data.

## Objective

This PR adds raw SQL queries to get the desired data.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/1785

## License

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,8 +347,8 @@ factory.get({ jobId, number }).then(model => {
 });
 ```
 
-| Parameter        | Type  |  Description |
-| :-------------   | :---- | :-------------|
+| Parameter      | Type  |  Description   |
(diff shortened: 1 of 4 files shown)


pritamstyz4ever (approved):

parthasl (approved):
👍 It would be great if comments can be added - which query is intended for which feature or some reference in rawQueries. forgot to mention this in @InderH PR as well.

PR #451, opened 2020-05-14
fix: Logging template name used for build

wahapo:
## Context
In order to measure the usage statistics of the template, I want to log the template name used at the timing when the build is executed. It may be possible to extract it from the DB by daily batch processing, but I don't think it's bad if drop it in the log. I would like to hear your opinion.

## Objective
If users use the template at build, it will be logged as follows.
```
{"message":"pipelineId:1: wahapo/test uses template wahapo/test@0.0.0","level":"info","timestamp":"2020-05-14T07:41:14.122Z"}
```

## License

### DIFF
--- a/lib/buildFactory.js
+++ b/lib/buildFactory.js
@@ -191,9 +191,13 @@ class BuildFactory extends BaseFactory {
         /* eslint-disable global-require */
         const JobFactory = require('./jobFactory');
         const StepFactory = require('./stepFactory');
+        const TemplateFactory = require('./templateFactory');
+        const PipelineFactory = require('./pipelineFactory');
         /* eslint-enable global-require */
         const factory = JobFactory.getInstance();
         const stepFactory = StepFactory.getInstance();
+        const templateFactory = TemplateFactory.getInstance();
+        const pipelineFactory = PipelineFactory.getInstance();
 
         return factory.get(jobId).then(job => {
             if (!job) {
@@ -265,6 +269,16 @@ class BuildFactory extends BaseFactory {
                     ];
 
                     modelConfig.templateId = job.templateId;
+                    if (job.templateId !== undefined && job.templateId !== null) {
+                        Promise.all([
+                            templateFactory.get({ id: job.templateId }),
+                            pipelineFactory.get({ id: job.pipelineId })
+                        ]).then(([t, p]) =>
+                            logger.info(
+                                `pipelineId:${p.id}: ${p.name} uses template ${t.namespace}/${t.name}@${t.version}`
+                            )
+                        );
+                    }
 
                     modelConfig.stats = {};
 
--- a/test/lib/buildFactory.test.js
+++ b/test/lib/buildFactory.test.js
@@ -34,11 +34,15 @@ describe('Build Factory', () => {
     let jobFactoryMock;
     let userFactoryMock;
     let stepFactoryMock;
+    let templateFactoryMock;
+    let pipelineFactoryMock;
     let buildClusterFactoryMock;
     let scmMock;
     let factory;
     let jobFactory;
     let stepFactory;
+    let templateFactory;
+    let pipelineFactory;
     let buildClusterFactory;
     const apiUri = 'https://notify.com/some/endpoint';
     const tokenGen = sinon.stub();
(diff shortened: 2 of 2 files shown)


lib/buildFactory.js
```diff
@@ -265,6 +269,16 @@ class BuildFactory extends BaseFactory {
                     ];
 
                     modelConfig.templateId = job.templateId;
+                    if (job.templateId !== undefined && job.templateId !== null) {
+                        Promise.all([
+                            templateFactory.get({ id: job.templateId }),
+                            pipelineFactory.get({ id: job.pipelineId })
+                        ]).then(([t, p]) =>
+                            logger.info(
+                                `pipelineId:${p.id}: ${p.name} uses template ${t.namespace}/${t.name}@${t.version}`
```
jithine, line 278: can we store this in build model ? We have lot's of users asking to display template information in build detail page.

ibu1224 (approved):
I am in favor of outputting to the log.

lib/buildFactory.js
wahapo: The build model already has a [`templateId`](https://github.com/screwdriver-cd/data-schema/blob/1cc4a369f3b77c6ea2090ae8fbc1a0fed428ac51/models/build.js#L131-L134). Therefore, it is possible to display template informations without saving it in the DB, and the flexibility is increased.

kumada626 (approved):

lib/buildFactory.js
jithine: can this is done based on log level ? https://github.com/screwdriver-cd/logger/blob/master/index.js#L6

This is `logger.info`, and logger will drop if log level > `info` and making 2 DB calls (already excessive) will be a waste in this situation.

jithine, line 278: We are already storing build count in DB and is available for each template. For eg: https://cd.screwdriver.cd/templates/python/validate_type

You don't even need to do batch processing in DB, this can be done at API level. Why add extra DB calls and logs for something which is already built in ?

jithine (changes requested):
This log as it is doesn't make sense to me without adding a buildId as well. We are already tracking template user on a per job and build level. And is already displayed in UI against each template.

lib/buildFactory.js
wahapo: Indeed. However, I think here that I should avoid providing if branching logic depending on the log level. I think there is no other way than to output only `templateId` (with `pipelineId` and `jobId`). Even then, I can achieve about half my goal.

wahapo: There are three reasons.

1. Our platform has a general-purpose mechanism that measures the number of occurrences from logs and graphs them. Using this is the most efficient way to get the aggregate results I want.
2. There are many templates prepared in YJ. I would like to investigate which template is used frequently and whether the number of users is increasing. To get usage statistics for multiple templates, we need to create a slightly more sophisticated system. If OSS concludes that the output to the log is not valid, then we(YJ) intends to do this.
3. I can also think that logging the template uses used is reasonable

jithine: Number of jobs and builds which uses a template is already stored with template, so we can already query this in an offline flow.  

We can keep the log, but please hide it behind a feature flag. Logging is not the problem, but the 2 extra DB calls done for each build is. So I don't want this to be default but controlled by a feature flag. Let's say `SD_DEBUG_LOGS` or something. 

This is not the same as [LOG_LEVEL](https://github.com/screwdriver-cd/logger/blob/master/index.js#L6), but we are using this logging to provide extra features in a build cluster. I would prefer to have SD cluster admins to have control over whether this extra DB calls is needed or not.

wahapo: I've been convinced right now that getting the data from the DB or API is the most reasonable way. So I want to close this PR without merging.

I understood the idea that `SD_DEBUG_LOGS` controls whether or not to log when more information is required even if the performance deteriorates. Certainly it looks good. However, when more logs are controlled by this flag in the future, I will have to turn this flag off for better performance. It will be a temporary solution to its original purpose, but it will not be permanent for me. When I need to turn the flag off, I'll change the aggregation mechanism to retrieve the data from the DB or API. It makes the most sense to get the data from the DB or API from the beginning.

Thanks, your opinion.

PR #453, opened 2020-05-28
feat(1044): Add exclude expression in SourcePath

kumada626:
## Context

If sourcePath parameter starts with `!`, that path treated as exclude path.

## Objective

Even if source path is matched, a build does not start if exclude path is matched.

## References

screwdriver-cd/screwdriver#1044

## License

### DIFF
--- a/test/lib/eventFactory.test.js
+++ b/test/lib/eventFactory.test.js
@@ -1463,6 +1463,36 @@ describe('Event Factory', () => {
             });
         });
 
+        it('should not start build if changed file is in exclude soucePath', () => {
+            jobsMock = [
+                {
+                    id: 1,
+                    pipelineId: 8765,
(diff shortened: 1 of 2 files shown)


ibu1224 (approved):

test/lib/eventFactory.test.js
```diff
@@ -1463,6 +1463,36 @@ describe('Event Factory', () => {
             });
         });
 
+        it('should not start build if changed file is in exclude soucePath', () => {
```
tkyi:
```suggestion
        it('should not start build if changed file is in exclude sourcePath', () => {
```
small comment, otherwise LGTM 👍

kumada626: Fixed.

tkyi (approved):

PR #454, opened 2020-05-29
fix: [1] Tag/Release trigger does not depend on changedFile

wahapo:
## Context
Currently, even if we create a tag, it does not start to build unless it matches changedFiles. Tag/Release triggers need to start built without depending on the changedFiles.

## Objective
Fixed to start build regardless of changedFile for tag/release triggers.

## References
https://github.com/screwdriver-cd/screwdriver/pull/2099

## License

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -120,6 +120,7 @@ function getJobsFromPR(config) {
  * @method startBuild
  * @param  {Object}   config                    configuration object
  * @param  {Object}   config.buildConfig        Build Config to create the build with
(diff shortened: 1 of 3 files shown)


tkyi:
I'm okay with this change, but not sure if there will be issues with our customers
@jithine is this expected behavior for our users?

jithine:
I am good with this change. Current behavior should be called out as a bug. Tag/Release represents the entire code base.

tkyi (approved):

yoshwata (approved):

PR #455, opened 2020-06-03
feat(2087): Pass token to executor for qs api call

pritamstyz4ever:
## Context

All calls to queue service must be authenticated and have scope sdapi.

## Objective

This PR passes token for all executor functions with scope sdapi.

## References

https://github.com/screwdriver-cd/screwdriver/issues/2087

## License

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -32,6 +32,24 @@ async function findMetrics(build, stepName, aggregateInterval) {
     return metrics;
 }
 
+/**
+ * @description Gets the token from token gen fn
+ * @param {Object} self -> this object
+ * @param {String} jobId
+ * @param {Object} pipeline
+ * @return {String} A jwt token
+ */
+function getToken(self, pipeline, jobId) {
+    const tokenGenConfig = {
+        pipelineId: pipeline.id,
+        jobId
+    };
+
+    const token = self[tokenGen](pipeline.username, tokenGenConfig, pipeline.scmContext, ['sdapi']);
+
+    return token;
+}
+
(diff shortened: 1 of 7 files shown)


lib/jobFactory.js
```diff
@@ -5,6 +5,24 @@ const { getAnnotations } = require('./helper');
 const Job = require('./job');
 let instance;
 
+/**
+ * @description Gets the token from token gen fn
+ * @param {Object} self -> this object
+ * @param {String} jobId
+ * @param {Object} pipeline
+ * @return {String} A jwt token
```
tkyi:
```suggestion
 * Gets the token from token gen fn
 * @method getToken
 * @param  {Object} self      This object
 * @param  {String} jobId     Job ID
 * @param  {Object} pipeline  Pipeline
 * @return {String}           A JWT token
```

lib/job.js
```diff
@@ -32,6 +32,24 @@ async function findMetrics(build, stepName, aggregateInterval) {
...
 }
 
+/**
+ * @description Gets the token from token gen fn
+ * @param {Object} self -> this object
+ * @param {String} jobId
+ * @param {Object} pipeline
+ * @return {String} A jwt token
+ */
+function getToken(self, pipeline, jobId) {
```
jithine: move this to https://github.com/screwdriver-cd/models/blob/master/lib/helper.js ?
this seems to be duplicated in jobFactory as well

jithine (approved):

parthasl (approved):
👍

PR #457, opened 2020-06-09
fix: support table prefix for raw queries

jithine:
## Context

Currently aggregate view does not work  in beta environment because table queries are hard coded.

## Objective

Support table prefix for raw queries

## License


vnugopal (approved):

parthasl (approved):
👍

PR #462, opened 2020-06-23
fix: fix error handling when steps do not exist.

sakka2:
## Context
`TypeError: Promise resolver undefined is not a function` happens if buildId does not have steps data.
```
data: TypeError: Promise resolver undefined is not a function
at new Promise (<anonymous>)
at /usr/src/app/node_modules/screwdriver-models/lib/build.js:490:28
```

## Objective
Fix error handling properly. (return promise.resolve())

## License

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -505,7 +505,7 @@ class BuildModel extends BaseModel {
 
             return this.getSteps().then(steps => {
                 if (steps.length === 0) {
-                    throw new Error('Steps do not exist');
+                    return Promise.resolve();
                 }
 
                 return Promise.all(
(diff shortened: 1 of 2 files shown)


lib/build.js
```diff
@@ -505,7 +505,7 @@ class BuildModel extends BaseModel {
 
             return this.getSteps().then(steps => {
                 if (steps.length === 0) {
-                    return new Promise();
+                    throw new Error('Steps do not exist');
```
tk3fftk: This error seems not to be caught in at least L526 and [here](https://github.com/screwdriver-cd/screwdriver/blob/f805359255d903c153d75de39fad0413e253dbe4/plugins/builds/update.js#L119), is it fine?
And I think we should return a correct Promise object, not an error.

sakka2: Thanks, fixed to return Promise.resolve()

MysticDoll (approved):
looks good to me

tk3fftk (approved):

tkyi (approved):

PR #464, opened 2020-07-06
fix: return 404 when steps does not exists

MysticDoll:
## Context
API Server returns 500 response when we send request to get build information that does not have `steps` in `steps` table of DB.
In this case, it should be 404 and this PR fixes it.

## Objective
Fix to return 404 response If `steps` does not exist in `models.toJsonWithSteps`.

## License

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const boom = require('boom');
 const dayjs = require('dayjs');
 const hoek = require('hoek');
 const deepcopy = require('deepcopy');
@@ -656,7 +657,7 @@ class BuildModel extends BaseModel {
             })
             .then(steps => {
                 if (!steps.length) {
-                    throw new Error('Steps do not exist');
+                    throw boom.notFound('Steps do not exist');
                 }
 
                 // This if statement should be removed after enough time has passed since build.steps removed.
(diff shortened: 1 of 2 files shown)


lib/build.js
```diff
@@ -656,7 +657,7 @@ class BuildModel extends BaseModel {
             })
             .then(steps => {
                 if (!steps.length) {
-                    throw new Error('Steps do not exist');
```
MysticDoll, line 659: This error will be handled [this line](https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/builds/get.js#L49) and a raw Error object becomes 500 response by `boom.boomify` method.
Therefore the object made by `boom.notFound` should be thrown. (and `boom.notFound()` will still be 404 response.)

tk3fftk (approved):
It looks working as expected, lgtm:+1:
```
> function hoge() {throw boom.notFound(123)}
undefined
> boom.boomify(hoge())
Thrown:
{ Error: 123
    at hoge (repl:1:29)
  data: null,
  isBoom: true,
  isServer: false,
  output:
   { statusCode: 404,
     payload: { statusCode: 404, error: 'Not Found', message: '123' },
     headers: {} },
  reformat: [Function],
  typeof: [Function: notFound] }
```

tkyi (approved):

PR #465, opened 2020-07-09
fix(2138): Remove scheduler when deleted from buildPeriodically

wahapo:
## Context
Fix issue https://github.com/screwdriver-cd/screwdriver/issues/2138.

## Objective
If the setting of buildPeriodically removed, the schedule should also be deleted.

`screwdriver.yaml`:
```diff
 jobs:
   main:
-    annotations:
-      screwdriver.cd/buildPeriodically: H * * * *
   image: node:12
   steps:
    - ls
```

It is not necessary to call the API to check if buildPeriodically existed before update. Even if it call stopPeriodic when there is no scheduled task, sdqueue-service(`DELETE /v1/queue/message?type=periodic`) responds with 200, so I think this modification is sufficient.

```
IncomingMessage {
  statusCode: 200,
  statusMessage: 'OK',
  request: Request {
    path: '/v1/queue/message?type=periodic',
    method: 'DELETE',
    body: '{"jobId":1,"pipelineId":1,"token":"***"}',
(snip)
```

## References
https://github.com/screwdriver-cd/screwdriver/issues/2138

## License

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -212,15 +212,22 @@ class Job extends BaseModel {
         // https://nodejs.org/api/modules.html#modules_cycles
         /* eslint-disable global-require */
         const PipelineFactory = require('./pipelineFactory');
+        const JobFactory = require('./jobFactory');
         /* eslint-enable global-require */
         const pipelineFactory = PipelineFactory.getInstance();
+        const jobFactory = JobFactory.getInstance();
+
+        const oldPeriodic = await jobFactory
+            .get(this.id)
+            .then(job => getAnnotations(job.permutations[0], 'screwdriver.cd/buildPeriodically'));
 
         const newJob = await super.update();
         const pipeline = await pipelineFactory.get(newJob.pipelineId);
 
         try {
-            // FIXME:: Make this call only if buildCron config is updated
-            if (getAnnotations(this.permutations[0], 'screwdriver.cd/buildPeriodically')) {
+            const newPeriodic = getAnnotations(this.permutations[0], 'screwdriver.cd/buildPeriodically');
+
+            if (newPeriodic && newPeriodic !== oldPeriodic) {
                 await this[executor].startPeriodic({
                     pipeline,
                     job: newJob,
@@ -229,7 +236,7 @@ class Job extends BaseModel {
                     apiUri: this[apiUri],
                     isUpdate: true
                 });
-            } else {
+            } else if (!newPeriodic && oldPeriodic) {
                 await this[executor].stopPeriodic({
                     jobId: this.id,
                     pipelineId: pipeline.id,
--- a/lib/job.js
+++ b/lib/job.js
@@ -229,6 +229,12 @@ class Job extends BaseModel {
                     apiUri: this[apiUri],
                     isUpdate: true
                 });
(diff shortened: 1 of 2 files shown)


test/lib/job.test.js
```diff
@@ -442,6 +442,17 @@ describe('Job Model', () => {
                 assert.calledOnce(datastore.update);
             });
         });
+
+        it('remove periodic job', () => {
+            job.state = 'DISABLED';
```
yoshwata: Is this needed? Job's `ENABLED/DISABLED` status seems to be unrelated.

wahapo: fixed

yoshwata (approved):

lib/job.js
```diff
@@ -229,6 +229,12 @@ class Job extends BaseModel {
                     apiUri: this[apiUri],
                     isUpdate: true
                 });
+            } else {
+                await this[executor].stopPeriodic({
```
jithine, line 240: we will be making this Queue service call for each `job.update()` and this happens all the time as part of `pipeline.sync()`, 
 1. when webhooks are fired.
 2. when a build is started. 

While this is the right thing to do when `buildPeriodically` config is removed, I'm afraid current implementation will lead to an increase in load of queue service by many fold, especially for pipelines with large number of jobs. 

So we should be doing this only if the config was there before. (like the comment I've left [here](https://github.com/screwdriver-cd/models/blob/81dc78171961324b456c56fdedc1bd5d4daa1d48/lib/job.js#L222))

wahapo: I fixed it.

ibu1224 (approved):

wahapo:
Added a test that does not call `startPeriodically()` when the value of `screwdriver.cd/buildPeriodically` has not changed.

yoshwata (approved):

jithine (approved):

tkyi (approved):

pritamstyz4ever:
There is an issue with the deletion logic of periodic builds. Periodic builds before the change for which sd.yaml was updated, will remain in the queue and run periodically with an UNKNOWN status.
Like here: https://cd.screwdriver.cd/pipelines/4629/events
In this case the jobs with UNKNOWN status is publish3 which has been renamed to published5
https://github.com/pritamstyz4ever/sd-sample-build/commit/ace414f10f20ee6a149bae6382c3507957cd6160
still remains in the queue.
The logic here to get the old settings returns undefined as the model was updated before change but redis was not cleared:
https://github.com/screwdriver-cd/models/blob/master/lib/job.js#L220-L222
so both newperiodic and old periodic values are undefined and the stopPeriodic never gets called.
@kumada626  @wahapo @jithine

pritamstyz4ever:
similarly, in many jobs where there was a pipeline update and had an existing periodic, they got removed from redis. 
as per this logic 
https://github.com/screwdriver-cd/models/blob/master/lib/job.js#L239.
Now as model is updated, for both newPeriodic and oldPeriodic would return the same value and hence the periodic jobs never get added again.

So there are two issues:
1. old periodic job references not cleared from queue
2. actual jobs cleared as model was updated

PR #467, opened 2020-07-23
fix: Add check for disabled and archived job

pritamstyz4ever:
## Context

Change for removing settings of periodic builds misses certain cases where settings was removed but not cleared from queue before changes were pushed which leads to some jobs with empty events and some jobs not being added to queue unless settings are removed and re-added.

## Objective
This PR will fix the issue by adding additional checks for disabled or archived jobs.

## References

https://github.com/screwdriver-cd/models/pull/465#issuecomment-662623305
https://github.com/screwdriver-cd/models/pull/465#issuecomment-662647213
https://github.com/screwdriver-cd/screwdriver/issues/2138

## License

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -226,10 +226,10 @@ class Job extends BaseModel {
         try {
             const newPeriodic = getAnnotations(newJob.permutations[0], 'screwdriver.cd/buildPeriodically');
             const isJobEnabled = newJob.state === 'ENABLED';
-            const isSettingUpdated = newPeriodic && newPeriodic !== oldPeriodic;
-            const wasJobDisabled = newPeriodic === oldPeriodic && oldJob.state === 'DISABLED';
+            const isSettingUpdated = newPeriodic ? newPeriodic !== oldPeriodic : !!oldPeriodic;
+            const wasJobDisabled = oldJob.state === 'DISABLED';
 
-            if ((isSettingUpdated || wasJobDisabled) && isJobEnabled) {
+            if (newPeriodic && (isSettingUpdated || wasJobDisabled) && isJobEnabled) {
                 await this[executor].startPeriodic({
                     pipeline,
                     job: newJob,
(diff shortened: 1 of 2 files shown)


lib/job.js
```diff
@@ -224,10 +224,12 @@ class Job extends BaseModel {
         const newJob = await super.update();
         const pipeline = await pipelineFactory.get(newJob.pipelineId);
 
+        const isDisabled = this.state === 'DISABLED';
+
         try {
             const newPeriodic = getAnnotations(this.permutations[0], 'screwdriver.cd/buildPeriodically');
 
-            if (newPeriodic && newPeriodic !== oldPeriodic) {
+            if (newPeriodic && newPeriodic !== oldPeriodic && !isDisabled) {
```
jithine: need to add a check to detect  transition from old job state disabled -> new state enabled, else jobs with disabled -> enabled state transition won't have periodic build configuration in redis

lib/job.js
```diff
@@ -217,17 +217,19 @@ class Job extends BaseModel {
...
 
         try {
-            const newPeriodic = getAnnotations(this.permutations[0], 'screwdriver.cd/buildPeriodically');
+            const newPeriodic = getAnnotations(newJob.permutations[0], 'screwdriver.cd/buildPeriodically');
+            const isJobEnabled = newJob.state === 'ENABLED';
+            const isSettingUpdated = newPeriodic && newPeriodic !== oldPeriodic;
+            const wasJobDisabled = newPeriodic === oldPeriodic && oldJob.state === 'DISABLED';
 
-            if (newPeriodic && newPeriodic !== oldPeriodic) {
+            if ((isSettingUpdated || wasJobDisabled) && isJobEnabled) {
```
jithine: shouldn't this be
```
const wasJobDisabled = oldJob.state === 'DISABLED';

 if (isSettingUpdated || (wasJobDisabled && isJobEnabled)) { 


```

pritamstyz4ever: But if, job is not enabled, then shouldn't it remove the periodic settings?

jithine: won't then `isJobEnabled` will be false, and and then the following `if` block catches it ?

jithine (approved):

PR #468, opened 2020-07-25
Configurable config file path and extending api

ysaakpr:
## Context

There is no way to modify the screwdriver.yaml config file for a pipeline. And along with that there are no way to extend or dynamically create screwdriver.yaml file from external apis. Which is more generic way to pipeline templating.

## Objective
1. Give an option to Set global screwdriver.yaml path, the name can be changed to some other name using this feature.
2. No pipeline level config file override is possible in this PR, after the review of this we can provide per pipeline override for the config file.
3. Ability to extend the scredriver.yaml from an external API as configured in the screwdriver installation.
4. No pipeline level extension apis supported in this PR, can be provided in the subsequent PR based on the feedback.

## References

https://github.com/screwdriver-cd/screwdriver/issues/2145

## License

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -312,10 +312,7 @@ class PipelineModel extends BaseModel {
                             path,
                             scmRepo: this.scmRepo
                         };
-                        console.log({
-                            config,
-                            screwdriverConf
-                        })
+
                         return request({
                             followAllRedirects: true,
                             json: true,
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -7,6 +7,7 @@ const parser = require('screwdriver-config-parser');
 const workflowParser = require('screwdriver-workflow-parser');
 const hoek = require('hoek');
 const dayjs = require('dayjs');
+const request = require('requestretry');
 const _ = require('lodash');
 const { PR_JOB_NAME, EXTERNAL_TRIGGER_ALL } = require('screwdriver-data-schema').config.regex;
 const logger = require('screwdriver-logger');
@@ -282,10 +283,11 @@ class PipelineModel extends BaseModel {
                 .then(user => user.unsealToken())
                 // fetch the screwdriver.yaml file
                 .then(token => {
+                    const path = pipelineFactory.getScrewdriverPath();
                     const config = {
                         scmUri: this.scmUri,
                         scmContext: this.scmContext,
-                        path: 'screwdriver.yaml',
+                        path,
                         token,
                         scmRepo: this.scmRepo
                     };
@@ -299,6 +301,50 @@ class PipelineModel extends BaseModel {
 
                         logger.error(message, err);
                         throw new Error(message);
+                    });
+                })
+                .then(screwdriverConf => {
+                    if (pipelineFactory.isExtendingEnabled()) {
+                        const path = pipelineFactory.getScrewdriverPath();
+                        const config = {
+                            scmUri: this.scmUri,
+                            scmContext: this.scmContext,
+                            path,
+                            scmRepo: this.scmRepo
+                        };
+                        console.log({
+                            config,
+                            screwdriverConf
+                        })
+                        return request({
+                            followAllRedirects: true,
+                            json: true,
+                            method: 'POST',
+                            body: {
+                                config,
+                                screwdriverConf
+                            },
+                            url: pipelineFactory.getScrewdriverExtendApi()
+                        })
+                            .then(response => {
+                                if (response.statusCode >= 200 || response.statusCode < 300) {
+                                    return response.body;
+                                }
+                                const message = `pipelineId:${this.id}: Failed to generate screwdriver.yaml. Error Code  ${response.statusCode}`;
+
+                                logger.error(message, response.body);
+                                throw new Error(message);
+                            })
+                            .catch(error => {
+                                const message = `pipelineId:${this.id}: Failed to generate screwdriver.yaml.`;
+
+                                logger.error(message, error);
+                                throw new Error(message);
+                            });
+                    }
+
+                    return new Promise(resolve => {
+                        resolve(screwdriverConf);
                     });
                 })
                 // parse the content of the yaml file
--- a/lib/pipelineFactory.js
+++ b/lib/pipelineFactory.js
@@ -141,7 +141,7 @@ class PipelineFactory extends BaseFactory {
      * @return {Boolean}
      */
     getScrewdriverExtendApi() {
-        this.screwdriverConfig.defaultExtendApi.trim();
+        return this.screwdriverConfig.defaultExtendApi.trim();
     }
 
     /**
--- a/lib/pipelineFactory.js
+++ b/lib/pipelineFactory.js
@@ -17,6 +17,9 @@ class PipelineFactory extends BaseFactory {
         super('pipeline', config);
         this.multiBuildClusterEnabled = config.multiBuildClusterEnabled;
         this.externalJoin = config.externalJoin || false;
+        this.screwdriverConfig = config.screwdriverConfig || {
+            defaultPath: 'screwdriver.yaml'
+        };
     }
 
     /**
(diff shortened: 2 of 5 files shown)


ysaakpr:
Semantic version update may be needed, either the maintainers can update it, or let me know what is the next version, I can update the source.

jithine (changes requested):
We need to define the use cases first, before making this into code. I think building on top of existing Template feature will be a better fit than creating a separate micro service to handle the same use case, except that we are pushing out the job template stitching to this service which SD nor configured SCM  has visibility into. 

I think once we define the use case and product design we can start looking into implementation. Right now it's too early to get this code in.

ysaakpr:
Thanks @jithine for taking some time to look into this during your busy schedules.  I have used pipeline level templates pretty much in many cd systems. That gives us to define a standard pipeline process. But cant validates a project is following certain pipeline and configured certain mandatory steps in the workflow. A remote API to fetch screwdriver.yaml gives lots of possibilities for the end user. We can still use the templating. But think one use case that the screwdriver.yaml is maintained outside of the src repo. and managed by a devops team and they want certain validation, step injection, which will be beyond the scope of templating, 

And a feature like this is not a replacement for template feature any day. For reference you can look at Gitlab ci, you can configure the gitlab-ci.yaml file name and load the ci pipeline from a remote api.

ysaakpr:
And since this step is just before parsing the loaded screwdriver.yaml, there is no feature conflict by using a remote api to fetch/validate the loaded config. Think over it, this can increase the screwdriver adoption, because we already have certain yaml files created in the repo, and if the api can generate the scredriver.yaml from it, we have zero dependency from developers to migrate to new cd tool. Which is very important aspect for a migration to orgs where we already have running cd system..

ysaakpr:
I can give you 100s of possibilities by enabling different naming for the cd config file, and ability to use a remote api to derive the final config. Its much beyond the templating feature that you have mentioned.

ysaakpr:
And frankly, without this feature, even my org we not going to adopt it. because it's difficult for moving and setting up 100s of services to follow a new YAML structure for their pipelines. If you really understand the power an external api brings to screwdriver, we can discuss further and make these configurations much more cleaner, this is a demo code that i have written to test the behaviour.

lib/pipeline.js
```diff
@@ -301,6 +303,47 @@ class PipelineModel extends BaseModel {
                         throw new Error(message);
                     });
                 })
+                .then(screwdriverConf => {
+                    if (pipelineFactory.isExtendingEnabled()) {
```
jithine, line 307: this should based on a specific individual pipeline, if an organization wants to enforce this for some but not all.

jithine, line 307: this is part of `pipeline.sync` which happens for all event creation or webhook delivery. and will be a waste of processing time. Webhooks already have strict sla of < 10s. So I would prefer to have this generation an  offline process, which is not part of run time.

jithine (changes requested):
@ysaakpr I am not opposed to having remote fetching on sd.yaml config. But some specifics on this approach. 

1. This should be an offline flow, current implementation is having this called as part of `sync` flow, which happens during every build creation & webhook delivery. Since `screwdriver.yaml` doesn't change that often, I believe a better approach would be to push the changes to config, rather than pull it during each build.
2. Before we introduce big architectural changes, we create a design doc at https://github.com/screwdriver-cd/screwdriver/tree/master/design and get it reviewed by team. For example, I am not clear on what is the auth model for this service is, and are there specifics on input ? (output we know it's a generated screwdriver.yaml`)
3. Have this feature available at individual pipeline level instead of a single config for the SD cluster. We can do this by taking in a config during pipeline creation in UI, to specify that fetch this pipeline's `yaml` config from a remote location. I agree that top-down approach of External config is not be a good fit here. 

I believe this is a very good feature to have. We just need to work on design, before we start implementation. It helps settle expectations on PRs and from a feature perspective makes sure everyone is on same page.

ysaakpr:
> This should be an offline flow, the current implementation is having this called as part of sync flow, which happens during every build creation & webhook delivery. Since screwdriver.yaml doesn't change that often, I believe a better approach would be to push the changes to config, rather than pull it during each build.

I don't think that is correct, you are fetching the screwdriver.yaml at every build using git API from the source repo. Why do you believe that the pipeline config won't change in every commit? Let's not assume, whatever we assume restrict the user. According to me this is not just a loading of screwdriver.yaml from a remote location, its more of validating/extending/enriching of the initial configs by the help of a third party API.

> Before we introduce big architectural changes, we create a design doc at https://github.com/screwdriver-cd/screwdriver/tree/master/design and get it reviewed by team. For example, I am not clear on what is the auth model for this service is, and are there specifics on input ? (output we know it's a generated screwdriver.yaml`)

I really don't think its an architectural change, there is no change in the architecture. The screwdriver system always gonna work with the screwdriver.yaml. And its getting loaded from the repo as usual. The only thing these changes are bringing is to specify another path for hardcoded `screwdriver.yaml` + before giving the loaded `screwdriver.yaml` to the parser, we pass it through a function, and functions output will be considered as the final content of screwdriver.yaml
Since these changes are currently configured buy HTTP apis, in a first go we can do basic authentication support, and server-level option to force https.

>Have this feature available at individual pipeline level instead of a single config for the SD cluster. We can do this by taking in a config during pipeline creation in UI, to specify that fetch this pipeline's yaml config from a remote location. I agree that top-down approach of External config is not be a good fit here.

In my PR I have mentioned, this is not a foolproof PR, This is PR i have made to tell you the possibilities and the how simple it is to add on the existing code. As I mentioned, we can give a global server level option, and repo level further overrides. But i believe just giving it only at the repo level will be again an overhead for teams where central team managing/enforcing certain process to bring certain standard. Global configs are always helping in that case, and we definitely can provide further overrides at individual repo level.

> I believe this is a very good feature to have. We just need to work on design, before we start implementation. It helps settle expectations on PRs and from a feature perspective makes sure everyone is on same page.

I really thank you for spending time on this ticket. Appreciate it. If you can propose this with the team, and if the team see value in this feature, definitely we can go through the PR and make the best effort to improve this solution further. As you mentioned, we can start with a PRD and feature expectation documentation before going ahead with implementation.

lib/pipeline.js
ysaakpr: How much time/ Error handling cases can be tuned further using HTTP API timeouts, and defaults for the API configs. But again limiting everywhere on the user is what the first reason for other choices. Think from the user perspective. If there is a strong reason that we have to finish the total process in a certain time we can cap the max timeout for this API in the code. But there should be a reason why we caping it. And making the user aware about the impact of long HTTP timeout is a better way to go ahead is what my kind opinion.

ysaakpr: that can be again as per the org decision. In my case, if we are adopting, we want all the repo to follow that, and in case any repo wants to try something else they could choose it by overriding it at repo level. We should provide both. If the user don't want global override, ignore the config, else add a config. Again, give freedom to the user's needs. a good product always needs flexibility to adoption. Take an example of Jira. along with defining an std process to follow, they have given away to override each and everything, and that's the single reason for the enterprise adoption of the solution.

jithine: we don't have control of this, this is GHE webhook time out. So we have to be mindful of not breaking webhook processing. We can't penalize users for this, so our design has to take into account on this limit. 
We do have plans to make entire webhook processing an offline process. https://github.com/screwdriver-cd/screwdriver/issues/1769

jithine: My point was more on the lines of, instead of generating this dynamically during each build run, this screwdrive.yaml config can be generated when ever there is a change in per repo config  or there is a change in upstream service which defines overall pipeline config.

jithine: my feedback is around if user provides a complete `screwdriver.yaml` use it instead of always calling external API even if it's enabled.

Are you looking at more of a lock-in per SD instance, where all users must use pipeline config from this external API ?

jithine:
Closing this, since this needs more discussions to detail out data flows and overall design.

PR #474, opened 2020-09-10
fix(2194): trigger from external trigger in child pipeline

wahapo:
## Context
Fix issue https://github.com/screwdriver-cd/screwdriver/issues/2194.

## Objective
As shown in the screenshot below, the child pipeline can trigger external triggers on other pipelines.

## License

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -275,7 +275,7 @@ class PipelineModel extends BaseModel {
                     return configPipeline.getConfiguration({});
                 }
 
-                return configPipeline.getConfiguration({ ref, id: this.id });
+                return configPipeline.getConfiguration({ ref, id });
             });
         }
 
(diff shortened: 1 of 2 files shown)


ibu1224 (approved):

yoshwata (approved):

lib/pipeline.js
```diff
@@ -273,7 +275,7 @@ class PipelineModel extends BaseModel {
                     return configPipeline.getConfiguration({});
                 }
 
-                return configPipeline.getConfiguration({ ref });
+                return configPipeline.getConfiguration({ ref, id: this.id });
```
tkyi:
```suggestion
                return configPipeline.getConfiguration({ ref, id });
```
should use the new variable too?

wahapo: Currently that pattern probably doesn't appear, but logically it's certainly better:+1:

tkyi (approved):

PR #475, opened 2020-09-17
feat(2199): Add method to get user settings [2]

tkyi:
## Context

It would be nice to be able to get user settings.

## Objective

This PR adds a method to get user settings.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2199

## License

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -434,7 +434,7 @@ function updateWorkflowGraph(config) {
 
 /**
  * Standardize event parameters
- * @method updateWorkflowGraph
+ * @method updateEventParameters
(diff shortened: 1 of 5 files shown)


tkyi:
Not sure if it would be better to just write a generic `get()` for users

adong:
> Not sure if it would be better to just write a generic `get()` for users

I think either way is okay 🤷‍♂️ . This is for internal use. We can later expose API route `GET /user` that returns 

```
{ 
   user: { 
       settings: { /*  ... */ }
   } 
}
```

adong (approved):

PR #478, opened 2020-10-20
feat(2199): Enhance metrics with event downtime information [2]

tkyi:
## Context

Need to add event downtime data to metrics function.

## Objective

This PR enhances the metrics method with event downtime information.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2199

## License

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -26,12 +26,15 @@ const { getAllRecords, getBuildClusterName } = require('./helper.js');
 const JOB_CHUNK_SIZE = process.env.JOBS_PARALLEL_COUNT || 5;
 const SD_API_URI = process.env.URI;
 
+const DEFAULT_DOWNTIME_JOBS = [];
+const DEFAULT_DOWNTIME_STATUSES = ['FAILURE'];
+
 // Process jobs JOB_CHUNK_SIZE at at time
 const getJobChunks = jobs => _.chunk(jobs || [], JOB_CHUNK_SIZE);
 
 // Calculate event duration by using max endTime - min startTime of builds
 // that belong this event
-const eventMetrics = async event => {
+const eventMetrics = async ({ event, downtimeJobs, downtimeStatuses }) => {
     const buildMetrics = await event.getMetrics();
 
     // Somehow this event doesn't have any builds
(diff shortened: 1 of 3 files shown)


lib/pipeline.js
```diff
@@ -46,6 +49,8 @@ const eventMetrics = async event => {
     let totalQueuedTime = queuedTime || 0;
     let totalImagePullTime = imagePullTime || 0;
     const builds = [{ id, jobId, duration, status }];
+    let isDowntimeEvent =
+        ((downtimeJobs.length === 0 || downtimeJobs.includes(jobId)) && downtimeStatuses.includes(status)) || false;
```
pritamstyz4ever: do we need the || false?? It should evaluate to boolean right?

lib/pipeline.js
```diff
@@ -74,14 +79,21 @@ const eventMetrics = async event => {
             duration,
             status
         });
+
+        // Set downtime event flag if job is in downtimeJobs and status is in downtimeStatuses
+        if ((downtimeJobs.length === 0 || downtimeJobs.includes(jobId)) && downtimeStatuses.includes(status)) {
+            isDowntimeEvent = true;
+        }
     }
```
pritamstyz4ever, line 87: Looks like the condition can be reused ?

lib/pipeline.js
```diff
@@ -1544,6 +1573,26 @@ class PipelineModel extends BaseModel {
...
+                        dayjs(downtimeEvents[i].maxEndTime),
+                        'second'
+                    );
+                }
+
+                downtimeEvents[i].downtimeDuration = dayjs(new Date()).diff(
+                    dayjs(downtimeEvents[i].maxEndTime),
+                    'second'
+                );
+            }
```
pritamstyz4ever, line 1594: Probably use map fn?

test/lib/pipeline.test.js
```diff
@@ -2791,13 +2791,14 @@ describe('Pipeline Model', () => {
         });
     });
 
-    describe('get metrics', () => {
+    describe.only('get metrics', () => {
```
jithine: remove `only`

lib/pipeline.js
tkyi: the way it's written, the first element in buildMetrics is set manually, so I'm also setting `isDowntimeEvent` that way
do you have a better suggestion?

lib/pipeline.js
tkyi: need to do something different with last element in array
gets the duration based off current time vs next element's time

test/lib/pipeline.test.js
tkyi: updated

lib/pipeline.js
tkyi: removed

jithine (commented):
need to fix tests, otherwise LGTM

jithine (approved):

PR #483, opened 2020-12-14
fix(1044): support blanket exclusions

ibu1224:
## Context

Now sourcePaths will not start the build if exclude only sourcePaths is set.
For example, `screwdriver.yaml` like the following
```
jobs:
  build1:
    image: node12
    requires: [ ~commit  ]
    sourcePaths: ["!README.md"]
    steps:
      - echo "this is build1"
```

## Objective

Add a step so that the build will start when sourcePaths is exclude only.

## References

issue: https://github.com/screwdriver-cd/screwdriver/issues/1044

## License

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -238,6 +238,13 @@ function startBuild(config) {
                 return isMatchExclude;
             });
 
+            // sourcePath is only exclude
+            const notOnlyExclude = paths.some(source => source.startsWith('!') === false);
+
+            if (!notOnlyExclude && isFileExclude === false) {
+                return true;
+            }
+
             return isFileMatch && !isFileExclude;
         });
     }
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -239,9 +239,9 @@ function startBuild(config) {
             });
 
             // sourcePath is only exclude
-            const notOnlyExclude = paths.some(source => source.startsWith('!') === false);
+            const onlyExclude = paths.every(source => source.startsWith('!') === true);
 
-            if (!notOnlyExclude && isFileExclude === false) {
+            if (onlyExclude && !isFileExclude) {
                 return true;
             }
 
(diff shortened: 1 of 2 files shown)


wahapo (approved):

yoshwata (approved):

tkyi (approved):

jithine:
Does this implementation support mixed mode ? Will following logic work ?  `["!do-not-build-on-changes-here/", "src/"]

ibu1224:
@jithine 
I  tested it with the following case.

---
**Case1**
`sourcePaths`:
```
["!do-not-build-on-changes-here/", "src/"]
```
`changeFiles`:
```
["src/test"]
```
Result:
=> Build started
**Case2**
`sourcePaths`:
```
["!do-not-build-on-changes-here/", "src/"]
```
`changeFiles`:
```
["README.md"]
```
Result:
Build not started.

---

I added this time is when sourcePaths is exclude only.
For example, in the following cases.
`sourcePaths`:
```
["!do-not-build-on-changes-here/", "!src/"]
```

In the case of mixed sourcePaths, it seems to be implemented in the following PR.
https://github.com/screwdriver-cd/models/pull/453/

jithine:
@ibu1224 for case 2 test did you test it with `["!README.md", "src/"]` other wise `["!do-not-build-on-changes-here/", "src/"]` should have started build 

ie `["!README.md", "src/"]` should not start build for `README.md` change
but `["!do-not-build-on-changes-here/", "src/"]` should start build for `README.md` change

ibu1224:
@jithine 

> @ibu1224 for case 2 test did you test it with ["!README.md", "src/"] other wise ["!do-not-build-on-changes-here/", "src/"] should have started build

I tested the sourcePaths in case2 as follows
```
["!do-not-build-on-changes-here/", "src/"]
```

> ie ["!README.md", "src/"] should not start build for README.md change
but ["!do-not-build-on-changes-here/", "src/"] should start build for README.md change

I think it is better not to start the build.
if you start the build by changing the `README.md`, I think the multiple sourcePaths setting will be meaningless.

If the user adds `src/` to the sourcePaths in this case, I think it means that the user wants to start building only when the files in `src/` are changed.

jithine:
@ibu1224 you are right. Now I get it.

PR #486, opened 2021-01-22
fix(2312): add latest to template and command

klu909:
## Context

adding a latest field to template/command model
## Objective

update latest field when creating new template/command

## References

https://github.com/screwdriver-cd/screwdriver/issues/2312

## License

### DIFF
--- a/lib/commandFactory.js
+++ b/lib/commandFactory.js
@@ -65,21 +65,32 @@ class CommandFactory extends BaseFactory {
         const [, major, minor] = VERSION_REGEX.exec(config.version);
         const searchVersion = minor ? `${major}${minor}` : major;
 
-        return this.getCommand(`${config.namespace}/${config.name}@${searchVersion}`).then(latest => {
-            if (!latest) {
-                config.version = minor ? `${major}${minor}.0` : `${major}.0.0`;
-            } else {
-                // eslint-disable-next-line max-len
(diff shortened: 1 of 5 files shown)


lib/commandFactory.js
```diff
@@ -74,9 +74,13 @@ class CommandFactory extends BaseFactory {
                 const patch = parseInt(latestPatch.slice(1), 10) + 1;
 
                 config.version = `${latestMajor}${latestMinor}.${patch}`;
+                latest.latestVersion = false;
+
+                await latest.update();
```
tkyi: does the function on line 64 need `async` since you're using `await` here?

lib/templateFactory.js
```diff
@@ -156,9 +156,13 @@ class TemplateFactory extends BaseFactory {
                 const patch = parseInt(latestPatch.slice(1), 10) + 1;
 
                 result.version = `${latestMajor}${latestMinor}.${patch}`;
+                latest.latestVersion = false;
+
+                await latest.update();
```
tkyi: same comment as above

lib/commandFactory.js
jithine: yep, hence the build failure

tkyi (commented):
any way to test this change?

klu909:
> any way to test this change?

not really..

tkyi (approved):

PR #485, opened 2021-01-21
fix(2311): Make /v4/events performance better

yoshwata:
## Context

`v4/events` performance issue makes problem like https://github.com/screwdriver-cd/screwdriver/issues/2311
This PR make the performance better.

In my sandbox the performance becomes for 6500 archived jobs:
120028ms -> 82047ms
(still its too large)

## Objective

Make `v4/events` performance better by reducing enormous `DELETE /v1/queue/message`

## References

Please do **not** close these issues after this PR is merged. This is not complete solution.

https://github.com/screwdriver-cd/screwdriver/issues/2318
https://github.com/screwdriver-cd/screwdriver/issues/2311

## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #489, opened 2021-02-15
fix(2318): Unnecessary job update

ibu1224:
## Context

`v4/events` performance issue makes problem like https://github.com/screwdriver-cd/screwdriver/issues/2318
We tried to fix it before with the following PR, but it still seems a little slow.
https://github.com/screwdriver-cd/models/pull/485

When we looked into it, we found a process that was doing an unnecessary UPDATE.
https://github.com/screwdriver-cd/models/blob/3a2dbd03a42dda7b7b0bbaaf09d2d72c116fdb01/lib/pipeline.js#L867-L868

The following is an example of a change in `v4/events` velocity.
However, the speed will be affected by the network speed and other factors.
In my sandbox, it looks like this
```
Archived jobs : 1000
Before: 9.61s
After: 7.00s
```
```
Archived jobs : 4000
Before: 16.63s
After: 14.53s
```

It looks like it will perform a little better.
## Objective

If archived is already true, do not update it.

## References

- https://github.com/screwdriver-cd/models/pull/485
- https://github.com/screwdriver-cd/screwdriver/issues/2318
- https://github.com/screwdriver-cd/screwdriver/issues/2311
## License

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -864,8 +864,10 @@ class PipelineModel extends BaseModel {
                         job.archived = false;
                         updatedJobs.push(await job.update());
                     } else if (!job.isPR()) {
-                        job.archived = true;
-                        updatedJobs.push(await job.update());
+                        if (!job.archived) {
+                            job.archived = true;
+                            updatedJobs.push(await job.update());
+                        }
                     }
 
                     // sync external triggers for existing jobs


wahapo (approved):

yoshwata:
[When I checked before](https://github.com/screwdriver-cd/models/pull/485) it spends over 80000ms, so 1000ms or a few seconds seems to be very small. It might not be considered an effective measure.

yoshwata (approved):
Thank you for fixing the overview. It may be a little faster.

tkyi (approved):

PR #493, opened 2021-03-18
fix: add log on build stop

yoshwata:
## Context

I wanted to see the error when the build stop failed, but I couldn't.

## Objective

We can see the error log when a build stop failed.

## License

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -591,7 +591,7 @@ class BuildModel extends BaseModel {
             )
             .then(() => this)
             .catch(err => {
-                console.log(`Failed to stop a build: ${err}`);
+                logger.error(`Failed to stop a build: ${err}`);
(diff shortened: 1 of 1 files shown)


lib/build.js
```diff
@@ -589,7 +589,12 @@ class BuildModel extends BaseModel {
                         return this[executor].stop(config).then(() => this[executor].stopTimer(config));
                     })
             )
-            .then(() => this);
+            .then(() => this)
+            .catch(err => {
+                console.log(`Failed to stop a build: ${err}`);
```
tkyi: should this be `logger.error`?

yoshwata: You're right. Fixed.

tkyi (approved):

PR #494, opened 2021-03-19
fix(2071): add try catch to build create

klu909:
## Context

build create will throw unique constraint error after https://github.com/screwdriver-cd/data-schema/pull/443
## Objective

add try catch to guard build create
## References

https://github.com/screwdriver-cd/screwdriver/issues/2071
## License

### DIFF
--- a/lib/buildFactory.js
+++ b/lib/buildFactory.js
@@ -268,7 +268,14 @@ class BuildFactory extends BaseFactory {
 
                     modelConfig.stats = {};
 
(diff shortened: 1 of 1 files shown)


jithine (changes requested):
This is not the place to handle it, Here it should throw error back to API/models. And that is the where which should be handled. 
And all this should be done in the logic for `triggerNextJobs` in API. Because it need to catch the error and if there is uniqe constraint error, then it should fetch the created build instead and update it.

PR #503, opened 2021-05-18
Revert "fix: Revert to default meta status"

ombharatiya:
Reverts screwdriver-cd/models#499

Reverting this "Revert Default Status PR" to add "pending" status as default.


jithine:
Looks like the same eslint error as we saw for API

tkyi (approved):

PR #496, opened 2021-04-21
fix: Allow pending status for GitHub PR meta status check [1]

tkyi:
## Context

Users might want to report on a PENDING status for GitHub PR.

## Objective

This PR adds a default message for PENDING status, also changes default message to match default status (SUCCESS).

## References

N/A

## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #506, opened 2021-06-18
fix: add a status map to allow only valid SCM statuses

ombharatiya:
## Context

Add SCM Status Map in Build Models to allow only PENDING, SUCCESS & FAILED status

## Objective

Builds status and SCM status should not be mixed. In PR build status checks, it only need to validate status supported by SCM (GitHub, GitLab)

## References

Jira:

screwdriver-cd/screwdriver#2449

More related PRs:

screwdriver-cd/scm-github#189

screwdriver-cd/models#496

https://github.com/screwdriver-cd/models/pull/506

https://github.com/screwdriver-cd/scm-github/pull/190

https://github.com/screwdriver-cd/scm-gitlab/pull/43

https://github.com/screwdriver-cd/data-schema/pull/450

## License


lib/build.js
```diff
@@ -548,7 +564,7 @@ class BuildModel extends BaseModel {
                         jobId: job.id,
                         startTime: this.startTime,
                         annotations: job.permutations[0].annotations,
-                        buildStatus: this.status,
```
jithine: why is config passed to `executor.startTimer` being modified ?

lib/build.js
```diff
@@ -16,6 +16,22 @@ const uiUri = Symbol('uiUri');
 const ABORT_CODE = 130; // 128 + SIGINT 2 (^C)
 const TEMPORAL_JWT_TIMEOUT = 12 * 60; // 12 hours in minutes
 
+// SCM handles only SUCCESS, PENDING & FAILED status
+const SCM_STATE_MAP = {
```
jithine: BLOCKED: 'PENDING',
COLLAPSED: '',
 FROZEN: ''
  FIXED: '',
  ones empty should not be handled

test/lib/build.test.js
```diff
@@ -38,6 +38,22 @@ describe('Build Model', () => {
         }
     };
     const TEMPORAL_JWT_TIMEOUT = 12 * 60;
+    // SCM handles only SUCCESS, PENDING & FAILURE status
+    const SCM_STATE_MAP = {
```
jithine: have this exported from build model. do not redefine

lib/build.js
ombharatiya: Keeping `buildStatus` in `config`  consistent throughout the system. Should executor queue not maintaining a log of various commit status, this can be removed.

jithine: scm status should only be passed to scm functions `updateCommitStatus` I don't get why it's beeing passed to executor.startTimer

jithine (approved):

lib/helper.js
```diff
@@ -5,6 +5,21 @@ const hoek = require('@hapi/hoek');
...
+// SCM handles only SUCCESS, PENDING & FAILURE status
+const SCM_STATE_MAP = {
+    ABORTED: 'FAILURE',
+    CREATED: 'PENDING',
+    FAILURE: 'FAILURE',
+    QUEUED: 'PENDING',
+    RUNNING: 'PENDING',
+    SUCCESS: 'SUCCESS',
+    BLOCKED: 'PENDING',
+    UNSTABLE: 'FAILURE',
```
tkyi: UNSTABLE status means the next build will not be triggered, but does not mean failure; maybe PENDING would be more equivalent
```suggestion
    UNSTABLE: 'PENDING',
```

lib/helper.js
```diff
@@ -5,6 +5,21 @@ const hoek = require('@hapi/hoek');
 const dayjs = require('dayjs');
 const TEMPLATE_NAME_REGEX_WITH_NAMESPACE = schema.config.regex.FULL_TEMPLATE_NAME_WITH_NAMESPACE;
 const formatDate = dateTime => dayjs(dateTime).format('YYYY-MM-DD');
+// SCM handles only SUCCESS, PENDING & FAILURE status
+const SCM_STATE_MAP = {
```
tkyi: is this map kept in models and each scm plugin?

lib/helper.js
```diff
@@ -5,6 +5,21 @@ const hoek = require('@hapi/hoek');
...
+    ABORTED: 'FAILURE',
+    CREATED: 'PENDING',
+    FAILURE: 'FAILURE',
+    QUEUED: 'PENDING',
+    RUNNING: 'PENDING',
+    SUCCESS: 'SUCCESS',
+    BLOCKED: 'PENDING',
+    UNSTABLE: 'FAILURE',
+    PENDING: 'PENDING',
```
tkyi: Is PENDING a valid status?
https://github.com/screwdriver-cd/data-schema/blob/master/models/build.js#L14-L25

tkyi (changes requested):

lib/build.js
ombharatiya: Made the changes accordingly.

lib/helper.js
ombharatiya: Yes, it is handled in each scm plugin(github & gitlab) individually.

tkyi: If that is the case, is there any reason why we need it here then?

lib/helper.js
ombharatiya: PENDING is not a valid build status. Changes are done to remove it from this map.

lib/helper.js
ombharatiya: Mappings:

In Models:
Build Status to Generic SCM Statuses(SUCCESS, PENDING, FAILURE)

In Individual SCM plugins:
Generic SCM Statuses(SUCCESS, PENDING, FAILURE) to Specific SCM Statuses(e.g. FAILURE => failed in GitLab)

pritamstyz4ever: If these are *build statuses* repeating them all over is only going to be confusing.

ombharatiya: Build statuses are repeated all over in existing system and creating confusion. That's the problem we are solving with this change.

We're defining SCM status and handling them differently with proper mappings.

In earlier versions Build Statuses were confused with SCM Statuses and code were written to handle all the build statuses in SCM, but we only need to handle 3 SCM status (SUCCESS, PENDING, FAILURE) and that we're doing with these mappings.

Please refer to @jithine's comment here on it:
https://github.com/screwdriver-cd/data-schema/pull/450#discussion_r656401198

jithine: We need mapping from build status to inidividual SCM status check API values., we also need mapping form meta status check values to individual SCM status check API values.

What's happening with this changes is a mapping from Build status -> meta status check values (handled in models) -> individual check API status values (handled in SCM plugins)

lib/helper.js
```diff
@@ -5,6 +5,20 @@ const hoek = require('@hapi/hoek');
...
+    ABORTED: 'FAILURE',
+    CREATED: 'PENDING',
+    FAILURE: 'FAILURE',
+    QUEUED: 'PENDING',
+    RUNNING: 'PENDING',
+    SUCCESS: 'SUCCESS',
+    BLOCKED: 'PENDING',
+    UNSTABLE: 'PENDING',
+    COLLAPSED: '',
+    FIXED: '',
```
tkyi: this status is only sent to notifications, not a valid build status
https://github.com/screwdriver-cd/data-schema/blob/master/models/build.js#L14-L25

lib/helper.js
jithine: @ombharatiya maybe moving this to data scheme would be a better home for this. https://github.com/screwdriver-cd/data-schema/blob/master/plugins/scm.js

lib/helper.js
ombharatiya: But in the model it is defined as build status. Please check here:
https://github.com/screwdriver-cd/data-schema/blob/master/models/build.js#L113-L116

And, if it's not build status, then where to find only the build statuses?

tkyi: this is where all the build statuses are: https://github.com/screwdriver-cd/data-schema/blob/master/models/build.js#L14-L25

lib/build.js
```diff
@@ -187,12 +188,23 @@ function getStatusConfig({ metadata }) {
             }
         }
 
+        const status = hoek.reach(checkObject, 'status', { default: 'PENDING' });
+
+        if (!['SUCCESS', 'FAILURE', 'PENDING'].includes(status)) {
```
tkyi: Can you use the data-schema: https://github.com/screwdriver-cd/data-schema/blob/master/plugins/scm.js#L354

lib/build.js
```diff
@@ -280,12 +292,17 @@ class BuildModel extends BaseModel {
...
                     scmContext: pipeline.scmContext,
                     sha: this.sha,
-                    buildStatus: this.status,
+                    buildStatus: SCM_STATE_MAP[this.status],
                     jobName: job.name,
                     url: buildUrl,
                     pipelineId: pipeline.id
                 };
 
+                // Skiping handling the empty build status
```
tkyi:
```suggestion
                // Skipping handling the empty build status
```

tkyi (approved):

PR #507, opened 2021-06-22
fix(2437): Read-only child pipelines should use parent admins

tkyi:
## Context

Cannot save read-only user as admin in read-only child pipeline because it is not logged into Screwdriver. We should use the parent's admins instead and inject the proper `token` where appropriate.

## Objective

This PR looks for the child pipeline's parent admins if the child pipeline belongs to a read-only SCM in:
- update
- sync
- addWebhooks
- getFirstAdmin
- get token
- pipelineFactory.create

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2437

## License

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -410,7 +410,6 @@ class PipelineModel extends BaseModel {
      */
     async addWebhooks(webhookUrl) {
         const webhookUrlsWithActionsList = [];
-
         const webhookUrlWithActions = {
             webhookUrl,
             actions: [],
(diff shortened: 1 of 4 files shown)


pritamstyz4ever (approved):

lib/pipeline.js
```diff
@@ -467,43 +476,41 @@ class PipelineModel extends BaseModel {
...
-                        token
-                    })
-                ]).then(([existingJobs, openedPRs]) => {
-                    const synced = [];
+    async syncPRs() {
+        const token = await this.token;
 
-                    openedPRs.forEach(openedPR => {
-                        const prNum = openedPR.name.split('-')[1];
+        return Promise.all([
```
jithine, line 482: we should avoid mixing async/await & promise callbacks, these following lines could also be moved to async/await

jithine (approved):

PR #508, opened 2021-06-23
fix: Omit collecting PR jobs in getBlockedByIds

wahapo:
## Context
If there are a large number of remote pipelines (in the order of 1000) in blockedBy, there will be a large amount of access to SCM.

`screwdriver.yaml`:
```
jobs:
  main:
    blockedBy:
      - ~sd@123:job1
      - ~sd@123:job2
      - ~sd@123:job3
      ......
      - ~sd@123:job1000
      - ~sd@456:job1
      - ~sd@456:job2
      ......
```

PR job should not be necessary for [blockedBy](https://docs.screwdriver.cd/user-guide/configuration/workflow#blocked-by).

> This feature does not apply to PR jobs.

## Objective
Save unnecessary access to SCM.

## License

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -65,7 +65,8 @@ function getBlockedByIds(pipeline, job) {
     // Remove ~ prefix for job names
     blockedByNames = blockedByNames.map(name => name.replace('~', ''));
 
-    return pipeline.getJobs().then(pipelineJobs => {
+    // Without PR jobs
+    return pipeline.getJobs({ type: 'pipeline' }).then(pipelineJobs => {
         // Get internal blocked by first
         blockedByIds = blockedByIds.concat(findIdsOfMatchedJobs(pipelineJobs, blockedByNames));
 
@@ -92,7 +93,7 @@ function getBlockedByIds(pipeline, job) {
                                 return [];
                             }
 
-                            return p.getJobs();
+                            return p.getJobs({ type: 'pipeline' });
                         })
                         .then(jobs => findIdsOfMatchedJobs(jobs, [jobname]));
                 })
--- a/lib/build.js
+++ b/lib/build.js
@@ -65,8 +65,7 @@ function getBlockedByIds(pipeline, job) {
     // Remove ~ prefix for job names
     blockedByNames = blockedByNames.map(name => name.replace('~', ''));
 
-    // Without PR jobs
-    return pipeline.getJobs({ type: 'pipeline' }).then(pipelineJobs => {
+    return pipeline.getJobs().then(pipelineJobs => {
         // Get internal blocked by first
         blockedByIds = blockedByIds.concat(findIdsOfMatchedJobs(pipelineJobs, blockedByNames));
 
@@ -93,6 +92,7 @@ function getBlockedByIds(pipeline, job) {
                                 return [];
                             }
 
+                            // Without PR jobs
                             return p.getJobs({ type: 'pipeline' });
                         })
                         .then(jobs => findIdsOfMatchedJobs(jobs, [jobname]));
--- a/test/lib/build.test.js
+++ b/test/lib/build.test.js
@@ -1181,6 +1181,11 @@ describe('Build Model', () => {
                 id: 222,
                 isPR: () => false
             };
+            const prJob = {
+                name: `PR-999:${blocking2.name}`,
+                isPR: () => true,
+                id: 333
+            };
 
             pipelineMockB = {
                 id: pipelineId,
@@ -1195,7 +1200,8 @@ describe('Build Model', () => {
                         blocking1,
                         { id: 123, name: 'somejob', isPR: () => false },
                         blocking2,
-                        { id: 456, name: 'someotherjob', isPR: () => false }
+                        { id: 456, name: 'someotherjob', isPR: () => false },
+                        prJob
                     ])
             };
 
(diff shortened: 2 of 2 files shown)


kumada626 (approved):

lib/build.js
```diff
@@ -65,7 +65,8 @@ function getBlockedByIds(pipeline, job) {
     // Remove ~ prefix for job names
     blockedByNames = blockedByNames.map(name => name.replace('~', ''));
 
-    return pipeline.getJobs().then(pipelineJobs => {
+    // Without PR jobs
+    return pipeline.getJobs({ type: 'pipeline' }).then(pipelineJobs => {
```
pritamstyz4ever: should we add this as a parameter instead and allow caller to decide whether to get or jobs or not?

pritamstyz4ever: Also, will this work, for the time when PR job is blocked by another PR job —e.g., in case we restart a PR job when one is running

wahapo: This is just called from [start](https://github.com/screwdriver-cd/models/blob/4f62727856f659e8a529fd8f433b7ac9bc4c1835/lib/build.js#L460) and [stop](https://github.com/screwdriver-cd/models/blob/4f62727856f659e8a529fd8f433b7ac9bc4c1835/lib/build.js#L584), it seems to be enough to get the same result in both cases. So there seems to be no need to pass parameters to separate the processes.

The logic of the (PR) build blocking itself is the following part.
https://github.com/screwdriver-cd/models/blob/4f62727856f659e8a529fd8f433b7ac9bc4c1835/lib/build.js#L57

And by the way, with this fix, pipelineJobs will be changed as follows. I found that in the case where `main` is in `blockedBy`, until now we also get the job ID of `PR-1:main`, so the commit build blocked by `main` was also blocked by the `PR-1:main` build. (FYI: [findIdsOfMatchedJobs](https://github.com/screwdriver-cd/models/blob/4f62727856f659e8a529fd8f433b7ac9bc4c1835/lib/build.js#L26-L45)) I think it is a strange phenomenon that commit builds are blocked by PR builds and vice versa. With this fix, PR builds are no longer completely bound to blockedBy.
```
pipelineJobs = [
  Job { ..., name: "main", ... },
//  Job { ..., name: "PR-1:main", ... }    //If set type:'pipeline', this line is omitted
]
```

tkyi (approved):

jithine:
How about chain PR use cases ? Is blocked by not applicable there as well ? Will this lead to a regression if someone with using chain PR was relying on this ?

wahapo:
@jithine If the user using the chain PR relying on blockedBy, it will be a regression.
It would block the commit event and the PR event without distinguishing between them, but we should the blockedBy feature be enabled even in PR builds?

wahapo:
This leaves only the changes that pr and commit build will no longer be blocked by the PR build of the remote pipeline.
Even if users use chainPR, I believe the impact of this change will be minor because PR builds cannot trigger remote pipelines.

wahapo:
Until now, for example, if user has the following `screwdriver.yaml`, the `main` build of PR and commit events is blocked when the `PR-1:remoteJob` build is running in the remote pipeline(1234). With this PR fix, this will not be blocked. The `main` build of PR and commit events will be correctly blocked for a `remoteJob` running on a remote pipeline(1234) commit event.
```
annotations:
  screwdriver.cd/chainPR: true

jobs:
  main:
    requires: [ ~commit, ~pr ]
    image: node:14
    blockedBy: [ ~sd@1234:remoteJob ]
    steps:
      - ls
```

ibu1224 (approved):

wahapo:
@jithine @tkyi @pritamstyz4ever 
I have also created a related issue, however I personally think that this fix is fine for this PR. I hope you can confirm that everything looks fine:pray:
https://github.com/screwdriver-cd/screwdriver/issues/2499

lib/build.js
jithine: I think this issue is very much similar to https://github.com/screwdriver-cd/screwdriver/issues/2496

lib/build.js
```diff
@@ -92,7 +92,8 @@ function getBlockedByIds(pipeline, job) {
                                 return [];
                             }
 
-                            return p.getJobs();
+                            // Without PR jobs
+                            return p.getJobs({ type: 'pipeline' });
```
jithine, line 96: if current job is a PR job, should this return jobs pertaining to that PR ?

lets say something like 
```
if (current job is a PR job)
   return p.getJobs({ type: 'pr', prNum: prNum });
else  
   return p.getJobs({ type: 'pipeline' });
``` 

I think we don't support `p.getJobs({ type: 'pr', prNum: prNum });` right now. Idea behind PR https://github.com/screwdriver-cd/models/pull/509/files is also similar, limit the jobs being fetched for a PR, ideally it should be able to do `p.getJobs({ type: 'pr', prNum: prNum });`

wahapo: I think that works great for line 69. This may take some time to implement.

Line 96 is the process to determine if it is blocked by remote pipeline jobs. Since the jobs retrieved here are from remote pipelines, the fact that the PR number is the same is not important for identification purposes. It is either blocked only on commit events, as in this fix, or it is not blocked at all.

jithine (approved):

lib/build.js
jithine: thanks for the clarification

tkyi (approved):

PR #509, opened 2021-06-24
feat: add flag archived false while fetching jobs

pritamstyz4ever:
## Context

The pipeline model fetches all the jobs (archived and unarchived) from the data-store via the `jobs()` function, leading to perf issues and fetch an unnecessary load of data. This becomes problematic during syncPRs via /v4/webhook API

## Objective

This PR adds a param to skip all archived jobs while fetching the list of jobs.

## References

## License

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -1249,7 +1249,8 @@ class PipelineModel extends BaseModel {
     get jobs() {
         const listConfig = {
             params: {
-                pipelineId: this.id
+                pipelineId: this.id,
+                archived: false
             }
         };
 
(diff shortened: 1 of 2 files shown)


jithine:
This will impact the archive <-> unarchive transition for close/open PRs. While I do agree that this change is needed, we need to find an alternative solution for PR jobs

@screwdriver-cd/yahoo-japan  this change can single handedly improve webhook performance by a lot. One repo where we found POST /webhooks was crashing the API pod, had 70k jobs, of which only 150 were non-archived

tkyi:
Will this prevent archiving to happen correctly?
https://github.com/screwdriver-cd/models/blob/master/lib/pipeline.js#L577

jithine:
This needs a comprehensive fix which address archive/unarchive job use cases

Filed https://github.com/screwdriver-cd/screwdriver/issues/2496 to track.

PR #510, opened 2021-07-02
fix(2503): limit concurrent process number when pipelines are removed

itleigns:
## Context
When large pipelines are erased, the number of concurrent processes becomes too large and API is crashed. We want to limit the number of concurrent processes.

## Objective
We limit the number of concurrent processes to delete jobs and events and change that of builds.

## References
https://github.com/screwdriver-cd/screwdriver/issues/2503

## License

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -9,7 +9,7 @@ const executor = Symbol('executor');
 const tokenGen = Symbol('tokenGen');
 const apiUri = Symbol('apiUri');
 const START_INDEX = 3;
-const MAX_COUNT = 1000;
+const MAX_METRIC_GET_COUNT = 1000;
 const MAX_BUILD_DELETE_COUNT = 100;
 const DEFAULT_COUNT = 10;
 const DEPLOY_KEY_SECRET = 'SD_SCM_DEPLOY_KEY';
@@ -342,7 +342,7 @@ class Job extends BaseModel {
             sort: 'ascending',
             sortBy: 'id',
             paginate: {
-                count: MAX_COUNT
+                count: MAX_METRIC_GET_COUNT
             },
             readOnly: true
         };
--- a/lib/job.js
+++ b/lib/job.js
@@ -10,6 +10,7 @@ const tokenGen = Symbol('tokenGen');
 const apiUri = Symbol('apiUri');
 const START_INDEX = 3;
 const MAX_COUNT = 1000;
+const MAX_BUILD_DELETE_COUNT = 100;
 const DEFAULT_COUNT = 10;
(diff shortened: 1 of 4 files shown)


lib/job.js
```diff
@@ -279,7 +280,7 @@ class Job extends BaseModel {
     remove() {
         const removeBuilds = () =>
             // Iterate through the builds and remove MAX_COUNT at a time
```
wahapo:
```suggestion
            // Iterate through the builds and remove MAX_BUILD_DELETE_COUNT at a time
```

wahapo (approved):

itleigns:
I do not know what is the appropriate value for the maximum number of concurrent processes. I will change it if some other value seems better.

ibu1224 (approved):

lib/job.js
```diff
@@ -10,6 +10,7 @@ const tokenGen = Symbol('tokenGen');
 const apiUri = Symbol('apiUri');
 const START_INDEX = 3;
 const MAX_COUNT = 1000;
```
tk3fftk: This can also be more detailed name :pray:

lib/pipeline.js
```diff
@@ -17,6 +17,8 @@ const REGEX_CAPTURING_GROUP = {
     job: 2 // main or undefined if using legacy
 };
 const MAX_COUNT = 1000;
```
tk3fftk: This can also be more detailed name 🙏

lib/job.js
itleigns: I changed.

lib/pipeline.js
itleigns: I changed.

tk3fftk (approved):

jithine (approved):

lib/pipeline.js
```diff
@@ -16,7 +16,9 @@ const REGEX_CAPTURING_GROUP = {
     pr: 1, // PR-1
     job: 2 // main or undefined if using legacy
 };
-const MAX_COUNT = 1000;
+const MAX_METRIC_GET_COUNT = 1000;
+const MAX_JOB_DELETE_COUNT = 10;
```
tkyi, line 20: wondering why 10 for job delete, but it's okay we can change later

tkyi (approved):

lib/pipeline.js
itleigns: This is because the maximum number of concurrent processes is `MAX_JOB_DELETE_COUNT` * `MAX_BUILD_DELETE_COUNT` when pipelines are removed. We expect one of the two value to be smaller than 100.

PR #514, opened 2021-09-24
feat(2346): Support for job level parameters

sagar1312:
## Context

Ability to define parameters at job level

## Objective

Create builds with job level parameters

## References

https://github.com/screwdriver-cd/screwdriver/issues/2346

## License

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -10,6 +10,7 @@ const {
 } = require('screwdriver-data-schema').config.regex;
 const workflowParser = require('screwdriver-workflow-parser');
 const logger = require('screwdriver-logger');
+const hoek = require('@hapi/hoek');
 const BaseFactory = require('./baseFactory');
 const Event = require('./event');
 let instance;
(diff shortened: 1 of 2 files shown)


lib/eventFactory.js
```diff
@@ -473,53 +474,100 @@ function updateWorkflowGraph(config) {
...
- * @method updateEventParameters
+ * Determines the parameters for the build by compiling the list of parameters with default values from the
+ * `defaultParameters` and override the values if specified in `eventParameters`. Parameters from `eventParameters` are
+ * dropped if matching definition is not found in `defaultParameters`.
+ * @param {Object} defaultParameters Default Pipeline or Job parameter definitions
+ * @param {Object} eventParameters   Customized build parameters
+ * @resolves {Object}
+ */
+function validateAndMergeParameters(defaultParameters, eventParameters) {
+    const allowedParameters = Object.create(null);
```
tkyi, line 485: wondering why we're using `Object.create(null)` vs `{}`, but I see that it was used in updateEventParameters() already

lib/eventFactory.js
```diff
@@ -473,53 +474,100 @@ function updateWorkflowGraph(config) {
...
+                }
+
+                if (actualParameterValue !== undefined) {
+                    allowedParameters[name] = { value: actualParameterValue };
+                }
+            }
+        });
+    }
+
+    return allowedParameters;
```
tkyi, line 517: will this return an Object vs null if `defaultParameters` does not exist?

lib/eventFactory.js
jithine: apparently quite interesting side effects https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create#custom_and_null_objects

lib/eventFactory.js
jithine: it's an object, but a bit special https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create#custom_and_null_objects

jithine (approved):

PR #522, opened 2022-01-21
fix: return 403 when pipeline has no admin

yk634:
## Context

When a pipeline does not have an administrator, the 500 error is returned.
However, since this is an authorization error caused by users, I think that the 403 error should be returned.
This change will allow the error data to be aggregated correctly.

## Objective

Use `hapi/boom.forbidden` to throw the error.

## References

Example of status code control using `hapi/boom`
https://github.com/screwdriver-cd/models/blob/18b0bfaab3bdb15358f40a38ec668997d7c89973/lib/build.js#L753-L756

## License

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -5,6 +5,7 @@
 
 const parser = require('screwdriver-config-parser');
 const workflowParser = require('screwdriver-workflow-parser');
+const boom = require('@hapi/boom');
 const hoek = require('@hapi/hoek');
 const dayjs = require('dayjs');
 const _ = require('lodash');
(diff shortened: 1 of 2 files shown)


jithine (approved):

tkyi (approved):

jithine:
@yk634 I think for some weird reason, we are seeing 404 for response code

```
"pipelineId:XXX: Pipeline has no admin."
[response,api,pipelines] https://api.screwdriver.....com: [1;32mget[0m /v4/pipelines/XXX/admin 
{}
 [33m404[0m (262ms)
```

PR #534, opened 2022-05-05
fix: do not auto delete obsolete child pipelines

sagar1312:
## Context

Child pipelines are programmatically deleted when there is only domain change in the SCM URI of child pipelines. This is not recoverable and lead to loss of build history of child pipelines.

## Objective

Stop deleting child pipelines programmatically

## References

https://github.com/screwdriver-cd/screwdriver/issues/2580

## License

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -837,58 +837,6 @@ class PipelineModel extends BaseModel {
     }
 
     /**
-     * Remove a pipeline given a scmUrl
-     * @method _removePipeline
-     * @param {String} scmUrl  checkout url for a repository
-     * @return {Promise}
-     */
-    async _removePipeline(scmUrl) {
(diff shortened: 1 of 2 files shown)


lib/pipeline.js
```diff
@@ -897,15 +897,15 @@ class PipelineModel extends BaseModel {
     async _syncChildPipelines(scmUrls) {
         const toCreateOrUpdate = [];
         const toRemove = [];
-        const oldScmUrls = hoek.reach(this.childPipelines, 'scmUrls') || [];
+        // const oldScmUrls = hoek.reach(this.childPipelines, 'scmUrls') || [];
```
pritamstyz4ever: probably, we should add some comments explaining why the code is commented.

jithine (approved):

PR #537, opened 2022-05-13
fix(2704): Do not run PR job if original job is disabled

tkyi:
## Context

Sometimes a user might disable a job in the main workflow. We currently don't allow users to individually disable PR jobs, but we are not handling when a PR job shouldn't run when its original job is disabled.

## Objective

This PR checks if the original job is disabled before trying to start a PR job.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2704

## License

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -5,7 +5,7 @@ const dayjs = require('dayjs');
 const hoek = require('@hapi/hoek');
 const deepcopy = require('deepcopy');
 const logger = require('screwdriver-logger');
-const { PR_JOB_NAME, EXTERNAL_TRIGGER } = require('screwdriver-data-schema').config.regex;
+const { EXTERNAL_TRIGGER } = require('screwdriver-data-schema').config.regex;
 const { SCM_STATE_MAP, SCM_STATUSES } = require('screwdriver-data-schema').plugins.scm;
 const BaseModel = require('./base');
 
@@ -26,16 +26,7 @@ const TEMPORAL_JWT_TIMEOUT = 12 * 60; // 12 hours in minutes
  */
 function findIdsOfMatchedJobs(jobs, jobNames) {
     return jobs.reduce((match, j) => {
-        let { name } = j;
-
-        // Get the actual job name if the matching job is a PR job
-        if (j.isPR()) {
-            const prJobName = j.name.match(PR_JOB_NAME);
-
-            if (prJobName) {
-                name = prJobName[2];
-            }
-        }
+        const name = j.parsePRJobName('job') || j.name;
 
         if (jobNames.includes(name)) {
             match.push(j.id);
--- a/lib/helper.js
+++ b/lib/helper.js
@@ -3,23 +3,8 @@
 const schema = require('screwdriver-data-schema');
 const hoek = require('@hapi/hoek');
 const dayjs = require('dayjs');
-const TEMPLATE_NAME_REGEX_WITH_NAMESPACE = schema.config.regex.FULL_TEMPLATE_NAME_WITH_NAMESPACE;
 const formatDate = dateTime => dayjs(dateTime).format('YYYY-MM-DD');
-const PR_JOB_NAME_REGEX = schema.config.regex.PR_JOB_NAME;
-
-/**
- * Get the original job name and PR number from full PR job name
- * @param  {String} name Full PR job name (e.g.: PR-100:main)
- * @return {Object}      PR number and job name
- */
-function parsePRJobName(name) {
-    const [, fullPrNum, jobName] = PR_JOB_NAME_REGEX.exec(name);
-
-    return {
-        prNum: fullPrNum,
-        name: jobName
-    };
-}
+const TEMPLATE_NAME_REGEX_WITH_NAMESPACE = schema.config.regex.FULL_TEMPLATE_NAME_WITH_NAMESPACE;
 
 /**
  * Get the value of the annotation that matches name
@@ -239,7 +224,6 @@ function getToken(fn, pipeline, jobId) {
 }
 
 module.exports = {
-    parsePRJobName,
     getAnnotations,
     parseTemplateConfigName,
     getAllRecords,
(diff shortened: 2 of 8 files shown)


lib/helper.js
```diff
@@ -5,6 +5,21 @@ const hoek = require('@hapi/hoek');
...
 const TEMPLATE_NAME_REGEX_WITH_NAMESPACE = schema.config.regex.FULL_TEMPLATE_NAME_WITH_NAMESPACE;
 const formatDate = dateTime => dayjs(dateTime).format('YYYY-MM-DD');
+const PR_JOB_NAME_REGEX = schema.config.regex.PR_JOB_NAME;
+
+/**
+ * Get the original job name and PR number from full PR job name
+ * @param  {String} name Full PR job name (e.g.: PR-100:main)
+ * @return {Object}      PR number and job name
+ */
+function parsePRJobName(name) {
```
sagar1312: Looks like we have similar implementation in [`pipeline`](https://github.com/screwdriver-cd/models/blob/master/lib/pipeline.js#L430) model. 

May be we could consolidate and move these to `job` model. Similar to [`isPR()` and `prNum`](https://github.com/screwdriver-cd/models/blob/master/lib/job.js#L134) in the `job` model.

sagar1312 (changes requested):

lib/helper.js
tkyi: Yeah that was the original intent of the helper function
was going to do that in a follow up PR

sagar1312: This function need not be in generic helper. This seems to be specific to `job`, can we abstract this into `job` model and expose derived attributes or methods on the `job` model.

This would remove redundant implementation we have in other models
- [`pipeline`]([`pipeline`](https://github.com/screwdriver-cd/models/blob/master/lib/pipeline.js#L430))
- [`build`](https://github.com/screwdriver-cd/models/blob/master/lib/build.js#L33)

sagar1312: This function need not be in generic helper. This seems to be specific to `job`, can we abstract this into `job` model and expose derived attributes or methods on the `job` model.

This would remove redundant implementation we have in other models
- [`pipeline`](https://github.com/screwdriver-cd/models/blob/master/lib/pipeline.js#L430)
- [`build`](https://github.com/screwdriver-cd/models/blob/master/lib/build.js#L33)

lib/pipeline.js
```diff
@@ -444,7 +429,7 @@ class PipelineModel extends BaseModel {
 
         existingPRs.forEach(job => {
             // getting PR and number e.g. PR-1
-            const prName = this._getPartialJobName(job.name, 'pr');
+            const prName = job.parsePRJobName('pr');
```
sagar1312, line 432:
```suggestion
            const prName = job.prName;
```

lib/build.js
```diff
@@ -30,11 +30,7 @@ function findIdsOfMatchedJobs(jobs, jobNames) {
 
         // Get the actual job name if the matching job is a PR job
         if (j.isPR()) {
```
sagar1312: This can be reduced to 
```
const name = j.prParentJobName || j.name;

if (jobNames.includes(name)) {
    match.push(j.id);
}

return match;
```

lib/job.js
```diff
@@ -127,12 +131,22 @@ class Job extends BaseModel {
...
 
+    /**
+     * Get part of a PR job name based on given type
+     * @method parsePRJobName
+     * @param  {String}    type         Type can either be pr or job, e.g. pr => PR-1, job => main
+     * @return {String}                 Substring of a PR job name based on given type
+     */
+    parsePRJobName(type) {
+        return this.name.match(PR_JOB_NAME)[REGEX_CAPTURING_GROUP[type]];
+    }
```
sagar1312: Can we split this into two separate functions one for each type, so that we need not specify the type when used?
Also this could throw error if invoked on non PR job name. Ex: "component". We should check and return `null` if it is not a PR job
```suggestion
    get prName() {
        return this.isPR() ? this.name.match(PR_JOB_NAME)[REGEX_CAPTURING_GROUP['pr']] : null;
    }
    
    get prParentJobName() {
        return this.isPR() ? this.name.match(PR_JOB_NAME)[REGEX_CAPTURING_GROUP['job']] : null;
    }
```

lib/job.js
```diff
@@ -127,12 +131,22 @@ class Job extends BaseModel {
...
+
     /**
      * Return PR number. Returns null if this is not a PR job
      * @return {Number}     PR number
      */
     get prNum() {
-        return this.isPR() ? parseInt(this.name.slice(START_INDEX), 10) : null;
+        return this.isPR() ? this.parsePRJobName('pr').split('-')[1] : null;
```
sagar1312: We need to return number
```suggestion
        return this.isPR() ? parseInt(this.prName.split('-')[1], 10) : null;
```

lib/job.js
```diff
@@ -3,12 +3,16 @@
...
+const { PR_JOB_NAME } = require('screwdriver-data-schema').config.regex;
 const BaseModel = require('./base');
 const { getAnnotations, getAllRecords, getToken } = require('./helper');
 const executor = Symbol('executor');
 const tokenGen = Symbol('tokenGen');
 const apiUri = Symbol('apiUri');
-const START_INDEX = 3;
+const REGEX_CAPTURING_GROUP = {
+    pr: 1, // PR-1
+    job: 2 // main or undefined if using legacy
```
sagar1312: We need to update the comment. `null` or `undefined` is overwritten to "main" in `pipeline` model

sagar1312 (changes requested):

lib/job.js
tkyi: ?

lib/job.js
tkyi: I don't think it makes sense to split this into two functions when they're almost literally the exact same code. We should keep the generic one and just pass in the type

lib/pipeline.js
tkyi: see my other comment

lib/job.js
sagar1312: The logic to return "main" when job name is `undefined` is in [`pipeline` model](https://github.com/screwdriver-cd/models/pull/537/files#diff-2977a16e13f278992ce964024e936111d8f1c3b68f5f698c7e06195258912721R455). Having this comment in `job` model could be confusing as `job` model does not do any overrides by itself.

DekusDenial (approved):

sagar1312 (approved):

PR #539, opened 2022-06-01
fix: Set DISABLED state for new job if screwdriver.cd/jobDisabledByDefault annotation is true [2]

tkyi:
## Context

Sometimes users might have repos/pipelines with build periodic configs. If developers fork and create their own pipelines, they might not want these cron jobs to get auto triggered. It would be nice to have an additional annotation for disabling jobs upon creation so users are required to consciously enable them if they want the periodic builds.

## Objective

This PR sets a new job's state to DISABLED upon job creation if `screwdriver.cd/jobDisabledByDefault: true`.

## References

Blocked by https://github.com/screwdriver-cd/data-schema/pull/488

## License

### DIFF
--- a/lib/jobFactory.js
+++ b/lib/jobFactory.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const BaseFactory = require('./baseFactory');
-const { getAnnotations, getToken } = require('./helper');
+const { getAnnotations, getToken, convertToBool } = require('./helper');
 const Job = require('./job');
 const { getQueries, PR_JOBS_FOR_PIPELINE_SYNC } = require('./rawQueries');
 let instance;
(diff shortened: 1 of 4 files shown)


sagar1312 (approved):

lib/jobFactory.js
```diff
@@ -55,8 +55,11 @@ class JobFactory extends BaseFactory {
      */
     create(config) {
         const c = config;
+        const jobDisabledByDefault = convertToBool(
+            getAnnotations(c.permutations[0], 'screwdriver.cd/jobDisabledByDefault')
+        );
 
-        c.state = 'ENABLED';
+        c.state = jobDisabledByDefault === true ? 'DISABLED' : 'ENABLED';
```
pritamstyz4ever, line 62: should we call it `enabled` instead and by default it will be true, users can manually set it to false?

sagar1312: Either option looks good. I believe what really matters is the default value based on how we name them.

tkyi: Talked with @pritamstyz4ever  offline and we agreed to leave it as is for now
open to changes in the future if we prefer `jobEnabledByDefault`, `defaultJobState`, etc

PR #542, opened 2022-06-08
fix(2496): Pull request sync fails when there are no open pull requests

sagar1312:
## Context

Pull request sync workflow uses raw DB query to fetch the jobs associated with the pull requests that need to synced. Query expects 2 parameters
- pipeline identifier
- an array of open pull request names (Ex: `[ 'PR-23', 'PR-32']`)

Query for MySQL DB
```
SELECT *
	FROM  `jobs`
        WHERE pipelineId = :pipelineId
        AND prParentJobId IS NOT NULL
        AND (archived = false OR (SUBSTRING(name, 1, POSITION(':' IN name) - 1)) IN (:prNames))
        ORDER BY id ASC
```

Job will be included if one of the below criteria is met
- unarchived
- archived but associated with the specified pull request names

This query resulted in SQL error when there are no pull requests.
```
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '))
        ORDER BY id ASC' at line 5
```

This is caused by empty parenthesis for the `IN` operator in the interpolated query 
```
SELECT *
	FROM  `jobs`
        WHERE pipelineId = 1
        AND prParentJobId IS NOT NULL
        AND (archived = false OR (SUBSTRING(name, 1, POSITION(':' IN name) - 1)) IN ())
        ORDER BY id ASC
```

## Objective
This PR overrides the `prNames` param to `null` when there are no open pull requests. Thus fixing the interpolated query to 
```
SELECT *
	FROM  `jobs`
        WHERE pipelineId = 1
        AND prParentJobId IS NOT NULL
        AND (archived = false OR (SUBSTRING(name, 1, POSITION(':' IN name) - 1)) IN (NULL))
        ORDER BY id ASC
```

## References

https://github.com/screwdriver-cd/models/pull/538

## License

### DIFF
--- a/test/lib/jobFactory.test.js
+++ b/test/lib/jobFactory.test.js
@@ -298,7 +298,6 @@ describe('Job Factory', () => {
 
     describe('getPullRequestJobsForPipelineSync', () => {
         let config;
-        let expectedQueryConfig;
 
         beforeEach(() => {
             sinon.stub(JobFactory.prototype, 'query').returns();
@@ -308,13 +307,9 @@ describe('Job Factory', () => {
             };
         });
 
-        afterEach(() => {
-            assert.calledWith(datastore.query, expectedQueryConfig);
-        });
-
         it('returns pull request jobs for specified pipelineId when there are open pull requests', () => {
             config.prNames = ['PR-2', 'PR-3'];
-            expectedQueryConfig = {
+            const expectedQueryConfig = {
                 queries: getQueries('', PR_JOBS_FOR_PIPELINE_SYNC),
                 replacements: {
                     pipelineId: config.pipelineId,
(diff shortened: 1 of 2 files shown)


test/lib/jobFactory.test.js
```diff
@@ -298,64 +298,137 @@ describe('Job Factory', () => {
...
 
             config = {
-                pipelineId: '12345',
-                prNames: ['PR-2', 'PR-3']
+                pipelineId: '12345'
+            };
+        });
+
+        afterEach(() => {
+            assert.calledWith(datastore.query, expectedQueryConfig);
```
tkyi: we don't usually call assert in the afterEach hook
maybe move the assert to each test?

sagar1312: Done

tkyi (approved):

PR #545, opened 2022-06-13
feat(2669): Add stage model [3]

tkyi:
## Context

Users might want to group jobs together visually in the UI into stages.

## Objective

This PR adds *stages* model. Also called `syncStages()` during pipeline `sync()`.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2669

## License

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -233,6 +233,80 @@ function syncExternalTriggers(config) {
     });
 }
 
+/**
+ * Sync stages
+ * Update stage status to ARCHIVED if it is no longer in yaml
+ * Update stage fields if changes are made
+ * Create the stages if it is defined and was not in the database yet
+ * @method syncStages
+ * @param  {Object}     config              config
+ * @param  {Array}      config.pipelineJobs Pipeline jobs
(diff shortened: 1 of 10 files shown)


adong (approved):

lib/pipeline.js
```diff
@@ -899,12 +1004,14 @@ class PipelineModel extends BaseModel {
...
 
         const existingJobs = await this.pipelineJobs;
-
         const jobsProcessed = [];
         const updatedJobs = [];
         const parsedConfigJobNames = Object.keys(parsedConfig.jobs);
         const pipelineId = this.id;
 
+        // Sync pipeline stages
+        await syncStages({ pipelineJobs: existingJobs, pipelineId, stages: parsedConfig.stages || [] });
```
adong, line 1013: Should we do a check here to see if users have `parsedConfig.stages` in their sd.yaml first before doing an `await`?

tkyi: that won't work because there might be existing stages in the DB but none in the yaml that should be `ARCHIVED`

PR #546, opened 2022-06-17
fix(2704): PR job without original job to work.

yk634:
## Context

Disabled jobs are no longer executed in PR builds at https://github.com/screwdriver-cd/models/pull/537.

However, jobs for which the original does not exist, such as newly created jobs in PR, are not working.
You are probably getting the following error.
`TypeError: Cannot read property 'state' of undefined`

## Objective

- Check for the existence of the original job and handle it as ENABLED or notArchived if it does not exist.
- According to the following fixes, the state should always refer to the original.
  - https://github.com/screwdriver-cd/ui/pull/795

## License


tkyi (approved):
Thank you! 🙇‍♀️

PR #549, opened 2022-06-27
fix(2720): fix get pipelineJobs

itleigns:
## Context
Every time the [sync function](https://github.com/screwdriver-cd/models/blob/e8df4c43843d430a51162cb62a00a5365a340095/lib/pipeline.js#L955) is called, PR jobs without `parentJobId` are archived. This is because the getter of `pipeline.js` named `pipelineJobs` gets not only non pull request jobs but also pull request jobs without `parentJobId`.

## Objective
We fix the `pipelineJobs` to only get non pull request job. 

## References
https://github.com/screwdriver-cd/screwdriver/issues/2720#issuecomment-1165358293

## License

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -1394,7 +1394,9 @@ class PipelineModel extends BaseModel {
         /* eslint-enable global-require */
         const factory = JobFactory.getInstance();
 
-        const jobs = factory.list(listConfig);
+        const jobs = factory.list(listConfig).then(value => {
+            return value.filter(j => !j.name.startsWith('PR-'));
+        });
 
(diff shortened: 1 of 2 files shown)


itleigns:
PR build does not work because `negative substring length not allowed` error occurs.

itleigns:
The following can be omitted from the standpoint of correct behavior, but I leave it to reduce the number of jobs to be fetched.
https://github.com/screwdriver-cd/models/blob/5aa2b8d879f9d846244215515eafe2d935bdbc55/lib/pipeline.js#L1386

sagar1312 (approved):

PR #550, opened 2022-06-27
fix(2720): Jobs fetch query for pull request sync fails

itleigns:
## Context
The following error occurs even after #548.
`negative substring length not allowed`

negative_error

## Objective
We update the raw SQL queries to prevent the error. 

## References
As the following documentation shows, in postgresql, bool expressions are not always evaluated from the left.
https://www.postgresql.org/docs/14/sql-expressions.html#SYNTAX-EXPRESS-EVAL

## License

### DIFF
--- a/lib/rawQueries.js
+++ b/lib/rawQueries.js
@@ -48,22 +48,28 @@ const Queries = {
     getPRJobsForPipelineSyncQuery: (tablePrefix = '') => `SELECT *
         FROM "${tablePrefix}jobs"
         WHERE "pipelineId" = :pipelineId
-        AND name LIKE 'PR-%'
-        AND (archived = 0 OR (SUBSTR(name, 1, INSTR(name, ':')-1) IN (:prNames)))
+        AND CASE 
+            WHEN name LIKE 'PR-%' THEN (archived = 0 OR (SUBSTR(name, 1, INSTR(name, ':')-1) IN (:prNames)))
+            ELSE 0
+        END
         ORDER BY "id" ASC`,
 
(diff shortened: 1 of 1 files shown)


itleigns:
PR build does not work because `negative substring length not allowed` error occurs.

sagar1312:
@itleigns You are noticing this error against which database?

tkyi (approved):

itleigns:
I notice this error against the following pipeline.
`https://cd.screwdriver.cd/pipelines/9/`.
So PostgreSQL is used.
Kumada626 took this picture.

sagar1312:
I believe the changes proposed in this PR may not fix the issue.
I have ran the query (pull request sync API) against few pipelines and it works as expected.

I suspect, few of the pipelines associated with screwdriver projects have some archived jobs (now archived) with name prefixed with `PR-` but does not match the latest format for PR job names `PR-<PR_NUMBER>:<PARENT_JOB_NAME>`.

itleigns:
@sagar1312 
I understand.
I would appreciate it if you could delete the non pull request job starting with `PR-` from database.

According to the documentation I put in the references, it looks like the current query is not guaranteed to work.
Is there any reason to leave it as it is now?

sagar1312:
The still exist with this fix.
As suspected found job names which violate the rules. For Ex: found these job names for https://cd.screwdriver.cd/pipelines/9
```
PR-147
PR-148
PR-149
PR-150
PR-151
PR-152
PR-153
PR-154
PR-155
PR-156
```

Will clean up database to delete these jobs.

sagar1312:
Here is the DB migration to remove pull request jobs with invalid name
https://github.com/screwdriver-cd/data-schema/pull/495

PR #552, opened 2022-07-01
fix(2581): Set parentEventId even in case PR

itleigns:
## Context
Currently restarted PR events do not fetch the metadata set in the parent event.

## Objective
We set `parentEventId` to events even in case PR event. Then metadata of parent event will be fetched [here](https://github.com/screwdriver-cd/launcher/blob/c566f097bb9bceb248799774f8520cf4f3e12ba6/launch.go#L488-L498).

## References
https://github.com/screwdriver-cd/screwdriver/issues/2581#issuecomment-1172195954

## License

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -714,10 +714,12 @@ class EventFactory extends BaseFactory {
                 // Sync pipeline to make sure workflowGraph is generated
                 .then(p => {
                     prevChainPR = p.chainPR;
+                    if (parentEventId) {
+                        modelConfig.parentEventId = parentEventId;
+                    }
+
                     // Sync pipeline with the parentEvent sha and create jobs based on that sha
                     if (parentEventId && !prRef) {
-                        modelConfig.parentEventId = parentEventId;
-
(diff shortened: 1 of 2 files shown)


sagar1312:
Is there a related UI change?
UI doesn't seem to set `parentEventId` in the payload when a PR job is restarted. Only `groupEventId` is set.

image

tkyi:
probably just need to add `parentEventId` here: https://github.com/screwdriver-cd/ui/blob/master/app/pipeline/events/controller.js#L573-L584

itleigns:
It seems `parentEventId` is already set in the payload.
chain-pr-payload

itleigns:
@sagar1312 Preview tab displays the response of API. There is no `parentEventId` in the response because the modification of this PR is not reflected.

itleigns:
And `parentEventId` is not necessarily required for payload.
In the case of Restart, it is taken [here](https://github.com/screwdriver-cd/screwdriver/blob/af926503392c3775f2b8f3a9b304e829495d73b6/plugins/events/create.js#L45).

tkyi (approved):

PR #554, opened 2022-07-15
fix(2669): Create new stage or use old stage based off groupEventId [2]

tkyi:
## Context

Since the UI needs to show old stage information in current and past events, it would be better to keep track of stages based off of `name`, `pipelineId`, and `groupEventId` in order to determine a unique row.
There was also feedback to keep user UI settings such as `color` from the screwdriver.yaml config file.
Adding `createTime` will help keep track of "latest" stages.

Since stages will be more tied to events, we no longer need `state` to keep track of "active" or "archived" stages, they will always be created if they are new or we will refer to an older entry if they are not. We will no longer update stage entries in the DB.

See design doc for more info: https://github.com/screwdriver-cd/screwdriver/pull/2733

## Objective

This PR moves stage creation to event create and removes from pipeline sync. Also removes `color` and `state` fields from stage.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2669

## License

### DIFF
--- a/README.md
+++ b/README.md
@@ -679,9 +679,8 @@ factory.create(config).then(model => {
 | config.pipelineId | Number | Yes | Pipeline that this stage belongs to |
 | config.name | String | Yes | Stage name |
 | config.jobIds | Array | No | Jobs IDs that belong to this stage. Default `[]`. |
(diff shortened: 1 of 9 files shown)


pritamstyz4ever (approved):
LGTM

PR #555, opened 2022-07-21
fix(2728): add handling for PR jobs

y-oksaku:
## Context

PR builds are no longer executed when original jobs is disabled at https://github.com/screwdriver-cd/models/pull/537 .

However, disabled PR builds are executed from the restart button on the detail page (https://github.com/screwdriver-cd/screwdriver/issues/2728).

## Objective

- Check the original job state of the job triggered from the PR job name
  - This check should be the same as a normal PR build.

Click on the restart button to see the following image.

fixed

## References

- Issues
  - https://github.com/screwdriver-cd/screwdriver/issues/2728

## License

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -98,7 +98,24 @@ function getJobsFromJobName(config) {
         return [];
     }
 
-    return jobs.filter(j => j.name === startFrom && j.state === 'ENABLED' && !j.archived);
+    return jobs.filter(j => {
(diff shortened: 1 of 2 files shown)


ibu1224 (approved):
LGTM

tkyi (commented):
Wonder if this is one of those cases where we should not display/should gray-out the Restart button if it is not a legal action
but we can discuss that in a future issue

tkyi (approved):

PR #561, opened 2022-09-01
feat: add reset in user settings

banusha2:
## Context

add a feature to reset in user settings.

## Objective

User can reset his details in user settings.

## License

### DIFF
--- a/lib/user.js
+++ b/lib/user.js
@@ -124,6 +124,16 @@ class UserModel extends BaseModel {
             return user.settings;
         });
     }
+
+    /**
+     * Remove user settings
+     * @method removeSettings
+     * @return {Promise}
+     */
+     removeSettings(){
+
+        return super.remove();
+    }
 }
 
 module.exports = UserModel;
--- a/lib/user.js
+++ b/lib/user.js
@@ -130,9 +130,12 @@ class UserModel extends BaseModel {
      * @method removeSettings
      * @return {Promise}
      */
-     removeSettings(){
+    removeSettings() {
+        this.settings = {};
 
-        return super.remove();
+        return super.update().then(user => {
+            return user.settings;
+        });
     }
 }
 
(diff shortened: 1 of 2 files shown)


lib/user.js
```diff
@@ -124,6 +124,16 @@ class UserModel extends BaseModel {
...
     }
+
+    /**
+     * Remove user settings
+     * @method removeSettings
+     * @return {Promise}
+     */
+     removeSettings(){
+
+        return super.remove();
```
adong: 🚫Warning. This will remove the user itself, not `user.settings`

adong (changes requested):

lib/user.js
adong: I would do something similar to `updateSettings`.

```js
removeSettings() {
    this.settings = {};
    return super.update().then(user => {
        return user.settings;
    });
}
```

test/lib/user.test.js
```diff
@@ -242,6 +242,20 @@ describe('User Model', () => {
         });
     });
 
+    describe('removeSettings', () => {
+        beforeEach(() => {
+            user.settings = {};
```
adong: Can you add some data other than `{}`?
If the old data were `{}`, and the expected new data is also `{}`, then we dont really know if `removeSettings` worked or not.

I have sample data like

```
const userSettingsMock = {
  1018240: {
    showPRJobs: true
  },
  1048190: {
      showPRJobs: false
  },
  displayJobNameLength: 30
};
```

adong (changes requested):

test/lib/user.test.js
```diff
@@ -242,6 +242,28 @@ describe('User Model', () => {
...
+                    showPRJobs: false
+                },
+                displayJobNameLength: 25
+            };
+        });
+
+        it('Remove user settings', () => {
+            datastore.update.resolves({});
+
+            return user.removeSettings(settings).then(data => {
```
adong: `user.removeSettings(settings)` => `user.removeSettings()`

Latter does not take parameters

adong (approved):

PR #570, opened 2022-12-21
feat(2767): Add stage setup and teardown jobIds [4]

tkyi:
## Context

Stages might want to always run initialization or cleanups needed before jobs in a stage can be executed.

## Objective

This PR sets stage setup/teardown jobIds.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2767

## License


lib/event.js
```diff
@@ -15,8 +15,59 @@ class EventModel extends BaseModel {
...
+        const StageBuildFactory = require('./stageBuildFactory');
+        /* eslint-enable global-require */
+        const stageBuildFactory = StageBuildFactory.getInstance();
+        const stageBuildPromises = [];
+
+        stages.forEach(stage => {
+            stageBuildPromises.push(stageBuildFactory.get({ params: { stageId: stage.id, eventId: this.id } }));
+        });
+
+        const stageBuilds = await Promise.all(stageBuildPromises);
```
sagar1312:
```suggestion
        const stageBuilds = await stageBuildFactory.list({params: {eventId: this.id}});
```

lib/eventFactory.js
```diff
@@ -453,50 +455,88 @@ async function convertStages(config) {
...
         });
 
+        newStage.eventId = eventId;
         newStage.jobIds = jobIds;
+
         delete newStage.jobs; // extra field from yaml parser
 
+        // Check for setup and teardown
+        const setupJobName = `${STAGE_PREFIX}${key}:setup`;
+        const teardownJobName = `${STAGE_PREFIX}${key}:teardown`;
```
sagar1312: Can the format for setup and teardown jobs be abstracted into to helper function?
```suggestion
        const setupJobName = getStageSetupJobName(key);
        const teardownJobName = getStageTeardownJobName(key);
```
Instead of hardcoding it in multiple places. Ex:

- https://github.com/screwdriver-cd/screwdriver/pull/2870/files#diff-e01d96efa0729764af51697e9dd42865efba14cb4562ebcb4207cb75528b4ffcR855
- https://github.com/screwdriver-cd/screwdriver/pull/2870/files#diff-e01d96efa0729764af51697e9dd42865efba14cb4562ebcb4207cb75528b4ffcR877

lib/eventFactory.js
```diff
@@ -453,50 +455,88 @@ async function convertStages(config) {
...
+        const setupJobExists = Object.keys(jobs).find(j => j === setupJobName || j === `~${setupJobName}`);
+        const teardownJobExists = Object.keys(jobs).find(j => j === teardownJobName || j === `~${teardownJobName}`);
+
+        if (setupJobExists) {
```
sagar1312: <strike>How do we handle the use case, when `setup` job exists but the steps has been updated in the config?
Same for `teardown`.
Please ignore if we are not allowing `setup`/`teardown` to be declared in the configuration in the initial version.</strike>

lib/eventFactory.js
```diff
@@ -453,50 +455,88 @@ async function convertStages(config) {
...
+            newStage.setup = [pipelineJobs.find(j => j.name === setupJobExists).id];
+        }
```
sagar1312: <strike>How do we handle the use when `setup` does not exist and needs to be created?

- `setup is explicitly specified in the `screwdriver.yaml` config 
- `setup is not specified in the `screwdriver.yaml` config

Same for `teardown`.</strike>

Assuming `setup` and `teardown` jobs would already be created/updated during Pipeline sync logic

lib/eventFactory.js
sagar1312: Same for `teardown`

lib/eventFactory.js
sagar1312: Same for `teardown`

lib/eventFactory.js
```diff
@@ -424,13 +426,14 @@ function createBuilds(config) {
...
- * @param  {Number}     config.groupEventId Group event ID
+ * @param  {Number}     config.eventId      Event ID
+ * @param  {Object}     config.jobs         Jobs
  * @param  {Object}     config.pipeline     Pipeline
  * @param  {Object}     config.stages       Pipeline stages
  * @return {Array}                New array with stages after up-converting
  */
 async function convertStages(config) {
-    const { pipeline, stages, groupEventId } = config;
+    const { pipeline, stages, eventId, jobs } = config;
```
sagar1312: `eventId` is not a valid field in stage model.
```suggestion
    const { pipeline, stages, jobs } = config;
```

lib/eventFactory.js
```diff
@@ -453,50 +455,88 @@ async function convertStages(config) {
             }
         });
 
+        newStage.eventId = eventId;
```
sagar1312:
```suggestion
        
```

lib/eventFactory.js
```diff
@@ -505,13 +545,42 @@ async function createStages({ pipeline, groupEventId }) {
         return Promise.resolve();
     }
 
-    const newStages = await convertStages({ pipeline, stages, groupEventId });
+    // list stage names from this pipeline that already exist
+    const existingStages = await stageFactory.list({ params: { pipelineId, name: stageNames } });
+    const existingStageNames = existingStages.map(stage => stage.name);
+    const convertedStages = await convertStages({ pipeline, stages, eventId, jobs, existingStages });
```
sagar1312: `existingStages` is not being used in the function `convertStages()`

lib/eventFactory.js
```diff
@@ -505,13 +545,42 @@ async function createStages({ pipeline, groupEventId }) {
...
+        }
+        if (stage.teardown) {
+            updateStageConfig.teardown = stage.teardown;
+        }
+        if (stage.archived) {
+            updateStageConfig.archived = stage.archived;
+        }
+
+        logger.info(`Updating stage:${JSON.stringify(stage)} for pipelineId:${pipelineId}.`);
+        processed.push(stageFactory.update(updateStageConfig));
```
sagar1312: `id` of existing stage has not been set for updating.

lib/eventFactory.js
```diff
@@ -505,13 +545,42 @@ async function createStages({ pipeline, groupEventId }) {
...
+        }
+        if (stage.setup) {
+            updateStageConfig.setup = stage.setup;
+        }
+        if (stage.teardown) {
+            updateStageConfig.teardown = stage.teardown;
+        }
+        if (stage.archived) {
+            updateStageConfig.archived = stage.archived;
+        }
```
sagar1312: Can this be simplified?
`archived` should be set to `false` (if the stage was previously removed and now added back in the screwdriver.yaml) 
```suggestion
        const updateStageConfig = {...stage, archived: false};
```

lib/eventFactory.js
```diff
@@ -505,13 +545,42 @@ async function createStages({ pipeline, groupEventId }) {
         return Promise.resolve();
     }
 
-    const newStages = await convertStages({ pipeline, stages, groupEventId });
+    // list stage names from this pipeline that already exist
+    const existingStages = await stageFactory.list({ params: { pipelineId, name: stageNames } });
+    const existingStageNames = existingStages.map(stage => stage.name);
+    const convertedStages = await convertStages({ pipeline, stages, eventId, jobs, existingStages });
+    const stagesToUpdate = convertedStages.filter(stage => existingStageNames.includes(stage.name));
```
sagar1312: We also need to update `archived=true` for stages that previously existed and have been removed now from the latest `screwdriver.yaml` config.

lib/eventFactory.js
```diff
@@ -453,50 +455,88 @@ async function convertStages(config) {
...
  * 2. Create the stages if it is defined and was not in the database yet
  * @method createStages
  * @param  {Object}     config              config
- * @param  {Number}     config.groupEventId Group event ID
+ * @param  {Number}     config.eventId      Event ID
  * @param  {Object}     config.pipeline     Pipeline
  * @return {Promise}
  */
-async function createStages({ pipeline, groupEventId }) {
+async function createStages({ pipeline, eventId }) {
```
sagar1312: Similar to rest of the pipeline configuration, shouldn't creating/updating stages be performed as part of pipeline sync flow?
https://github.com/screwdriver-cd/models/blob/master/lib/pipeline.js#LL864

Is there a reason why it is being deferred until an event is created?

lib/eventFactory.js
```diff
@@ -453,50 +455,88 @@ async function convertStages(config) {
...
+ * @param  {Object}     config              config
+ * @param  {Number}     config.eventId      Event ID
+ * @param  {Object}     config.pipeline     Pipeline
+ * @return {Promise}
+ */
+async function createStageBuilds({ stageFactory, stageBuildFactory, stageConfig }) {
+    const stageCreateConfig = stageConfig;
+    const stageBuildCreateConfig = {
+        eventId: stageConfig.eventId,
+        workflowGraph: stageConfig.workflowGraph,
```
sagar1312, line 496: `stageConfig.workflowGraph` would always be `undefined`. 

Looks like it is not being built using `stage.jobs`
https://github.com/screwdriver-cd/models/pull/570/files#diff-1c1c74750e13ea9a15f89f495aa0693768fb7ec332b4a01f9c83facdee0890deR476

lib/pipeline.js
```diff
@@ -1021,6 +1029,11 @@ class PipelineModel extends BaseModel {
                 // Handle internal nodes
                 if (job) {
                     node.id = job.id;
+                    // Add stageId
+                } else if (/^(?:~)?(stage@)/.test(node.name)) {
```
sagar1312, line 1033:
```suggestion
                } 
                if (/^(?:~)?(stage@)/.test(node.name)) {
```

sagar1312 (changes requested):

lib/pipeline.js
```diff
@@ -1021,6 +1029,11 @@ class PipelineModel extends BaseModel {
...
+                    const stage = await stageFactory.get({ params: { pipelineId: this.id, name: node.name } });
+
+                    node.stageId = stage.id;
```
sagar1312: This would throw an error, when stage does not exist in DB.
Creating/Updating stages should be done during Pipeline sync to be consistent with how we handle rest of the pipeline configuration.
Related to https://github.com/screwdriver-cd/models/pull/570#discussion_r1238622719

sagar1312: Also should make the node type explicit. Instead of implicitly identifying a node with `id` attribute as a job and a node with `stageId` as a stage. Something like
```
node.id = stage.id;
node.type = 'STAGE';
```

lib/eventFactory.js
sagar1312: Please ignore if we are not allowing `setup`/`teardown` to be declared in the configuration in the initial version.

sagar1312: I guess `setup` and `teardown` jobs would already be created/updated during Pipeline sync logic.
With that assumption, do we need this check?
Will there be a scenario where `setup` job does not exist for a stage?

lib/eventFactory.js
tkyi: removed

lib/eventFactory.js
tkyi: hmm should at least be an empty object

lib/eventFactory.js
tkyi: sure

lib/eventFactory.js
tkyi: can only update certain fields, so can't simplify it that way

lib/eventFactory.js
tkyi: added

lib/eventFactory.js
tkyi: thought it would be more straightforward since stageBuilds need to be created during event creation and we need to get the stage IDs and workflowGraphs to create stageBuilds
I can move stage creation to pipeline sync

lib/build.js
```diff
@@ -480,6 +480,57 @@ class BuildModel extends BaseModel {
         return secrets;
     }
 
+    /**
+     * Lazy load a stageBuild model for the build
+     * @property stageBuild
+     * @return {Promise}
+     */
+    async getStageBuild() {
```
sagar1312: Do we have any use case for this method?

lib/build.js
```diff
@@ -480,6 +480,57 @@ class BuildModel extends BaseModel {
...
+            throw new Error('Job does not exist');
+        }
+
+        const pipeline = await job.pipeline;
+
+        if (!pipeline) {
+            throw new Error('Pipeline does not exist');
+        }
+
+        const nodes = pipeline.workflowGraph.nodes;
```
sagar1312: `Job` could have moved to a different `Stage` or no longer associated to `Stage`.
We need to fetch and use `event` workflowGraph instead. This would give precise association of `build` to a `stage` when the event was created.

lib/build.js
```diff
@@ -480,6 +480,57 @@ class BuildModel extends BaseModel {
...
+
+        if (stageName) {
+            const stageFactory = StageFactory.getInstance();
+            const stageBuildFactory = StageBuildFactory.getInstance();
+
+            const stage = await stageFactory.get({ pipelineId: pipeline.id, name: stageName });
+
+            if (!stage || stage.length === 0) {
+                throw new Error('Stage does not exist');
+            }
```
sagar1312: Should the steps to fetch `stage` by name before fetching the `stageBuild` be abstracted into `StageBuildFatcory` if we are doing this on multiple occasions?

lib/helper.js
```diff
@@ -398,12 +399,23 @@ async function getBookendKey({ buildClusterName, annotations, pipeline, provider
...
 }
 
+/**
+ * Returns full stage name with correct formatting and setup or teardown suffix (e.g. stage@deploy:setup)
+ * @param  {String} stageName               Stage name
+ * @param  {String} type                    Type of stage job, either 'setup' or 'teardown'
+ * @return {String}                         Full stage name
+ */
+function getFullStageName({ stageName, type }) {
+    return `${STAGE_PREFIX}${stageName}:${type}`;
```
sagar1312:
```suggestion
function getFullStageJobName({ stageName, jobName }) {
    return `${STAGE_PREFIX}${stageName}:${jobName}`;
```

lib/event.js
```diff
@@ -15,8 +15,64 @@ class EventModel extends BaseModel {
...
+        const stageBuildFactory = StageBuildFactory.getInstance();
+
+        const stageBuilds = await stageBuildFactory.list({ params: { eventId: this.id, stageId: stageIds } });
+        const stageBuildsWithName = [];
+
+        stageBuilds.forEach(stageBuild => {
+            const stage = stages.find(s => s.id === stageBuild.stageId);
+
+            if (stage) {
+                stageBuild.stageName = stage.name;
```
sagar1312: Why do we need to add `stage` `name` to `stageBuild`?
Due to this the `stageBuild` schema is not consistent.
Ex: `stageBuilds` returned from the `stageBuildFactory` does not have `name` fields.

lib/event.js
```diff
@@ -15,8 +15,64 @@ class EventModel extends BaseModel {
...
+                stageNamesSet.add(n.stageName);
+            }
+        });
+
+        const stageNames = Array.from(stageNamesSet);
+        const stageListConfig = {
+            params: {
+                pipelineId: this.pipelineId,
+                name: stageNames,
+                archived: false
```
sagar1312: Entire logic to get the `stages` and merge some fields (`name`) into `stageBuilds` need not be done here.

If we decide to fetch `stages`, we should get the `stages` even if it is archived. If there is a reference to the archived stage in the workflow graph, means the stage was active when the event got created.
```suggestion
                name: stageNames
```

lib/pipeline.js
```diff
@@ -852,6 +853,128 @@ class PipelineModel extends BaseModel {
...
+                pipelineId,
+                ...value
+            };
+            const jobIds = [];
+
+            // Convert the jobNames to jobIds
+            value.jobs.forEach(jobName => {
+                const job = pipelineJobs.find(j => j.name === jobName);
+
+                if (job) {
```
sagar1312: Would there be a scenario, where a `stage` will have a `job` which do not exist?

lib/pipeline.js
```diff
@@ -852,6 +853,128 @@ class PipelineModel extends BaseModel {
...
+                    jobIds.push(job.id);
+                }
+            });
+
+            newStage.jobIds = jobIds;
```
sagar1312:
```suggestion
            newStage.jobIds = value.jobs.map(jobName => {
               return pipelineJobs.find(j => j.name === jobName).id;
            });
```

lib/pipeline.js
```diff
@@ -852,6 +853,128 @@ class PipelineModel extends BaseModel {
...
+            delete newStage.jobs; // extra field from yaml parser
+
+            // Check for setup and teardown
+            const setupJobName = getFullStageName({ stageName: key, type: 'setup' });
+            const teardownJobName = getFullStageName({ stageName: key, type: 'teardown' });
+            const setupJobExists = Object.keys(jobs).find(j => j === setupJobName || j === `~${setupJobName}`);
+            const teardownJobExists = Object.keys(jobs).find(j => j === teardownJobName || j === `~${teardownJobName}`);
+
+            newStage.setup = pipelineJobs.find(j => j.name === setupJobExists).id;
+            newStage.teardown = pipelineJobs.find(j => j.name === teardownJobExists).id;
```
sagar1312: `setup` and `teardown` jobs must always exist.
Can this be simplified to?
```suggestion
            newStage.setup = pipelineJobs.find(j => j.name === setupJobName).id;
            newStage.teardown = pipelineJobs.find(j => j.name === teardownJobName).id;
```

lib/pipeline.js
```diff
@@ -852,6 +853,128 @@ class PipelineModel extends BaseModel {
...
+     *  - [{ "name": "name", "jobs": [1, 2, 3], "pipelineId": 123, "description": "value" },
+     *     { "name": "name2", "jobs": [4, 5], "pipelineId": 123 }]
+     * @method convertStages
+     * @param  {Object}     config              config
+     * @param  {Object}     config.jobs         Jobs
+     * @param  {Object}     config.pipeline     Pipeline
+     * @param  {Object}     config.stages       Pipeline stages
+     * @return {Array}                New array with stages after up-converting
+     */
+    async _convertStages({ pipelineId, stages, jobs, pipelineJobs }) {
```
sagar1312: What is the difference bw `jobs` and `pipelineJobs`?
We can remove `jobs` if it is redundant.
```suggestion
     * @param  {Object}     config.pipelineJobs         Jobs
     * @param  {Object}     config.pipeline                 Pipeline
     * @param  {Object}     config.stages                   Pipeline stages
     * @return {Array}                New array with stages after up-converting
     */
    async _convertStages({ pipelineId, stages, pipelineJobs }) {
```

lib/pipeline.js
```diff
@@ -852,6 +853,128 @@ class PipelineModel extends BaseModel {
...
+        // Get new stages
+        const stages = parsedConfig.stages || {};
+        const jobs = parsedConfig.jobs || {};
+
+        // No stages in yaml, skip creation
+        if (Object.keys(stages).length === 0) {
+            logger.info(`No stages for pipelineId:${pipelineId}, skipping stage creation.`);
+
+            return Promise.resolve();
+        }
```
sagar1312: If no stages in yaml, we still need to proceed to update (archive) existing stages if any.
```suggestion
```

lib/pipeline.js
```diff
@@ -852,6 +853,128 @@ class PipelineModel extends BaseModel {
...
+
+        // list stage names from this pipeline that already exist
+        const existingStages = await stageFactory.list({ params: { pipelineId } });
+        const existingStageNames = existingStages.map(stage => stage.name);
+        // Format new stage data
+        const convertedStages = await this._convertStages({ pipelineId, stages, jobs, pipelineJobs });
```
sagar1312:
```suggestion
        const convertedStages = await this._convertStages({ pipelineId, stages, pipelineJobs });
```

lib/stageBuild.js
```diff
@@ -0,0 +1,20 @@
...
+
+class StageBuildModel extends BaseModel {
+    /**
+     * Construct a StageBuildModel object
+     * @method constructor
+     * @param  {Object}    config
+     * @param  {Object}    config.datastore     Object that will perform operations on the datastore
+     * @param  {Number}    config.eventId       Event ID
+     * @param  {Number}    config.stageId       Stage ID
+     * @param  {Object}    config.workflowGraph Stage workflowGraph
```
sagar1312:
```suggestion
     * @param  {String}    config.status Stage build status
```

lib/stage.js
```diff
@@ -9,9 +9,12 @@ class StageModel extends BaseModel {
      * @param  {Object}    config
      * @param  {Object}    config.datastore     Object that will perform operations on the datastore
      * @param  {String}    [config.description] Stage description
-     * @param  {Array}     [config.jobIds=[]]   Job Ids that belong to this stage
+     * @param  {Array}     [config.jobIds=[]]   Job IDs that belong to this stage
      * @param  {String}    config.name          Name of the stage
      * @param  {Number}    config.pipelineId    Pipeline the stage belongs to
+     * @param  {Array}     [config.setup]       Setup job IDs
+     * @param  {String}    [config.startFrom]   Stage start point (a job name, e.g. 'main')
```
sagar1312:
```suggestion
```

lib/stageFactory.js
```diff
@@ -27,14 +27,19 @@ class StageFactory extends BaseFactory {
...
-     * @param {Array}       [config.jobIds=[]]   Job Ids that belong to this stage
-     * @param {String}      config.name          Name of the stage
-     * @param {String}      config.pipelineId    Pipeline the stage belongs to
+     * @param {Object}  config
+     * @param {String}  [config.description] Stage description
+     * @param {Array}   [config.jobIds=[]]   Job IDs that belong to this stage
+     * @param {String}  config.name          Name of the stage
+     * @param {Number}  config.pipelineId    Pipeline the stage belongs to
+     * @param {Array}   [config.setup]       Setup job IDs
+     * @param {String}  [config.startFrom]   Stage start point (a job name, e.g. 'main')
```
sagar1312:
```suggestion
```

sagar1312 (changes requested):

lib/build.js
tkyi: removed

lib/helper.js
tkyi: updated

lib/event.js
tkyi: removed

lib/event.js
tkyi: removed

lib/pipeline.js
tkyi: if it's an archived stage, maybe?

lib/pipeline.js

lib/pipeline.js

lib/pipeline.js
tkyi: removed

lib/pipeline.js
tkyi: removed

lib/pipeline.js

lib/stageBuild.js
tkyi: updated

lib/stage.js
tkyi: updated

lib/stageFactory.js
tkyi: updated

lib/event.js
```diff
@@ -15,8 +15,51 @@ class EventModel extends BaseModel {
...
+        const stageIds = stages.map(s => s.id);
+
+        // Lazy load factory dependency to prevent circular dependency issues
+        // https://nodejs.org/api/modules.html#modules_cycles
+        /* eslint-disable global-require */
+        const StageBuildFactory = require('./stageBuildFactory');
+        /* eslint-enable global-require */
+        const stageBuildFactory = StageBuildFactory.getInstance();
+
+        return stageBuildFactory.list({ params: { eventId: this.id, stageId: stageIds } });
```
sagar1312: Filtering by `stageId` may not influence the result. Filtering only by `eventId` should suffice. 
```suggestion
        /* eslint-enable global-require */
        const stageBuildFactory = StageBuildFactory.getInstance();

        return stageBuildFactory.list({ params: { eventId: this.id } });
```

sagar1312 (changes requested):

lib/event.js
tkyi: updated

sagar1312 (approved):

PR #581, opened 2023-07-07
fix(2878): return meta for metrics

klu909:
## Context

returning meta for metrics to displaying in ui
## License


tkyi (approved):
Can you update description? otherwise lgtm

PR #580, opened 2023-06-05
feat: add build cluster scm org check for native cluster

pritamstyz4ever:
## Context

Access native build cluster with provider needs to be restricted by SCM organization

## Objective

This PR updates the build cluster fetching logic for provider configuration and restricts if not from same organization

## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #582, opened 2023-07-27
fix: always check scmOrganizations unless it's empty

DekusDenial:
always check the pipeline scm org with build cluster's scmOrganizations as long as it's not empty


tkyi (approved):

tkyi:
Related to https://github.com/screwdriver-cd/models/pull/580

adong (approved):

PR #594, opened 2023-10-27
fix(2767): Fix issues while parsing and creating/updating stage entities

sagar1312:
## Context

Noticed below issues during pipeline creation (and pipeline sync) involving stages

1. `id` of setup/teardown jobs are not available while creating a stage.
2. 'id' of setup/teardown jobs are set as an array in stage. This results DB error as it expects the value to be an integer.
3. Updating exiting stages invoked `update` method of stage factory  which does not exist. Update should always be performed by invoking `update` method of `stage` object retrieved from DB.
4. When a stage was removed from the pipeline configuration, it is not marked as `archived`. Only the stages present in the latest pipeline configuration were fetched from DB. Hence the obsolete stages were never fetched and archived.

## Objective
Fixed all the above mentioned issues.

## References
https://github.com/screwdriver-cd/screwdriver/issues/2767

## License

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -901,34 +901,13 @@ class PipelineModel extends BaseModel {
             const setupJobExists = Object.keys(jobs).find(j => j === setupJobName || j === `~${setupJobName}`);
             const teardownJobExists = Object.keys(jobs).find(j => j === teardownJobName || j === `~${teardownJobName}`);
 
-            newStage.setup = [pipelineJobs.find(j => j.name === setupJobExists).id];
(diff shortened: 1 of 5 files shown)


sagar1312:
Will mark as ready for review after https://github.com/screwdriver-cd/data-schema/pull/539 and https://github.com/screwdriver-cd/config-parser/pull/146 are merged

tkyi (commented):
You don't need to update any tests?

sagar1312:
> You don't need to update any tests?

Done

tkyi (approved):

PR #595, opened 2023-11-02
fix(2767): Pipeline workflow graph should only contain job nodes

sagar1312:
## Context

Unified workflow graph would always contain nodes representing jobs. Stage will never be represented as nodes in the unified workflow graph.

## Objective

Remove the logic that add nodes for stages

## References

https://github.com/screwdriver-cd/screwdriver/issues/2767

## License

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -913,13 +913,15 @@ class PipelineModel extends BaseModel {
     /**
      * Sync stages
      * 1. Convert new stages into correct format, prepopulate with jobIds
-     * 2. Create the stages if it is defined and was not in the database yet
-     * @method createStages
+     * 2.a. Create the stages if it is defined and was not in the database yet
+     * 2.b. Update the existing stages with the new configuration
+     * 2.c. Archive then existing stages if they no longer exist in the configuration
+     * @method _createOrUpdateStages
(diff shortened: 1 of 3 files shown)


lib/pipeline.js
```diff
@@ -913,13 +913,15 @@ class PipelineModel extends BaseModel {
     /**
      * Sync stages
      * 1. Convert new stages into correct format, prepopulate with jobIds
-     * 2. Create the stages if it is defined and was not in the database yet
-     * @method createStages
+     * 2.a. Create the stages if it is defined and was not in the database yet
+     * 2.b. Update the existing stages with the new configuration
+     * 2.c. Archive then existing stages if they no longer exist in the configuration
```
tkyi, line 918:
```suggestion
     * 2.c. Archive the existing stages if they no longer exist in the configuration
```

tkyi:
```suggestion
     * 2.a. Create stages if they are defined and were not already in the database
     * 2.b. Update existing stages with the new configuration
     * 2.c. Archive existing stages if they no longer exist in the configuration
```

sagar1312: Done

jithine (approved):

PR #597, opened 2023-11-20
feat: add list pipeline template operations

pritamstyz4ever:
## Context

Add operation list for `pipelineTemplate` meta and `pipelineTemplateVersions`

## Objective

Below changes have been made
- added list() function to `PipelineTemplateFactory`
- added list() function to `PipelineTemplateVersionsFactory`

## References

https://github.com/screwdriver-cd/screwdriver/issues/2135

## License


lib/templateMetaFactory.js
```diff
@@ -23,18 +23,49 @@ class TemplateMetaFactory extends BaseFactory {
...
 
+    /**
+     * Create pipeline template meta
+     * @method list
+     * @param  {Object}   config                  Config object for template meta
+     * @param  {Object}   config.pipelineId       Identifier of the pipeline that is publishing the template
+     * @return {Promise}                          Resolve pipeline template meta
+     */
     create(config) {
         config.templateType = this._getTemplateType();
```
tkyi:
```suggestion
        config.templateType = this.getTemplateType();
```

lib/pipelineTemplateVersionFactory.js
```diff
@@ -130,6 +130,46 @@ class PipelineTemplateVersionFactory extends BaseFactory {
...
+            params.namespace = namespace;
+        }
+
+        const pipelineTemplateMeta = await templateMetaFactory.get(params);
+
+        if (!pipelineTemplateMeta) {
+            return [];
+        }
+
+        templateId = pipelineTemplateMeta.id;
```
tkyi: I don't love that we're potentially overwriting a param

lib/templateMetaFactory.js
```diff
@@ -23,18 +23,49 @@ class TemplateMetaFactory extends BaseFactory {
...
+     * List all pipeline template meta
+     * @method list
+     * @param  {Object}   config                  Config object for template meta
+     * @param  {Object}   config.params           Parameters to filter on
+     * @return {Promise}                          Resolve pipeline template meta list
+     */
+    list(config) {
+        config.params.templateType = this.getTemplateType();
+
+        return super.list(config);
```
sagar1312, line 71: We should do something similar what is being done in `[TemplateTagFactory](https://github.com/screwdriver-cd/models/blob/master/lib/templateTagFactory.js#L149)`
```suggestion
        if (!config.params) {
            config.params = {};
        }
        config.params.templateType = this.getTemplateType();

        return super.list(config);
```

lib/pipelineTemplateFactory.js
```diff
@@ -14,6 +14,21 @@ class PipelineTemplateFactory extends TemplateMetaFactory {
...
+     * @return {Promise}                          Resolve pipeline template meta
+     */
+    async list(config) {
+        if (config.params) {
+            config.params.templateType = this._getTemplateType();
+        }
+
+        return super.list(config);
+    }
+
```
sagar1312: This is not needed. This should be abstracted into parent class `TemplateMetaFactory`
```suggestion
```

lib/pipelineTemplateVersionFactory.js
sagar1312: Second that. We should make `name + namespace` and `templateId` mutually exclusive.

lib/pipelineTemplateVersionFactory.js
```diff
@@ -130,6 +130,46 @@ class PipelineTemplateVersionFactory extends BaseFactory {
...
+
+        return super.list({
```
sagar1312: We may also have to pass other configs Ex: `config.paginate`

sagar1312 (changes requested):

lib/pipelineTemplateVersionFactory.js
pritamstyz4ever: I agree, let me fix that

lib/pipelineTemplateVersionFactory.js
pritamstyz4ever: ok, so we do intend to support pagination and sorting? It's not in the current design doc.

tkyi (approved):

lib/pipelineTemplateVersionFactory.js
sagar1312: It is not called out explicitly. The list api should allow filtering and sorting.

lib/pipelineTemplateVersionFactory.js
```diff
@@ -130,6 +130,47 @@ class PipelineTemplateVersionFactory extends BaseFactory {
...
 
+    /**
+     * List all the versions of a pipeline template
+     * @method create
+     * @param  {Object}     config               Config object
+     * @param templateMetaFactory
+     * @param  {Datastore}  config.datastore     Object that will perform operations on the datastore
+     * @param  {String}     config.name          The template name
+     * @param  {String}     config.namespace     The template namespace
+     * @param  {String}     config.templateId    Identifier of the template
```
sagar1312:
```suggestion
     * @param  {String}     config.params.templateId    Identifier of the template
```

lib/pipelineTemplateVersionFactory.js
```diff
@@ -130,6 +130,47 @@ class PipelineTemplateVersionFactory extends BaseFactory {
...
+            return [];
+        }
+
+        return super.list({
+            params: {
+                templateId: pipelineTemplateMeta.id
+            },
+            paginate: config.paginate
+        });
+    }
```
sagar1312: The list API should be generic to allow filtering on any field. Accept filter criteria `config.params` and pass it with any modifications to the query.

```suggestion
        return super.list(config);
    }
```

lib/pipelineTemplateVersionFactory.js
```diff
@@ -130,6 +130,47 @@ class PipelineTemplateVersionFactory extends BaseFactory {
...
+        if (name && namespace) {
+            params.name = name;
+            params.namespace = namespace;
+        }
+
+        const pipelineTemplateMeta = await templateMetaFactory.get(params);
+
+        if (!pipelineTemplateMeta) {
+            return [];
+        }
```
sagar1312: We could skip getting templateMeta when `name` and `namespace` are not specified.

```suggestion
        if (name && namespace) {
            const params = {};
        
            params.name = name;
            params.namespace = namespace;
            
            if (config.params.templateId) {
                params.id = config.params.templateId;
            }
            
            const pipelineTemplateMeta = await templateMetaFactory.get(params);
            
            if(pipelineTemplateMeta) {
               config.params.templateId = pipelineTemplateMeta.id;
            }
            else {
              return [];
            }
        }
```

sagar1312 (changes requested):

lib/pipelineTemplateVersionFactory.js
pritamstyz4ever: removing templateId, as per docs the API is to list the versions by name and namespace.

lib/pipelineTemplateVersionFactory.js
```diff
@@ -130,6 +130,47 @@ class PipelineTemplateVersionFactory extends BaseFactory {
...
+     * @param  {Object}     config.paginate             Pagination parameters
+     * @param  {Number}     config.paginate.count       Number of items per page
+     * @param  {Number}     config.paginate.page        Specific page of the set to return
+     * @return {Promise}                         List of PipelineTemplateVersionModel
+     */
+    async list(config, templateMetaFactory) {
+        const { name, namespace } = config.params;
+
+        if (!name || !namespace) {
+            return [];
```
sagar1312: Returning empty list when only `name` or `namespace` is specified could be misleading
We should throw error indicating both `name` and `namespace` is mandatory.

lib/pipelineTemplateVersionFactory.js
```diff
@@ -130,6 +130,47 @@ class PipelineTemplateVersionFactory extends BaseFactory {
...
+        }
+
+        return super.list({
+            params: {
+                templateId: pipelineTemplateMeta.id
+            },
+            paginate: config.paginate,
+            sort: config.sort
+        });
```
sagar1312: Plucking specific config makes the API rigid and not extensible.
List API should support [all the configs](https://github.com/screwdriver-cd/models/blob/master/lib/baseFactory.js#L110) supported by the base class implementation.
```suggestion
        if(!config.params) {
          config.params = {};
        }
        
        config.params.templateId = pipelineTemplateMeta.id;
        
        return super.list(config);
```

sagar1312 (dismissed):

jithine (approved):

PR #598, opened 2023-12-01
feat(2465): Prevent event's baseBranch from being null

itleigns:
## Context
There are instances where the `baseBranch` of the DB is `null` in the event table. In such cases, the branch that the event refers to does not get displayed in the UI.

## Objective
I made changes so that the `baseBranch` is also set for events created through remote triggers and the UI.

## References
https://github.com/screwdriver-cd/screwdriver/issues/2465#issuecomment-1276827077

## License

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -709,8 +709,7 @@ class EventFactory extends BaseFactory {
             creator: config.creator || null,
             meta: config.meta || {},
             pr: {},
-            prNum,
-            baseBranch: config.baseBranch || null
+            prNum
         };
         let prevChainPR = '';
         let decoratedCommit;
@@ -770,6 +769,16 @@ class EventFactory extends BaseFactory {
                         if (prInfo.prBranchName) {
                             modelConfig.pr.prBranchName = prInfo.prBranchName;
                         }
+                    }
+
+                    if (config.baseBranch) {
+                        modelConfig.baseBranch = config.baseBranch;
+                    }
+                    else if(prInfo && prInfo.baseBranch) {
+                        modelConfig.baseBranch = prInfo.baseBranch;
+                    }
+                    else if(p.scmRepo && p.scmRepo.branch) {
+                        modelConfig.baseBranch = p.scmRepo.branch;
                     }
 
                     if (prTitle) {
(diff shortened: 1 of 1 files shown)


lib/eventFactory.js
```diff
@@ -772,6 +771,14 @@ class EventFactory extends BaseFactory {
                         }
                     }
 
+                    if (config.baseBranch) {
+                        modelConfig.baseBranch = config.baseBranch;
+                    } else if (prInfo && prInfo.baseBranch) {
+                        modelConfig.baseBranch = prInfo.baseBranch;
+                    } else if (p.scmRepo && p.scmRepo.branch) {
+                        modelConfig.baseBranch = p.scmRepo.branch;
+                    }
```
itleigns, line 783: When triggered by a webhook, or when there is a parentEvent such as a restart case, the config value is set in the first scenario. When a PR event is begun from the UI, the PR's target is set in the second scenario. For regular events initiated through the UI or triggered by remote pipeline, the pipeline's branch is set in the third scenario.

jithine: perhaps leave this as a comment around the code block for documentation ?

lib/eventFactory.js
```diff
@@ -772,6 +771,14 @@ class EventFactory extends BaseFactory {
                         }
                     }
 
+                    if (config.baseBranch) {
+                        modelConfig.baseBranch = config.baseBranch;
+                    } else if (prInfo && prInfo.baseBranch) {
+                        modelConfig.baseBranch = prInfo.baseBranch;
+                    } else if (p.scmRepo && p.scmRepo.branch) {
```
y-oksaku: I think using `optional chaining` and `nullish coalescing operator` would be preferable.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing#relationship_with_the_optional_chaining_operator

itleigns: Sorry, it is hard to write that way when trying to add comments naturally.

y-oksaku: OK, but you can use optional chaining within if conditions. What about this?

itleigns: I got an error in lint.
We need to version up to es2020.
I leave it as is for now.
https://github.com/screwdriver-cd/eslint-config-screwdriver/blob/15860e11d9a6567fdece92ab6f4928c8cb9c9db2/index.js#L10

y-oksaku: Noted with thanks! 👍

y-oksaku (approved):

lib/eventFactory.js
itleigns: OK! I added the comment.

ibu1224 (approved):

tkyi (approved):

PR #600, opened 2024-01-17
fix(2767): Create stageBuild during build create

tkyi:
## Context

We should create stageBuild during build create to avoid duplication in multiple places.

## Objective

This PR checks if build to be created is stage setup build, then create stageBuild if it is.

## References

Related to https://github.com/screwdriver-cd/screwdriver/pull/2870#discussion_r1448110743, https://github.com/screwdriver-cd/screwdriver/pull/2870

## License

### DIFF
--- a/lib/buildFactory.js
+++ b/lib/buildFactory.js
@@ -226,12 +226,12 @@ class BuildFactory extends BaseFactory {
                     });
 
                     // Create stage build if current job is stage setup
-                    const nextJobIsStageSetup = job.name.match(STAGE_SETUP_PATTERN);
+                    const stageSetupJob = job.name.match(STAGE_SETUP_PATTERN);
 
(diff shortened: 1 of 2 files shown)


lib/buildFactory.js
```diff
@@ -219,6 +225,23 @@ class BuildFactory extends BaseFactory {
                         provider
                     });
 
+                    // Create stage build if current job is stage setup
+                    const nextJobIsStageSetup = job.name.match(STAGE_SETUP_PATTERN);
```
sagar1312:
```suggestion
                    const jobIsStageSetup = job.name.match(STAGE_SETUP_PATTERN);
```

sagar1312 (changes requested):

sagar1312 (approved):

PR #607, opened 2024-02-08
fix: use template id to list pipeline template versions

klu909:
## Context

attempting to list pipeline template version would throw `Error: Invalid param "namespace"`

## Objective
we only need template id to list pipeline template, name and namespace are not necessary 

## License

### DIFF
--- a/lib/pipelineTemplateVersionFactory.js
+++ b/lib/pipelineTemplateVersionFactory.js
@@ -162,7 +162,11 @@ class PipelineTemplateVersionFactory extends BaseFactory {
             return [];
         }
 
-        config.params.templateId = pipelineTemplateMeta.id;
+        config = {
+            params: {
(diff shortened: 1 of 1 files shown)


VonnyJap (approved):

lib/pipelineTemplateVersionFactory.js
```diff
@@ -162,9 +162,11 @@ class PipelineTemplateVersionFactory extends BaseFactory {
...
 
-        config.params.templateId = pipelineTemplateMeta.id;
-
-        return super.list(config);
+        return super.list({
+            params: {
+                templateId: pipelineTemplateMeta.id
+            }
+        });
     }
```
pritamstyz4ever, line 170: revert this change, we need to pass config as it can contain, sort and paginate

pritamstyz4ever (commented):
Revert this change

PR #608, opened 2024-02-08
fix: refactor getWithMetadata in pipelineTemplateVersionFactory

klu909:
## Context

getWithMetadata was used in https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/pipelines/templates/getVersion.js

but it would throw error because parameter `versionOrTag` was not handled
## Objective

refactored getWithMetadata, removed `templatedId` parameter because it was not being used anywhere

## License

### DIFF
--- a/lib/pipelineTemplateVersionFactory.js
+++ b/lib/pipelineTemplateVersionFactory.js
@@ -5,6 +5,7 @@ const BaseFactory = require('./baseFactory');
 const PipelineTemplateVersionModel = require('./pipelineTemplateVersion');
 
 const EXACT_VERSION_REGEX = schema.config.regex.EXACT_VERSION;
+const TEMPLATE_TAG_NAME_REGEX = schema.config.regex.TEMPLATE_TAG_NAME;
 const VERSION_REGEX = schema.config.regex.VERSION;
 let instance;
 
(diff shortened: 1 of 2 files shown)


lib/pipelineTemplateVersionFactory.js
```diff
@@ -203,38 +204,48 @@ class PipelineTemplateVersionFactory extends BaseFactory {
...
+     * @param  {String}     config.name           The template name
+     * @param  {String}     config.namespace      The template namespace
+     * @param  {String}     [config.version]      The template version
+     * @param  {String}     [config.versionOrTag] The template version or tag
+     * @param  {Factory}    templateMetaFactory   Template meta factory
+     * @param  {Factory}    [templateTagFactory]  Template tag factory
+     * @return {Promise}                          {PipelineTemplateVersionModel, PipelineTemplateMetaModel}
      */
-    async getWithMetadata(config, templateMetaFactory) {
+    async getWithMetadata(config, templateMetaFactory, templateTagFactory) {
```
sagar1312, line 216: We should deprecate and remove the usage of generic `TemplateTagFactory`. Instead use concrete implementation.
```suggestion
    async getWithMetadata(config, templateMetaFactory, pipelineTemplateTagFactory) {
```

lib/pipelineTemplateVersionFactory.js
```diff
@@ -203,38 +204,48 @@ class PipelineTemplateVersionFactory extends BaseFactory {
...
-        } else {
-            pipelineTemplateMeta = await templateMetaFactory.get({
+        pipelineTemplateMeta = await templateMetaFactory.get({
+            name: config.name,
+            namespace: config.namespace
+        });
+
+        if (!pipelineTemplateMeta) {
+            return null;
+        }
```
sagar1312, line 227: Factory should not be designed for a specific (current) need.
We should allow accepting `templateId`.

```suggestion
        const {templateId, name, namespace, versionOrTag, ...versionGetConfig} = config;
        const metaGetConfig = {};
        if(name || namespace) {
            if(name && namespace) {
              metaGetConfig.id = name;
              metaGetConfig.id = namespace;
            }
            else {
               return null; // or throw error
            }
        }
        
        if(templateId) {
            metaGetConfig.id = templateId;
        }
        
        pipelineTemplateMeta = await templateMetaFactory.get(metaGetConfig);

        if (!pipelineTemplateMeta) {
            return null;
        }
        
        versionGetConfig.templateId = pipelineTemplateMeta.id;
```

lib/pipelineTemplateVersionFactory.js
```diff
@@ -203,38 +204,48 @@ class PipelineTemplateVersionFactory extends BaseFactory {
...
-     * @param  {Datastore}  config.datastore     Object that will perform operations on the datastore
-     * @param  {String}     [config.templateId]  The template id
-     * @param  {String}     config.name          The template name
-     * @param  {String}     config.namespace     The template namespace
-     * @param  {String}     config.version       The template version
-     * @param  {Factory}   templateMetaFactory  Template meta factory
-     * @return {Promise}                         {PipelineTemplateVersionModel, PipelineTemplateMetaModel}
+     * @param  {Object}     config                Config object
+     * @param  {String}     config.name           The template name
+     * @param  {String}     config.namespace      The template namespace
```
sagar1312, line 209: Factory should not be designed for a specific need.
We should allow accepting `templateId`.

lib/pipelineTemplateVersionFactory.js
```diff
@@ -203,38 +204,48 @@ class PipelineTemplateVersionFactory extends BaseFactory {
...
-        if (!pipelineTemplateMeta) {
-            return null;
+            templateVersion = templateTag.version;
         }
 
-        const pipelineTemplateVersion = await super.get(config);
+        const pipelineTemplateVersion = await super.get({
+            templateId: pipelineTemplateMeta.id,
+            version: templateVersion
+        });
```
sagar1312, line 248: We should not cherry pick the params

```suggestion
        const pipelineTemplateVersion = await super.get(versionGetConfig);
```

lib/pipelineTemplateVersionFactory.js
```diff
@@ -203,38 +204,48 @@ class PipelineTemplateVersionFactory extends BaseFactory {
...
             }
-        }
 
-        if (!pipelineTemplateMeta) {
-            return null;
+            templateVersion = templateTag.version;
         }
 
-        const pipelineTemplateVersion = await super.get(config);
+        const pipelineTemplateVersion = await super.get({
```
sagar1312, line 245: Looks like `versionOrTag` overrides `version`  which doesn't seem right(when both `version` and `versionOrTag`).
We should handle the scenario when both `version` and `versionOrTag` is specified.
```suggestion
       if(versionGetConfig.version !== templateVersion) {
          return null; // or throw error
       }
       
       if(templateVersion) {
           versionGetConfig.version = templateVersion;
       }

        const pipelineTemplateVersion = await super.get({
```

sagar1312 (changes requested):

pritamstyz4ever (changes requested):
I don't think these changes need to be here, we should revert it.
I will create a PR for the versionorTag logic

lib/pipelineTemplateVersionFactory.js
klu909: Providing a tag is same as providing a version, since tag is pinned to a specific version. So tag will override version if they are both provided, because doesn't make sense to begin with.

lib/pipelineTemplateVersionFactory.js
klu909: pipelineTemplateMeta.id is the source of truth for template id no? can you elaborate a bit more?

PR #609, opened 2024-02-08
fix: properly list version with sort and pagination

klu909:
## Context

last pr did not account for pagination or sort
## Objective

now properly pass in pagination and sort for listing versions
## License

### DIFF
--- a/lib/pipelineTemplateVersionFactory.js
+++ b/lib/pipelineTemplateVersionFactory.js
@@ -139,6 +139,7 @@ class PipelineTemplateVersionFactory extends BaseFactory {
      * @param  {String}     config.params.name          The template name
      * @param  {String}     config.params.namespace     The template namespace
      * @param  {String}     config.sort                 The sort order of the list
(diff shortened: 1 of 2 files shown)


lib/pipelineTemplateVersionFactory.js
```diff
@@ -162,11 +163,25 @@ class PipelineTemplateVersionFactory extends BaseFactory {
...
+
+        if (config.sortBy) {
+            listConfig.sortBy = config.sortBy;
+        }
+
+        if (config.paginate) {
+            listConfig.paginate = config.paginate;
+        }
+
+        return super.list(listConfig);
```
pritamstyz4ever, line 184: we don't need to do this, we can pass the config directly

pritamstyz4ever (changes requested):

lib/pipelineTemplateVersionFactory.js
klu909: we can't pass in config directly because it does not accept name & namespace

PR #611, opened 2024-02-09
fix: list pipeline template versions

pritamstyz4ever:
## Context

fix the issue with listing pipeline template versions

## Objective

fix the issue with listing pipeline template versions

## License

### DIFF
--- a/lib/pipelineTemplateVersionFactory.js
+++ b/lib/pipelineTemplateVersionFactory.js
@@ -136,8 +136,8 @@ class PipelineTemplateVersionFactory extends BaseFactory {
      * @param  {Object}     config               Config object
      * @param templateMetaFactory
      * @param  {Datastore}  config.datastore     Object that will perform operations on the datastore
-     * @param  {String}     config.params.name          The template name
-     * @param  {String}     config.params.namespace     The template namespace
+     * @param  {String}     config.name          The template name
+     * @param  {String}     config.namespace     The template namespace
      * @param  {String}     [config.params.templateId]         The template id
      * @param  {String}     config.sort                 The sort order of the list
      * @param  {Object}     config.paginate             Pagination parameters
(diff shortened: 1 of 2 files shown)


minghay (approved):

sagar1312:
What was the issue?

lib/pipelineTemplateVersionFactory.js
```diff
@@ -162,9 +162,12 @@ class PipelineTemplateVersionFactory extends BaseFactory {
...
         }
 
-        config.params.templateId = pipelineTemplateMeta.id;
-
-        return super.list(config);
+        return super.list({
+            ...config,
+            params: {
+                templateId: pipelineTemplateMeta.id
+            }
```
sagar1312: We should not override the params with just `templateId`. There could be other params (Ex: `createTime`) besides `name` and `namesapce`.
Instead, exclude `name` and `namespace` from the params.

sagar1312 (changes requested):

tkyi (approved):

lib/pipelineTemplateVersionFactory.js
pritamstyz4ever: not really, `pipelineTemplateVersion` schema does not have name,namespace, only way to list is via templateId

sagar1312: I agree that we don't need to exclude `name` and `namespace` as these get ignored while constructing DB query. But, it is still unclear on why we are picking only `templateId` from params.
Model interface should not be designed/restricted due to current use cases in the Screwdriver API. It should allow filtering on any of the `pipelineTemplateVersion` field.

pritamstyz4ever: the params don't get ignored I think, so we need to remove them.
https://github.com/screwdriver-cd/models/blob/master/lib/baseFactory.js#L132

sagar1312: Yes, looks like if there are params not specified in the schema, error is thrown
https://github.com/screwdriver-cd/datastore-sequelize/blob/master/index.js#L475

Then, as originally suggested, let us exclude `name` and `namespace` from `config.params`.

Or, if we want to be strict and reserve `config.params` to only allow the fields of `pipelineTemplateVersion`, then should we have `name` and `namespace` on `config` and not in `config.params`?

pritamstyz4ever: Yes, I think we can move `name` and `namespace` from `config.params`

lib/pipelineTemplateVersionFactory.js
```diff
@@ -149,19 +150,22 @@ class PipelineTemplateVersionFactory extends BaseFactory {
             config.params = {};
         }
 
-        if (!config.params.name || !config.params.namespace) {
+        if (config.params.templateId) {
+            return super.list(config);
+        }
```
sagar1312, line 155: When `templateId` params exist, why are we not considering `name` and `namespace`?
When `templateId`, `name` and `namespace` are specified, we should return versions associated with the template matching `templateId`, `name` and `namespace`.

sagar1312 (changes requested):

sagar1312 (approved):

PR #612, opened 2024-02-09
fix(3024): template usage and metrics reporting to exclude archived jobs

VonnyJap:
## Context
The current implementation of tracking the job template usage and metrics include the archived jobs which leads to wrong reporting. 

## Objective
Exclude archived jobs from the usage and metrics implementation.

## References

https://github.com/screwdriver-cd/screwdriver/issues/3024

## License

### DIFF
--- a/lib/rawQueries.js
+++ b/lib/rawQueries.js
@@ -3,17 +3,17 @@
 const Queries = {
     getPipelineUsageCountForJobTemplatesQuery: (
         tablePrefix = ''
-    ) => `SELECT COUNT(DISTINCT "pipelineId") as count, "templateId"
+    ) => `SELECT COUNT(DISTINCT "pipelineId") as count, "templateId", "id"
         FROM "${tablePrefix}jobs"
-        WHERE "templateId" IN (:templateIds)
-        GROUP BY "templateId"`,
+        WHERE "templateId" IN (:templateIds) AND "id" IN (:jobIds)
+        GROUP BY "templateId", "id"`,
 
     getPipelineUsageCountForJobTemplatesQueryMySql: (
         tablePrefix = ''
-    ) => `SELECT COUNT(DISTINCT pipelineId) as count, templateId
+    ) => `SELECT COUNT(DISTINCT pipelineId) as count, templateId, id
         FROM \`${tablePrefix}jobs\`
-        WHERE templateId IN (:templateIds)
-        GROUP BY templateId`,
+        WHERE templateId IN (:templateIds) AND id IN (:jobIds)
+        GROUP BY templateId, id`,
     // getBuildStatuses()
     getStatusesQuery: (tablePrefix = '') => `SELECT "id", "jobId", "status", "startTime", "endTime", "meta"
         FROM (SELECT "id", "jobId", "status", "startTime", "endTime", "meta",
--- a/lib/rawQueries.js
+++ b/lib/rawQueries.js
@@ -3,17 +3,17 @@
 const Queries = {
     getPipelineUsageCountForJobTemplatesQuery: (
         tablePrefix = ''
-    ) => `SELECT COUNT(DISTINCT "pipelineId") as count, "templateId", "id"
+    ) => `SELECT COUNT(DISTINCT "pipelineId") as count, "templateId"
         FROM "${tablePrefix}jobs"
-        WHERE "templateId" IN (:templateIds) AND "id" IN (:jobIds)
-        GROUP BY "templateId", "id"`,
+        WHERE "templateId" IN (:templateIds) AND "archived" = false
+        GROUP BY "templateId"`,
 
     getPipelineUsageCountForJobTemplatesQueryMySql: (
         tablePrefix = ''
-    ) => `SELECT COUNT(DISTINCT pipelineId) as count, templateId, id
+    ) => `SELECT COUNT(DISTINCT pipelineId) as count, templateId
         FROM \`${tablePrefix}jobs\`
-        WHERE templateId IN (:templateIds) AND id IN (:jobIds)
-        GROUP BY templateId, id`,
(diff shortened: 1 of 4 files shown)


VonnyJap:
the unit tests are currently failing, so I am working to fix them

lib/rawQueries.js
```diff
@@ -3,17 +3,17 @@
 const Queries = {
     getPipelineUsageCountForJobTemplatesQuery: (
         tablePrefix = ''
-    ) => `SELECT COUNT(DISTINCT "pipelineId") as count, "templateId"
+    ) => `SELECT COUNT(DISTINCT "pipelineId") as count, "templateId", "id"
         FROM "${tablePrefix}jobs"
-        WHERE "templateId" IN (:templateIds)
-        GROUP BY "templateId"`,
+        WHERE "templateId" IN (:templateIds) AND "id" IN (:jobIds)
+        GROUP BY "templateId", "id"`,
```
sagar1312: Instead of filtering on `jobIds`, why don't we add the filter to ignore archived jobs

```suggestion
        WHERE "templateId" IN (:templateIds) AND "archived" = false
        GROUP BY "templateId"`,
```

lib/rawQueries.js
```diff
@@ -3,17 +3,17 @@
 const Queries = {
     getPipelineUsageCountForJobTemplatesQuery: (
         tablePrefix = ''
-    ) => `SELECT COUNT(DISTINCT "pipelineId") as count, "templateId"
+    ) => `SELECT COUNT(DISTINCT "pipelineId") as count, "templateId", "id"
```
sagar1312: Why do we need to sub group by `id`?
```suggestion
    ) => `SELECT COUNT(DISTINCT "pipelineId") as count, "templateId", "id"
```

lib/templateFactory.js
```diff
@@ -264,46 +265,60 @@ class TemplateFactory extends BaseFactory {
...
-            };
-
-            const metricBuildsListConfig = { ...metricJobsListConfig };
+            // first need to get the job ids that are not archived for the templates
+            return jobFactory
+                .list({
+                    params: { templateId: templateIds, archived: false },
+                    readOnly: true,
+                    aggregationField: 'id'
+                })
```
sagar1312: We can skip this step.
Instead we could just add `archived: false` filter wherever needed.

lib/templateFactory.js
```diff
@@ -264,46 +265,60 @@ class TemplateFactory extends BaseFactory {
...
+                .then(jobs => {
+                    // then use the jobIds and templateIds as the filters to get the metrics
+                    const jobIds = jobs.map(j => j.id);
+
+                    const metricJobsListConfig = {
+                        params: { templateId: templateIds, id: jobIds },
```
sagar1312:
```suggestion
                        params: { templateId: templateIds, archived: false },
```

sagar1312 (changes requested):

lib/rawQueries.js
```diff
@@ -3,17 +3,17 @@
...
 
     getPipelineUsageCountForJobTemplatesQueryMySql: (
         tablePrefix = ''
-    ) => `SELECT COUNT(DISTINCT pipelineId) as count, templateId
+    ) => `SELECT COUNT(DISTINCT pipelineId) as count, templateId, id
         FROM \`${tablePrefix}jobs\`
-        WHERE templateId IN (:templateIds)
-        GROUP BY templateId`,
+        WHERE templateId IN (:templateIds) AND id IN (:jobIds)
+        GROUP BY templateId, id`,
```
pritamstyz4ever: should we use `archived = false` instead of passing jobIds?

lib/templateFactory.js
VonnyJap: the buildFactory does not have `archived` field in it so we still need a way to filter out the archived job from the buildFactory. And in this case, this step is needed to get the list of jobIds that can be used when calling the buildFactory.

sagar1312: We need not filter out builds that are associated with archived jobs.

VonnyJap: Is there another way to do that using the current `buildFactory` model apart from passing the job ids to filter the builds belong to archived jobs out? When I looked at the [data-schema for the build](https://github.com/screwdriver-cd/data-schema/blob/master/models/build.js#L27-L104), `jobId` was the first came to mind. Opened to input.

sagar1312: We don't need to filter builds based on whether the latest configuration of the associated job no longer uses that template or uses a different template or the job is archived.

This PR should only fix pipeline usage and job usage metrics. There is no change needed for build usage metric.

lib/templateFactory.js
```diff
@@ -265,13 +266,16 @@ class TemplateFactory extends BaseFactory {
...
+                params: { templateId: templateIds, archived: false },
                 readOnly: true,
                 aggregationField: 'templateId'
             };
 
             const metricBuildsListConfig = { ...metricJobsListConfig };
 
+            // archived is not an attribute of build, so we need to filter it out
+            // builds are historical data so no need to filter out builds from archived jobs
+            metricBuildsListConfig.params = { templateId: templateIds };
```
sagar1312, line 278: minor suggestion.

Since `metricBuildsListConfig` is not inheriting any significant config from `metricJobsListConfig`, let us define it directly.
```suggestion
            // builds are historical data so no need to filter out builds from archived jobs
            const metricBuildsListConfig =  {
                params: { templateId: templateIds },
                readOnly: true,
                aggregationField: 'templateId'
            };;
```

sagar1312 (approved):

lib/templateFactory.js
VonnyJap: since this is trivial, I decided to ignore it for now and merged the PR. Thank you for the input though.

PR #613, opened 2024-02-15
fix(2767): Remove associated stage and stageBuilds when deleting pipeline

tkyi:
## Context

When a pipeline is deleted, the stages and the builds associated to the stages should be removed as well.

## Objective

This PR removes related stages and stageBuilds when pipeline is deleted.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/3025

## License

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -1688,9 +1688,10 @@ class PipelineModel extends BaseModel {
     }
 
     /**
-     * Remove all jobs & builds associated with this pipeline,
-     * remove id from all associated collection and the pipeline itself
-     * @return {Promise}        Resolves to null if remove successfully
+     * Remove all jobs & builds and stages and stageBuilds (if applicable)
+     * associated with this pipeline, remove pipeline ID from all associated
+     * collections, and remove the pipeline itself
+     * @return {Promise}        Resolves to null if removed successfully
      */
     remove() {
         const pipelineFactory = this._getPipelineFactory();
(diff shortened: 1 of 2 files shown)


lib/pipeline.js
```diff
@@ -1779,6 +1780,30 @@ class PipelineModel extends BaseModel {
...
+
+            const stages = await stageFactory.list({ params: { pipelineId: this.id } });
+
+            logger.info(`pipelineId:${this.id}: Removing stages and stageBuilds.`);
+
+            stages.map(async s => {
+                const stageBuilds = await stageBuildFactory.list({ params: { stageId: s.id } });
+
+                await Promise.all(stageBuilds.map(sb => sb.remove()));
+            });
```
sagar1312: We could potentially reduce the DB calls by getting stageBuilds for all the stages at once. 
```suggestion
            const stageIds = stages.map(s => s.id);
            const stageBuilds = await stageBuildFactory.list({ params: { stageId: stageIds} });
            await Promise.all(stageBuilds.map(sb => sb.remove()));
```

sagar1312 (changes requested):

lib/pipeline.js

sagar1312 (approved):

lib/pipeline.js
```diff
@@ -1779,6 +1789,27 @@ class PipelineModel extends BaseModel {
...
+            const stageFactory = StageFactory.getInstance();
+            const stageBuildFactory = StageBuildFactory.getInstance();
+            const stages = await stageFactory.list({ params: { pipelineId: this.id } });
+            const stageIds = stages.map(s => s.id);
+            const stageBuilds = await stageBuildFactory.list({ params: { stageId: stageIds } });
+
+            logger.info(`pipelineId:${this.id}: Removing stages and stageBuilds.`);
+
+            // Remove all stageBuilds and stages
+            await Promise.all(stageBuilds.map(sb => sb.remove()));
```
sagar1312, line 1809: We may have to wrap this inside `if(stages.length > 0)`.
Otherwise it may make unnecessary call to DB for pipelines without stages and fail as well.

PR #619, opened 2024-03-26
fix: Add getBuilds method for pipeline model [1]

tkyi:
## Context

Currently a lot of calls are being made on the PR tab page in the UI to: 
- `GET /pipelines/:id/events?count=1&page=1&prNum=prNum`
- `GET /jobs/:id/builds?count=10&page=1`
- `GET /events/:id/builds`

It would be nice if we could simplify these calls to get only latest builds in a PR and make fewer calls to the API.

## Objective

This PR adds a method for `pipeline.getBuilds()` to get builds for pipeline based on groupEventId.

## References
N/A

## License
I confirm that this contribution is made under a BSD license and that I have the authority necessary to make this contribution on behalf of its copyright owner.

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -20,7 +20,7 @@ const MAX_EVENT_DELETE_COUNT = 100;
 // The default page for fetching not by aggregatedInteval
 // And the start page for fetching by aggregatedInteval
 const DEFAULT_PAGE = 1;
-const DEFAULT_COUNT = 50;
+const DEFAULT_COUNT = 10;
 const SCM_NO_ACCESS_STATUSES = [401, 404];
 
 const { getAllRecords, getBuildClusterName, getFullStageJobName } = require('./helper');
@@ -1663,6 +1663,7 @@ class PipelineModel extends BaseModel {
      * Fetch builds belonging to a pipeline or events
      * @param  {Object}   [config]
      * @param  {Number}   [config.sort]                         Sort rangekey by ascending or descending
+     * @param  {Number}   [config.sortBy]                       Sortby field
      * @param  {Object}   [config.paginate]                     Pagination parameters
      * @param  {Number}   [config.paginate.count]               Number of items per page
      * @param  {Number}   [config.paginate.page]                Specific page of the set to return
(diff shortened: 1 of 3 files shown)


lib/pipeline.js
```diff
@@ -1658,6 +1658,75 @@ class PipelineModel extends BaseModel {
...
+
+                events.forEach(e => processed.push(e.getBuilds()));
+
+                return Promise.all(processed).then(builds => {
+                    // flatten array of arrays
+                    return builds.flat(1);
+                });
+            }
+
+            return events;
```
VonnyJap: is there a reason why we return events here instead of builds?

lib/pipeline.js
```diff
@@ -1658,6 +1658,75 @@ class PipelineModel extends BaseModel {
...
+            delete config.params.latest;
+        }
+
+        // Fetch builds for this pipeline
+        const defaultConfig = {
+            params: { pipelineId: this.id }
+        };
+        const listConfig = config ? hoek.applyToDefaults(defaultConfig, config) : defaultConfig;
+
+        return buildFactory.list(listConfig);
```
VonnyJap: Do we keep a retention period for historical builds? I am not sure how much data could be returned as a response for builds of a given pipeline, but I am leaning towards not implementing this relaxation for now. Meaning without the supported params (in this case, it is only `groupEventId` for now), the client will be returned `400 bad request`. 
However I can also agree to support returning all builds for a given pipeline as long as we implement pagination.

lib/pipeline.js
tkyi: It'll be an empty array anyways. so I can hardcode it to that if that makes more sense to you.

Actually since events should never be empty array I think I can remove that if section entirely.

lib/pipeline.js
tkyi: Added default pagination

lib/pipeline.js
```diff
@@ -1658,6 +1659,75 @@ class PipelineModel extends BaseModel {
...
+        if (latest !== undefined) {
+            delete config.params.latest;
+        }
+
+        // Fetch builds for this pipeline with default count and page
+        const defaultConfig = {
+            params: { pipelineId: this.id },
+            paginate: {
+                count: DEFAULT_COUNT
+            },
```
VonnyJap, line 1723: if page is not implicitly defined, it is defaulted to `1`? if yes, then maybe comment should be worded as such:
`// Fetch builds for this pipeline with default count and page (implicitly set to 1)`

tkyi: added

VonnyJap (approved):

PR #620, opened 2024-03-27
fix(3072): PipelineTemplateVersionFactory getWithMetadata() does not return a valid model

sagar1312:
## Context

The return value of `getWithMetadata()` of `PipelineTemplateVersionFactory` has below issues

1. Return value is a plain Javascript object instead of an instance of `PipelineTemplateVersion` model
2. it returns non-null object (containing attributes only from PipelineTemplate) when version does not exist
3. `id`, `createTime` of `PipelineTemplateVersion` instance is replaced by  `id`, `createTime` from the `PipelineTemplate` meta

## Objective

Update the logic to selectively merge `PipelineTemplate` meta fields into `PipelineTemplateVersion` instance
1. Return value must be an instance of `PipelineTemplateVersion` model
2. Return `null` when version does not exist
3. Cherry pick the fields (`name`, `namespace`, `latestVersion`, `pipelineId`, `maintainer`) from `PipelineTemplate` meta to be added to `PipelineTemplateVersion` instance

## References

https://github.com/screwdriver-cd/screwdriver/issues/3072

## License

### DIFF
--- a/lib/pipelineTemplateVersionFactory.js
+++ b/lib/pipelineTemplateVersionFactory.js
@@ -238,7 +238,16 @@ class PipelineTemplateVersionFactory extends BaseFactory {
 
         const pipelineTemplateVersion = await super.get(config);
 
-        return { ...pipelineTemplateVersion, ...pipelineTemplateMeta };
+        if (!pipelineTemplateVersion) {
+            return null;
+        }
+
+        // merge selected template meta fields into template version
+        ['pipelineId', 'namespace', 'name', 'maintainer', 'latestVersion'].forEach(fieldName => {
+            pipelineTemplateVersion[fieldName] = pipelineTemplateMeta[fieldName];
+        });
+
+        return pipelineTemplateVersion;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


pritamstyz4ever (approved):

lib/pipelineTemplateVersionFactory.js
```diff
@@ -238,7 +238,16 @@ class PipelineTemplateVersionFactory extends BaseFactory {
 
         const pipelineTemplateVersion = await super.get(config);
 
-        return { ...pipelineTemplateVersion, ...pipelineTemplateMeta };
+        if (!pipelineTemplateVersion) {
+            return null;
+        }
+
+        // merge selected template meta fields into template version
+        ['pipelineId', 'namespace', 'name', 'maintainer', 'latestVersion'].forEach(fieldName => {
```
adong, line 246: Does `description` need to be here?

VonnyJap (approved):

lib/pipelineTemplateVersionFactory.js
sagar1312: `description` is already a field in `PipelineTemplateVersion`. Here were are copying specific fields from `PipelineTemplate` to `PipelineTemplateVersion`

PR #621, opened 2024-03-29
fix: Cannot fetch builds based on pipelineId [1.5]

tkyi:
## Context

Currently a lot of calls are being made on the PR tab page in the UI to: 
- `GET /pipelines/:id/events?count=1&page=1&prNum=prNum`
- `GET /jobs/:id/builds?count=10&page=1`
- `GET /events/:id/builds`

It would be nice if we could simplify these calls to get only latest builds in a PR and make fewer calls to the API.

## Objective

This PR modifies `pipeline.getBuilds()` to get jobs for a pipeline and get builds for that pipeline based on jobIds.

## References
Related to https://github.com/screwdriver-cd/models/pull/619

## License
I confirm that this contribution is made under a BSD license and that I have the authority necessary to make this contribution on behalf of its copyright owner.

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -1715,9 +1715,15 @@ class PipelineModel extends BaseModel {
             delete config.params.latest;
         }
 
+        // Fetch jobs for this pipeline
+        const jobs = await this.getJobs({
+            params: { pipelineId: this.id }
+        });
+        const jobIds = jobs.map(j => j.id);
+
         // Fetch builds for this pipeline with default count and page (implicitly set to 1)
         const defaultConfig = {
-            params: { pipelineId: this.id },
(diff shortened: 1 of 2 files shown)


lib/pipeline.js
```diff
@@ -1715,9 +1715,15 @@ class PipelineModel extends BaseModel {
             delete config.params.latest;
         }
 
+        // Fetch jobs for this pipeline
+        const jobs = await this.getJobs({
+            params: { pipelineId: this.id }
+        });
+        const jobIds = jobs.map(j => j.id);
```
VonnyJap, line 1725: I think we should by default excluding jobs that are archived.

VonnyJap (approved):

PR #626, opened 2024-08-08
feat(3143): When a event is started from a job in the middle of a stage, execute setup hook before executing that job.

ThehamzaIftikhar:
## Context

Currently if the event is started from a job in the middle of the stage, the build for stage setup is not created

## Objective

If the event is started from a job in the middle of the stage, the startFrom gets replaced to the setup job of the same stage

## References

https://github.com/screwdriver-cd/screwdriver/issues/3143

## License

### DIFF
--- a/test/lib/eventFactory.test.js
+++ b/test/lib/eventFactory.test.js
@@ -180,11 +180,11 @@ describe('Event Factory', () => {
                         { name: '~commit' },
                         { name: 'main' },
                         { name: 'disabledJob' },
-                        { name: 'stage@integration:setup' },
-                        { name: 'int-deploy' },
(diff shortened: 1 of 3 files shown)


test/lib/eventFactory.test.js
```diff
@@ -1269,6 +1269,26 @@ describe('Event Factory', () => {
                 });
             });
 
+            it('should create build for stage setup job, if startFrom is a stage name', () => {
```
sagar1312:
```suggestion
            it('should create build for stage setup job, if startFrom is a job in the middle of a stage', () => {
```

sagar1312 (approved):

PR #629, opened 2024-09-18
fix(3055): remove k8s-vm name from test file [skip ci]

y-oksaku:
## Context

[executor-k8s-vm](https://github.com/screwdriver-cd/executor-k8s-vm) has been archived.
We need to remove this package dependence from active repositories.

## Objective

Remove `executor-k8s-vm` from this repository.
This repository is not dependent on `k8s-vm`, but this repository includes `k8s-vm` name as test text.

## References

https://github.com/screwdriver-cd/screwdriver/issues/3055

## License

### DIFF
--- a/test/lib/build.test.js
+++ b/test/lib/build.test.js
@@ -1597,11 +1597,11 @@ describe('Build Model', () => {
                 state: 'ENABLED',
                 archived: false,
                 pipeline: Promise.resolve(pipelineMockB),
-                permutations: [{ annotations: { 'beta.screwdriver.cd/executor:': 'k8s-vm' }, provider }],
+                permutations: [{ annotations: { 'beta.screwdriver.cd/executor:': 'k8s-test' }, provider }],
                 isPR: () => false,
                 prNum: Promise.resolve(null)
             });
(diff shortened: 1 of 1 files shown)


yk634 (approved):

kumada626 (approved):

test/lib/build.test.js
```diff
@@ -1597,11 +1597,11 @@ describe('Build Model', () => {
                 state: 'ENABLED',
                 archived: false,
                 pipeline: Promise.resolve(pipelineMockB),
-                permutations: [{ annotations: { 'beta.screwdriver.cd/executor:': 'k8s-vm' }, provider }],
+                permutations: [{ annotations: { 'beta.screwdriver.cd/executor:': 'k8s-test' }, provider }],
```
VonnyJap, line 1600: do we have an executor named `k8s-test`?

y-oksaku: No, we don't have the `k8s-test` executor.
However, this executor name is only used as a test string and does not have any dependencies.

Here, we are checking that the annotation is passed correctly, so I think using a test string is better since we can clearly see this string being set up from the test code.
(If we use `k8s`, then it might be set up from the default executor setting.)

PR #635, opened 2024-12-12
feat(3181): Skip execution of a virtual job when an event is started from that job

sagar1312:
BREAKING CHANGE: Virtual job at the beginning of the event is no longer queued for execution. API changes are needed to process such build to mark its status as 'SUCCESS' and trigger its downstream jobs. 

## Context

When an event is started from the virtual job, build for such job is created and queued for execution immediately after creating the event. 

## Objective

For virtual jobs at the beginning in the workflow for an event,
- set the status of the build to `CREATED` instead of `QUEUED` only if it doesn't have freeze windows

## References

https://github.com/screwdriver-cd/screwdriver/issues/3181

## License

### DIFF
--- a/lib/eventFactory.js
+++ b/lib/eventFactory.js
@@ -406,6 +406,13 @@ function createBuilds(config) {
 
                     buildConfig.configPipelineSha = pipelineConfig.configPipelineSha;
 
+                    const { annotations, freezeWindows } = j.permutations[0];
+                    const isVirtualJob = annotations ? annotations['screwdriver.cd/virtualJob'] === true : false;
+                    const hasFreezeWindows = freezeWindows ? freezeWindows.length > 0 : false;
+
+                    // Bypass execution of the build if the job is virtual
+                    buildConfig.start = !isVirtualJob || (isVirtualJob && hasFreezeWindows);
+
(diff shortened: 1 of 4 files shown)


lib/eventFactory.js
```diff
@@ -406,6 +406,13 @@ function createBuilds(config) {
 
                     buildConfig.configPipelineSha = pipelineConfig.configPipelineSha;
 
+                    const { annotations, freezeWindows } = j.permutations[0];
+                    const isVirtualJob = annotations ? annotations['screwdriver.cd/virtualJob'] === true : false;
+                    const hasFreezeWindows = freezeWindows ? freezeWindows.length > 0 : false;
+
+                    // Bypass execution of the build if the job is virtual
+                    buildConfig.start = !isVirtualJob || (isVirtualJob && hasFreezeWindows);
```
minghay:
```suggestion
                    buildConfig.start = !isVirtualJob || hasFreezeWindows;
```
`isVirtualJob` will always evaluate to true for the second  half of the or condition

sagar1312: Done

minghay (approved):

PR #639, opened 2025-01-21
feat(3266): make changes to models banner to include scope and scopeId

VonnyJap:
## Context

The current banner schema is implemented at a global level, lacking any concept of scope or scopeId.

## Objective

There are instances where a banner is required only at the pipeline or build level. This change incorporates this concept into the banner model.

## References

https://github.com/screwdriver-cd/screwdriver/issues/3266

## License

### DIFF
--- a/lib/bannerFactory.js
+++ b/lib/bannerFactory.js
@@ -45,10 +45,7 @@ class BannerFactory extends BaseFactory {
         if (!config.isActive) {
             config.isActive = false;
         }
-        if (!config.scope) {
-            config.scope = 'GLOBAL';
-            config.scopeId = null;
-        }
+        
         if ((config.scope === 'PIPELINE' || config.scope === 'BUILD') && !config.scopeId) {
             throw new Error(`scopeId is required when scope is ${config.scope}`);
         }
--- a/lib/bannerFactory.js
+++ b/lib/bannerFactory.js
@@ -1,8 +1,9 @@
 'use strict';
 
 const BaseFactory = require('./baseFactory');
+const PipelineFactory = require('./pipelineFactory');
+const BuildFactory = require('./buildFactory');
 const Banner = require('./banner');
-const { config } = require('screwdriver-data-schema');
 
 let instance;
 
@@ -27,6 +28,7 @@ class BannerFactory extends BaseFactory {
         return new Banner(config);
     }
 
+    // todo: need to make sure some uniqueness here
     /**
      * Create a Banner model
      * @param {Object}  config
@@ -37,7 +39,7 @@ class BannerFactory extends BaseFactory {
      * @param {String}  [config.scope='GLOBAL']    Scope of the banner (GLOBAL|PIPELINE|BUILD)
      * @memberof BannerFactory
      */
-    create(config) {
+    async create(config) {
         if (!config.type) {
             config.type = 'info';
         }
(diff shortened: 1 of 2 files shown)


lib/bannerFactory.js
```diff
@@ -33,21 +35,83 @@ class BannerFactory extends BaseFactory {
...
         if (!config.type) {
             config.type = 'info';
         }
         if (!config.isActive) {
             config.isActive = false;
         }
+        if (!config.scope) {
+            config.scope = 'GLOBAL';
+            config.scopeId = null;
+        }
```
sagar1312: This should be handled in the API endpoint.
https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/banners/create.js#L60
```suggestion
        
```

lib/bannerFactory.js
```diff
@@ -33,21 +35,83 @@ class BannerFactory extends BaseFactory {
...
+        if ((config.scope === 'PIPELINE' || config.scope === 'BUILD') && !config.scopeId) {
+            throw new Error(`scopeId is required when scope is ${config.scope}`);
+        }
```
sagar1312: same here. This should be taken care by the payload validation using schema in the API endpoint.
```suggestion
        
```

lib/bannerFactory.js
```diff
@@ -33,21 +35,83 @@ class BannerFactory extends BaseFactory {
...
+                throw new Error(`Pipeline ${config.scopeId} does not exist`);
+            }
+        }
+        if (config.scope === 'BUILD') {
+            const build = await BuildFactory.getInstance().get(config.scopeId);
+
+            if (!build) {
+                throw new Error(`Build ${config.scopeId} does not exist`);
+            }
+        }
```
sagar1312: Can we refactor the duplicate code block into a reusable function to validate the `scopeId` by choosing the appropriate factory based on the `scope`

lib/bannerFactory.js
```diff
@@ -33,21 +35,83 @@ class BannerFactory extends BaseFactory {
...
+     *
+     * @param {Object} config - The configuration object for the list request.
+     * @returns {Promise<Array>} A promise that resolves to an array of banners with default scope values set.
+     */
+    async list(config) {
+        const response = await super.list(config);
+
+        return response.map(banner => this._setScopeDefaults(banner));
+    }
+
```
sagar1312: I believe we don't need this. Once the DB schema changes are applied, the defaults should be set in the DB for all the existing records.
```suggestion
```

sagar1312 (changes requested):

lib/bannerFactory.js
```diff
@@ -33,16 +35,33 @@ class BannerFactory extends BaseFactory {
...
         if (!config.isActive) {
             config.isActive = false;
         }
 
+        const scopeFactories = {
+            PIPELINE: PipelineFactory,
+            BUILD: BuildFactory
+        };
+
+        if (scopeFactories[config.scope]) {
```
sagar1312: Other than `GLOBAL`, we need to ensure that we validate the `scopeId`
```suggestion
        if (config.scope !== 'GLOBAL') {
```

sagar1312 (changes requested):

sagar1312 (approved):

PR #642, opened 2025-02-27
feat(#11530): merge admin annotation override on job update

pritamstyz4ever:
## Context

Build Cluster annotation on a job is currently only derived from pipeline or from screwdriver config.
Screwdriver Admin should have a way to override the behavior to move jobs to specific cluster

## Objective

This PR adds logic 
1. to check for admin override annotation from db and merge it to the job annotation if it exits.
2. select build cluster for a job if the override annotation exists
3. add new fn updateBuildCluster to allow updating annotation without any other job side effects

## References

https://github.com/screwdriver-cd/screwdriver/issues/3292

## License

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -17,6 +17,7 @@ const MAX_METRIC_GET_COUNT = 1000;
 const MAX_BUILD_DELETE_COUNT = 100;
 const DEFAULT_COUNT = 10;
 const DEPLOY_KEY_SECRET = 'SD_SCM_DEPLOY_KEY';
+const BUILD_CLUSTER_OVERRIDE_ANNOTATION = 'screwdriver.cd/sdAdminBuildClusterOverride';
 
 /**
  * Find metrics and step metrics related to the build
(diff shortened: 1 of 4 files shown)


lib/job.js
```diff
@@ -240,6 +241,17 @@ class Job extends BaseModel {
         const oldJob = await jobFactory.get(this.id);
         const oldPeriodic = getAnnotations(oldJob.permutations[0], 'screwdriver.cd/buildPeriodically');
 
+        // check if override exits in datastore
+        const buildClusterOverride = getAnnotations(oldJob.permutations[0], BUILD_CLUSTER_OVERRIDE_ANNOTATION);
+
+        // merge admin annotation
+        if (buildClusterOverride) {
```
sagar1312, line 248: Can the new annotation be configured for any job in the `screwdriver.yaml`?

lib/job.js
```diff
@@ -429,6 +441,28 @@ class Job extends BaseModel {
...
         );
     }
+
+    /**
+     * Update a job's build cluster annotation information
+     * This method should not be used to produce any other side effects
+     * @method update
+     * @return {Promise}
+     */
+    async updateBuildCluster() {
```
sagar1312, line 451: It is not clear why do we need this method.

`update()` is called when the old and new value does not match, which persists the entire job model in DB.
So, even when the values match, there should not be any issues if we persist the entire job model in DB.
Why don't we directly call `update()`?

lib/job.js
VonnyJap: I wonder what will happen if a user has [this annotation](https://github.com/screwdriver-cd/data-schema/blob/master/config/annotations.js#L31) in the `screwdriver.yaml`? Technically this annotation should not be configurable from screwdriver config, can we confirm this behavior?

pritamstyz4ever: Annotation should not be configured manually for any job, this is only via admin api

lib/job.js
pritamstyz4ever: added an additional check to prevent calling this function for updating anything other than the annotation itself, this is called by the api's here.
https://github.com/screwdriver-cd/screwdriver/blob/feat/SCREWDRV-11529/plugins/jobs/buildCluster/update.js#L67
https://github.com/screwdriver-cd/screwdriver/blob/feat/SCREWDRV-11529/plugins/jobs/buildCluster/remove.js#L54
we don't want this function to be used for any other purpose then intended. 
We also have job.update() which can be called directly instead of this new function, however we also need the merging logic in job.update fn for the annotation (i.e. do not override the annotation if it exits in db), which means we cannot use it when deleting the annotation.

lib/job.js
sagar1312: Have we added the logic in config-parser to invalidate the pipeline configuration (screwdriver.yaml) if it contains this new annotation?

lib/job.js
VonnyJap: in general, I am fine and more leaning towards having a separate function for this purpose, it sounds unnecessary but I rather do that than having to risk a side effect by using the `job.update()`

minghay (approved):

PR #644, opened 2025-04-25
feat(3322): Add a DELETING state to the pipeline to block updates during the pipeline deletion [2]

y-oksaku:
## Context

Pipeline related data is remaining after delete pipeline.
Details - https://github.com/screwdriver-cd/screwdriver/issues/3322

## Objective

- Check no builds are running before delete pipeline
- Change `pipeline.state` to DELETING just before delete the pipeline
- Block update methods if the state is DELETING

## References

Issue - https://github.com/screwdriver-cd/screwdriver/issues/3322
PRs
- [1] https://github.com/screwdriver-cd/data-schema/pull/598
- [3] https://github.com/screwdriver-cd/screwdriver/pull/3327

## License

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -488,6 +488,10 @@ class PipelineModel extends BaseModel {
      * @return  {Promise}
      */
     async addWebhooks(webhookUrl) {
+        if (this.state === 'DELETING') {
+            throw boom.conflict('This pipeline is being deleted.');
+        }
+
         const webhookUrlsWithActionsList = [];
         const webhookUrlWithActions = {
             webhookUrl,
@@ -557,6 +561,10 @@ class PipelineModel extends BaseModel {
      * @return {Promise}
      */
     async syncPRs() {
+        if (this.state === 'DELETING') {
+            throw boom.conflict('This pipeline is being deleted.');
+        }
+
         const token = await this.token;
 
         return Promise.all([
@@ -591,6 +599,10 @@ class PipelineModel extends BaseModel {
      * @return  {Promise}
      */
     async syncPR(prNum) {
+        if (this.state === 'DELETING') {
+            throw boom.conflict('This pipeline is being deleted.');
+        }
+
         const token = await this.token;
         const prInfo = await this.scm.getPrInfo({
             scmUri: this.scmUri,
@@ -819,6 +831,12 @@ class PipelineModel extends BaseModel {
         const pipeline = await pipelineFactory.get({ scmUri });
 
         if (pipeline) {
+            if (pipeline.state === 'DELETING') {
+                logger.error(`pipelineId:${this.id}: Pipeline ${scmUrl} is been deleted.`);
+
+                return null;
+            }
+
             // Child pipeline belongs to this parent, update it
             if (pipeline.configPipelineId === this.id) {
                 pipeline.admins = newAdmins;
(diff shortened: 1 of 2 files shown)


yakanechi (approved):

lib/pipeline.js
```diff
@@ -1957,22 +1996,24 @@ class PipelineModel extends BaseModel {
...
+            await removeSecrets(); // remove secrets
+            await removeTokens(); // remove tokens
+            await removeTriggers(); // remove triggers
+            await removeJobs(true); // remove archived jobs
+            await removeJobs(false); // remove non-archived jobs
+            await removeEvents('pipeline'); // remove pipeline events
+            await removeEvents('pr'); // remove pr events
+            await removeChildPipelines(); // remove pr events
+            await removeStagesAndStageBuilds(); // remove stages and stageBuilds
+            await removeFromCollections();
```
minghay, line 2009: Is it safe to run these all in parallel rather than sequentially?

lib/pipeline.js
```diff
@@ -1806,7 +1828,24 @@ class PipelineModel extends BaseModel {
      * collections, and remove the pipeline itself
      * @return {Promise}        Resolves to null if removed successfully
      */
-    remove() {
+    async remove() {
+        const latestBuilds = await this.getBuilds();
+
+        if (latestBuilds.some(b => ['QUEUED', 'RUNNING', 'BLOCKED'].includes(b.status))) {
```
sagar1312, line 1834: Should we include all the non-terminal statuses?
Ex: `FROZEN`, `CREATED`
These may get transitioned to `QUEUED` or `RUNNING` when the deletion is in progress.

y-oksaku: I excluded the `FROZEN` and `CREATED` statuses for the following reasons:

- I think `FROZEN` is not running status, it is a type of scheduled build. Therefore we can block the execution of scheduled builds at the `POST /events` endpoint.
- Sometime `CREATED` builds remain.
  - For instance, the build `c`, which has `requires: [ a, b ]`, is created after the build `a`. However if the job `b` is disabled, the build `c` is not started and not deleted.

lib/pipeline.js
y-oksaku: We can run them in parallel.
However, it would put more pressure on the database, especially with a large pipeline.

Therefore, I believe it is safer to run them sequentially.

minghay (approved):

lib/pipeline.js
```diff
@@ -1806,7 +1828,24 @@ class PipelineModel extends BaseModel {
...
+        }
+
+        if (this.state === 'DELETING') {
+            logger.error(`pipelineId:${this.id}: This pipeline is being deleted.`);
+
+            throw boom.conflict('This pipeline is being deleted.');
+        }
+
+        this.state = 'DELETING';
+        await this.update();
```
pritamstyz4ever, line 1847: @y-oksaku This code `await this.update();` introduces an additional check during deletion, it looks for the repo admin to get the token and update the pipeline. However, in a scenario where user has already deleted a repo and SD admin is trying to delete the pipeline fails with unhandled exception as the repo does not exist. Do we need the pipeline update step in this case, can we manage the 'DELETING' state without needing to call the update fn?

pritamstyz4ever (commented):
Found a bug that was introduced due to this PR

lib/pipeline.js
y-oksaku: Apologies for the bug that was found.

Since updating the state is necessary, I have fixed it by making a direct call to `super.update()`.
https://github.com/screwdriver-cd/models/pull/650

PR #655, opened 2025-09-12
fix: Update stop all steps and build when executor error

yakanechi:
## Context

There is no error handling for the executor calling when starting or updating a build.
Consequently, if the executor has not started or cannot be called successfully due to a network error, the build status will remain queue.

スクリーンショット 2025-09-12 15 23 06

## Objective

Added error handling to ensure the build fails correctly if the Executor call fails.
And that error will also be notified to the user.

スクリーンショット 2025-09-12 15 22 45

## License

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -811,7 +811,7 @@ class BuildModel extends BaseModel {
                     }
 
                     return this[executor].startTimer(config).catch(err => {
-                        logger.error(`Failed to start ${this.id} build: ${err}`);
+                        logger.error(`Failed to update ${this.id} build: ${err}`);
 
                         return this.updateToFailureByExecutorError(pipeline).then(() => {
                             throw err;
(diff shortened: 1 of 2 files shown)


minghay (approved):

lib/build.js
```diff
@@ -731,7 +756,11 @@ class BuildModel extends BaseModel {
                                 config.provider = job.permutations[0].provider;
                             }
 
-                            return this[executor].start(config);
+                            return this[executor].start(config).catch(err =>
+                                this.updateToFailureByExecutorError(pipeline).then(() => {
+                                    throw err;
```
y-oksaku, line 763: I think it would be good to output the error log.

```js
logger.error('Failed to start a build ${this.id}: ${error})
```

y-oksaku (approved):

minghay (approved):

PR #656, opened 2025-09-26
fix(3400)[1]: add new badge raw query

yuichi10:
## Context

Pipeline id and create time index cause a loading problem in our environment.
https://github.com/screwdriver-cd/data-schema/pull/607

We want to use raw query to improve badge api instead of using an index.

## Objective

- Add raw query for badge api.

## References

https://github.com/screwdriver-cd/screwdriver/issues/3400

## License

### DIFF
--- a/lib/rawQueries.js
+++ b/lib/rawQueries.js
@@ -86,7 +86,31 @@ const Queries = {
             WHEN name LIKE 'PR-%' THEN (archived = false OR (SUBSTRING(name, 1, POSITION(':' IN name) - 1) IN (:prNames)))
             ELSE false
         END
-        ORDER BY id ASC`
+        ORDER BY id ASC`,
+
+    getPipelineTypeBuildEventsQuery: (tablePrefix = '') => `
+        SELECT *
+        FROM (
+            SELECT "id", "parentEventId", "groupEventId", "causeMessage", "createTime", "creator",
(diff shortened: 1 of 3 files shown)


lib/rawQueries.js
```diff
@@ -86,15 +86,40 @@ const Queries = {
...
+        SELECT *
+        FROM (
+            SELECT id, parentEventId, groupEventId, causeMessage, createTime, creator,
+                   pipelineId, sha, configPipelineSha, startFrom, type, pr, prNum, baseBranch, meta
+            FROM \`${tablePrefix}events\`
+            WHERE pipelineId = :pipelineId
+            AND type = 'pipeline'
+        ) AS filtered_results
+        WHERE meta LIKE '%"build":%'
+        ORDER BY createTime DESC`
```
minghay, line 113: Do you need to do a subselect query to make mySQL use the correct index?

yuichi10: The main reason is to reduce the number of rows on which the LIKE condition is applied.
As a secondary benefit, this approach can also help MySQL use indexes more effectively, since there’s a risk that indexes might be bypassed when filtering with meta LIKE.

We have confirmed that the query speed have improved in our environment.

minghay (approved):

foxtrot0304 (approved):
LGTM !

ibu1224 (approved):

PR #662, opened 2026-02-03
feat(3460): Migrate child pipelines when parent pipeline is migrated to another SCM

sagar1312:
## Context

When a Screwdriver pipeline having child pipelines configured is migrated from one SCM (Ex: `GitHub`) to another SCM (Ex: `BitBucket`), only the parent pipeline gets migrated.
New child pipelines are created for the new SCM child pipeline `scmUrls`.
Child pipelines for the previous SCM child pipeline `scmUrls` continue to exist in `ACTIVE` state.

This defeats the goal of reusing the pipeline ID for child pipelines.

**Sample - Pipeline configuration in source SCM (GitHub)**
https://github.com/sagar1312/screwdriver-migration-external-config-test-pipeline

**Sample - Pipeline configuration in source SCM (BitBucket)**
https://bitbucket.org/sagar-1312/screwdriver-migration-external-config-test-pipeline/src/main/screwdriver.yaml

**Child pipelines before migration**
Image

**Child pipelines after migration**
Image

## Objective

When a parent pipeline is migrated to another SCM, event the child pipelines must be migrated.

**Note:**
Child pipelines should be migrated only if repo name, branch and root directory match. Otherwise the child pipelines associated with previous SCM would be marked as `INACTIVE`.

Image

## References

https://github.com/screwdriver-cd/screwdriver/issues/3460

## License

### DIFF
--- a/test/lib/pipeline.test.js
+++ b/test/lib/pipeline.test.js
@@ -206,18 +206,15 @@ describe('Pipeline Model', () => {
     };
 
     const getChildPipelineMock = config => {
-        const childPipeline = {
+        return {
             ...config,
-            state: 'ACTIVE',
+            state: config.state || 'ACTIVE',
             id: Date.now(),
             configPipelineId: testId,
             remove: sinon.stub().resolves(null),
-            sync: sinon.stub().resolves(null)
+            sync: sinon.stub().resolves(null),
+            update: sinon.stub().resolves(null)
         };
-
-        childPipeline.update = sinon.stub().resolves(pipeline);
-
-        return childPipeline;
     };
 
     beforeEach(() => {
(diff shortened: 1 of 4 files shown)


test/lib/pipeline.test.js
```diff
@@ -187,7 +194,110 @@ describe('Pipeline Model', () => {
...
+
+        console.log(
+            [
+                childPipelineGitLabFooConfig,
+                childPipelineGitLabBarConfig,
+                childPipelineGitLabBazConfig,
+                childPipelineGitHubFooConfig,
+                childPipelineGitHubBarConfig,
+                childPipelineGitHubBazConfig
+            ].length
```
minghay:
```suggestion
```

DekusDenial (approved):

test/lib/pipeline.test.js
```diff
@@ -187,7 +194,107 @@ describe('Pipeline Model', () => {
...
+        );
```
minghay:
```suggestion
```

sagar1312: removed

minghay (approved):

